PR #35, opened 2017-02-13
Add homebrew formula

HazAT:
We should also probably automate updating this file


HazAT:
We moved it to a seperate repo ... we should still consider creating the formula in the CI
https://github.com/getsentry/homebrew-tools

PR #42, opened 2017-03-12
Add package & install script for npm

jskrzypek:
This PR adds the needed files to allow users to install via npm, as discussed in #31

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "sentry-cli",
+  "version": "0.25.0",
+  "description": "A command line utility to work with Sentry. https://docs.sentry.io/hosted/learn/cli/",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "preinstall": "INSTALL_DIR=./bin ./bin/install-sentry-cli"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/getsentry/sentry-cli.git"
+  },
+  "keywords": [
+    "sentry",
+    "sentry-cli",
+    "cli"
+  ],
+  "bugs": {
+    "url": "https://github.com/getsentry/sentry-cli/issues"
+  },
+  "bin": {
+    "sentry-cli": "bin/sentry-cli"
+  },
+  "homepage": "https://docs.sentry.io/hosted/learn/cli/",
+  "license": ""
+}
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "sentry-cli",
-  "version": "0.25.0",
+  "version": "0.28.0",
   "description": "A command line utility to work with Sentry. https://docs.sentry.io/hosted/learn/cli/",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "preinstall": "INSTALL_DIR=./bin ./bin/install-sentry-cli"
+    "preinstall": "./bin/install-sentry-cli"
   },
   "repository": {
     "type": "git",
@@ -22,5 +22,5 @@
     "sentry-cli": "bin/sentry-cli"
   },
   "homepage": "https://docs.sentry.io/hosted/learn/cli/",
-  "license": ""
+  "license": "BSD-3-Clause"
 }
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,11 @@
+# .npmignore
+
+# ignore all
+*
+
+# unignore desired files
+!package.json
+!README.md
+!License
+!bin/
+!bin/*
--- a/.npmignore
+++ b/.npmignore
@@ -7,5 +7,6 @@
 !package.json
 !README.md
 !License
+!install-sentry-cli
 !bin/
 !bin/*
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "sentry-cli",
+  "version": "0.25.0",
+  "description": "A command line utility to work with Sentry. https://docs.sentry.io/hosted/learn/cli/",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "preinstall": "INSTALL_DIR=./bin ./bin/install-sentry-cli"
```
mattrobenolt: Is there a way that we can just leverage our existing install script without duplicating it in two places?

If there are changes to name to the script to accommodate working for both, I think we'd be more inclined to support that, in one place, than a copy/paste in two places where one that we never use will absolutely break at some point in the future. :)

jskrzypek: I couldn't see where you had the source for the script​ in the repo, which is why I copied it. If its in a different repo, you could just as eaily put the `package.json`, `bin/`, & `.npmignore` files there and that way you'd just bump the version in the package & get-cli script simultaneously in your build process.

The other alternative (which seems hackier to me, but may be easier for you guys to support) is to change the `preinstall` npm script to be 
```json
"preinstall": "INSTALL_DIR=./bin curl -sL https://sentry.io/get-cli/"
```

jskrzypek: Let me know which path you prefer to take and i can modify/close the PR as needed

package.json
```diff
@@ -0,0 +1,26 @@
...
+    "cli"
+  ],
+  "bugs": {
+    "url": "https://github.com/getsentry/sentry-cli/issues"
+  },
+  "bin": {
+    "sentry-cli": "bin/sentry-cli"
+  },
+  "homepage": "https://docs.sentry.io/hosted/learn/cli/",
+  "license": ""
```
mitsuhiko: The license of sentry-cli atm is a 3 clause BSD. So that should match it.

package.json
mitsuhiko: I'm okay with having the duplicated script here but we should strip it down to the bare minimum. In particular the sudo stuff does not belong there any more and we should depend on `npm_package_version`.

mitsuhiko:
I'm happy to merge this if my above comments are addressed I think.

mitsuhiko:
(Though I do wonder if there is a better way to download this than curl. Because what about windows users?)

jskrzypek:
@mitsuhiko I was also wondering about the portability of curl, but I noticed that's what you're doing inside the install script you are serving at `https://sentry.io/get-cli/` so I stuck with it.

mitsuhiko:
Yeah, just that never supported windows and is documented as such :)

mattrobenolt:
Throwing in my two cents: since we're operating under such a controlled environment and the install script is so minimal, can we just port it to Node and let Node download and store the binary? Seems like we'd at least gain compatibility with platforms the original bash script is not compatible with. That would at least be a win for us, imo.

jskrzypek:
Well, it has the `Windows_i686` link... 😄 

@mattrobenolt You beat me to the suggestion.

An example of something like that can be seen here, where we're using this script internally to make sure we can get wkhtmltopdf on heroku:
https://github.com/Yoobic/wkhtmltopdf/blob/master/wkhtmltopdf.js

I need to check, but we may need to do write it as pure node without the dependency on wget-improved in order for it to run during "preinstall"

jskrzypek:
Ok so I tried it and I can't use the same `wget-improved` library in `preinstall` because it hasn't been installed yet (*duh*).
I'm working on a version that will use the `https` builtin module which is a bit of a headache and I need to handle the 302 redirects manually, but it should be fine. I should have something working by tomorrow.

mitsuhiko:
Awesome! 👍

jskrzypek:
Sorry for the delay, was kind of busy with work the past couple days and I didn't have time to get back to this. 

The new script runs, downloads a functional binary, and checks that the output of `<downloaded_sentry_cli> --version` matches the version in the `package.json` (which is mostly a check that ensures the binary executes correctly).

jskrzypek:
I realized you guys had released a new version, so I rebased & bumped the package.json version to 0.27.1. I suppose you'll need to add the `npm version` & `publish` commands to your build process.

mitsuhiko:
Awesome. Thanks so much. I will make sure with the next release (today or tomorrow) this gets merged and pushed to npm.

jskrzypek:
Great :)

ghaiat:
hello
Any update on this merge?
Cheers

mitsuhiko:
It will land with the next release of sentry-cli

Regards,
Armin

jskrzypek:
So I only just bothered to check that the sentry-cli name is available on
npm... Apparently it isn't :( but i guess you guys will figure out what to
do with that.

For now we're just going to point to the GitHub repo

mitsuhiko:
It's called `sentry-cli-binary` now and I just added it to the docs as well. https://docs.sentry.io/learn/cli/#installation-via-npm

PR #132, opened 2017-10-12
feat: Add --ignore and --ignore-file to releases files upload-sourcemaps

### LINKED ISSUE #126: Ignore directories flag
Given `sentry-cli releases files <release> upload-sourcemaps .` I want to be able to specify directories that should be ignored during traversal.

eg.

```
sentry-cli releases files a4mg732d upload-sourcemaps . --ignore-dir 'node_modules,some_other_dir'
```

jan-auer:
This PR adds two flags to the releases:files:upload-sourcemaps command:

 - `-i` or `--ignore`: Ignore a certain glob **relative to the cwd**.
   This allows to, for instance, exclude `node_modules/`.
 - `-I` or `--ignore-file`: Set a file containing one more more ignore
   patterns **relative to the file's path**. Any file that is a valid
   gitignore file can also be used here. This also includes overrides
   in the form of `!pattern`. If the specified file is not found in the
   cwd, sentry-cli traverses up the directory tree until it finds the
   file.

Both options can be used together, with the patterns declared in `--ignore`
taking precedence. There is no default ignore file configured, but it would
be easy to default to something like `.sentryignore`.

Along with these changes, file extension matching has been rewritten
to use the `ignore` crate. Command line options and defaults remain
the same.

Fixes #126

### DIFF
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -1,11 +1,12 @@
 //! Implements a command for managing releases.
-use std::path::{Path, PathBuf};
-use std::ffi::OsStr;
+use std::path::Path;
 use std::collections::HashSet;
 
 use clap::{App, AppSettings, Arg, ArgMatches};
-use walkdir::WalkDir;
 use chrono::{DateTime, Duration, Utc};
+use ignore::WalkBuilder;
+use ignore::types::TypesBuilder;
+use ignore::overrides::OverrideBuilder;
 use regex::Regex;
 
 use prelude::*;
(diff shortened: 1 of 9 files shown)


src/commands/releases.rs
```diff
@@ -227,6 +228,18 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
...
+                    .short("i")
+                    .value_name("IGNORE")
+                    .multiple(true)
+                    .help("Ignores all files and folders matching the given glob"))
+                .arg(Arg::with_name("ignore_file")
+                    .long("ignore-file")
+                    .short("I")
+                    .value_name("IGNORE_FILE")
+                    .help("Ignores all fiels and folders specified in the given \
+                            ignore file, e.g. .gitignore. Defaults to .sentryignore"))
```
kamilogorek: This default value is not set right now, or is it?

kamilogorek:
Tested using this PR https://github.com/getsentry/sentry-cli-webpack-plugin/pull/2/files and binary built from this branch and everything works just fine 👌

PR #151, opened 2017-11-20
feat: Optionally include binaries when uploading dSYMs

jan-auer:
This adds support to include binaries in dSYM uploads to allow server-side
stackwalking. Binaries are either executables or dynamic libraries. The Sentry
server will then extract relevant information from the binaries and merge
them into the dSYMs.

The reasoning behind uploading complete binaries is to maintain flexibility
for further processing choices. For now, this does not require too much
processing on the client side and we can stabilize our processing pipeline
on the server side, first. Once the process is stable, we can start pulling
symbolic into sentry-cli and process / strip debug symbols on the client
side already.

The PR contains the following changes:

 - First, the `filetype` flag is read from MachO headers. In FAT files, the
   file type is only located in the architecture's MachoHeaders. This forces
   us to iterate over the entire list of architectures and verify that all
   types match. Otherwise a type `MachoFileType::Inconsistent` is returned.
   We will have to see, if this happens out in the wild, but for now we can
   assume, that only matching file types get merged into a universal binary.

 - The `upload-dsym` CLI has a new parameter `--binaries`, which enables the
   search for binaries. A binary is either `MachoFileType::Executable`
   (`MH_EXECUTE`) or `MachoFileType::Dylib` (`MH_DYLIB`).

 - The file walker does not only check for dwarf data anymore, but also
   whether the filetype is one of the binary types. Since these files do not
   contain the `__DWARF` section, the `has_dwarf_data()` check is omitted.

 - Found files are now consistently keyed by a tuple `(Uuid, MachoFileType)`.
   Duplicate detection now also considers the file type in addition to the
   uuid. As a result, the early exit condition `BatchIter::found_all` now
   considers binaries too, if the `--binaries` option is set.

Note that the Sentry server currently ignores these binaries. Thus, they
will be uploaded repeatedly until we start to handle them.

### DIFF
--- a/src/commands/upload_dsym.rs
+++ b/src/commands/upload_dsym.rs
@@ -193,14 +193,18 @@ struct BatchIter<'a> {
     open_zip_index: usize,
     uuids: Option<&'a HashSet<Uuid>>,
     allow_zips: bool,
+    allow_bins: bool,
     found_uuids: RefCell<&'a mut HashSet<(Uuid, MachoFileType)>>,
 }
 
 impl<'a> BatchIter<'a> {
-    pub fn new<P: AsRef<Path>>(path: P, max_size: u64, uuids: Option<&'a HashSet<Uuid>>,
-                               allow_zips: bool, found_uuids: &'a mut HashSet<(Uuid, MachoFileType)>)
-        -> BatchIter<'a>
-    {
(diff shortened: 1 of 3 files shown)


jan-auer:
Yep, this will be in a separate PR, though.

Issue #168, opened 2017-12-05
Errors are printed to stdout instead of stderr?


kirkins:
I need to run sentry in a situation where connection is often lost. I'm trying:

    sentry-cli send-event -m "Hello from Sentry" || echo sentry failed do something else

I purposely disconnect the device from the Internet so `sentry-cli` will fail. It outputs an error but it seems an error isn't triggered because my `echo` doesn't run.

jan-auer:
Fix will be released with `1.26.0`

PR #169, opened 2017-12-05
Use fail when event can't send

kirkins:
https://github.com/getsentry/sentry-cli/issues/168


kirkins:
Sure thing, I was copying from this snippet. Must be the old way:

https://github.com/getsentry/sentry-cli/blob/e66a69e4604b5796c4fd39ae01860624b4bb82d6/src/config.rs#L263

jan-auer:
Indeed, we should probably streamline that 😄 
Thanks a lot! 🎉

Issue #180, opened 2017-12-08
Source maps options support


montogeek:
Hi!

I am using `sentry-cli` binary to uploads my source maps using npm script, `"uploadMaps": "sentry-cli releases files ${VERSION} upload-sourcemaps ./dist --validate --url-prefix '~/path/path' && rm dist/*.map",`, it works but not the best solution.

I want to migrate it to Sentry Source maps webpack plugin, but I see the new (merge 2 days ago) JS interface does not support all the possible options you can pass when using `release` command, at this moment it is passing by default `--rewrite` option.

https://github.com/getsentry/sentry-cli/blob/master/js/index.js#L62-L68

Is there any plan to support it? Would you accept a PR to add it?

Thanks!

jan-auer:
> Would you accept a PR to add it?

Absolutely!

montogeek:
Implemented on #182

PR #182, opened 2017-12-08
feat(node) Add support for all upload-source-maps options

montogeek:
As discussed on https://github.com/getsentry/sentry-cli/issues/180, this PR add support for all `upload-source-maps` command options.

Prettier formatting is kind of weird, but elegible

I tried to add a test, but it seems it test against the real CLI, anyway I left it commented.


js/index.js
```diff
@@ -8,6 +8,14 @@ var path = require('path');
...
 
 var DEFAULT_IGNORE = ['node_modules'];
+var SOURCEMAPS_OPTIONS = {
+  noSourceMapReference: '--no-sourcemap-reference',
+  stripPrefix: '--strip-prefix',
+  stripCommonPrefix: '--strip-common-prefix',
+  validate: '--validate',
+  urlPrefix: '--url-prefix',
+  ext: '--ext'
+};
```
jan-auer: It is still passing the `--rewrite` option by default (see line 76), you might wanna add this here.

js/index.js
```diff
@@ -68,6 +76,21 @@ SentryCli.prototype.uploadSourceMaps = function(options) {
         '--rewrite'
       ];
 
+      var sourceMapOptions = Object.keys(SOURCEMAPS_OPTIONS).reduce(function(
+        newOptions,
+        sourceMapOption
+      ) {
+        if (options[sourceMapOption]) {
+          newOptions.push(SOURCEMAPS_OPTIONS[sourceMapOption], options[sourceMapOption]);
```
jan-auer: This is a bit problematic because of two conditions: 
 - The `--strip-prefix` option allows either a single value or an array. In the latter case, the option name has to be repeated, e.g.: `--strip-prefix foo --strip-prefix bar`. Have a look at `transformIgnore` above.
 - Boolean flags like `--rewrite` or `--validate` do not expect a value, but this code actually invokes `--rewrite true`. This leads to an error, as "true" is interpreted as a path.

kamilogorek:
> Is there any reason to not let the user configure --rewrite? 

@jan-auer I talked with @mitsuhiko about this, and we agreed that it should be always passed, as we can only get better source maps, never worse.

js/index.js
kamilogorek: Agree. This whole function should be extracted into utility function and it should make use of `transformIgnore` (which can be modified to suit our needs).

jan-auer:
Please also merge `master` to get the fix from #183

montogeek:
@jan-auer I addressed your comments, still left `rewrite` by default per @kamilogorek comment.

js/__tests__/index.test.js
```diff
@@ -18,4 +18,29 @@ describe('SentryCli', function() {
       expect(e.message).toMatch('Command failed:');
     });
   });
+
+  // test('call sentry-cli with source map options', function() {
```
kirkins: Was this meant to be commented in the final PR?

montogeek: > Don't worry about the tests, we will do that afterwards. Just leave the comment in. However, you can test locally by calling npm install, which will also download the current binary.

kirkins: ah, sorry was just curious.

montogeek: The first test doesn't make sense either... :D

js/index.js
```diff
@@ -8,23 +8,58 @@ var path = require('path');
...
         return acc.concat(value);
       }, []);
   }
-  return ['--ignore', ignore];
+  return [option, values];
+}
+
+function normalizeOptions(options) {
+  var newOptions = [];
+  return Object.keys(SOURCEMAPS_OPTIONS).reduce(function(acc, sourceMapOption) {
```
kamilogorek: You don't need `newOptions` here, as this is what `acc` is for in `reduce` functions :)

js/index.js
```diff
@@ -68,16 +103,14 @@ SentryCli.prototype.uploadSourceMaps = function(options) {
...
       ];
 
-      if (options.ignoreFile) {
-        command = command.concat(['--ignore-file', options.ignoreFile]);
-      }
+      var normalizedOptions = normalizeOptions(options);
 
-      if (options.ignore) {
-        command = command.concat(transformIgnore(options.ignore));
+      if (normalizedOptions.length > 1) {
```
kamilogorek: concatenating an empty array will be a no-op, so we can skip this check

js/index.js
montogeek: ESLint doesn't like that, that is how I did it before, had to change it because ESLint as you can see in the commit history

kamilogorek:
```js
if (typeof options[sourceMapOption] !== 'undefined') {
  if (
    sourceMapOption === 'ignore' ||
    sourceMapOption === 'stripPrefix' ||
    sourceMapOption === 'stripCommonPrefix'
  ) {
    return newOptions.concat(
      transformOption(SOURCEMAPS_OPTIONS[sourceMapOption], options[sourceMapOption])
    );
  } else if (sourceMapOption === 'validate') {
    return newOptions.concat([SOURCEMAPS_OPTIONS[sourceMapOption]]);
  } else {
    return newOptions.concat(
      SOURCEMAPS_OPTIONS[sourceMapOption],
      options[sourceMapOption]
    );
  }
}
return newOptions;
```

montogeek:
Friendly ping @jan-auer @kamilogorek 🤙

kamilogorek:
Looks good to me. @jan-auer?

jan-auer:
Thank you, merging this now!

montogeek:
Thanks @kamilogorek @jan-auer 
When do you usually make new releases?

kamilogorek:
Usually, every few days if we have a lot of features/issues. Right now I think we'll make a release once we merge two remaining PRs.

PR #183, opened 2017-12-08
execute was undefined

roelvanhintum:
uploadSourceMaps lost scope when in the map function


montogeek:
I hit this in my tests too

PR #185, opened 2017-12-08
Replace uchardet with chardet

kirkins:
`uchardet` is [depreciated](https://github.com/emk/rust-uchardet/commit/22d9eb8f2e34e0018d18edf1ab7c2d4edff0e601#diff-04c6e90faac2675aa89e2176d2eec7d8R3) in favor of `chardet`.

With `uchardet` you always need to have `libuchardet-dev` installed, otherwise the compiled rust program will error.

### DIFF
--- a/src/utils/enc.rs
+++ b/src/utils/enc.rs
@@ -13,14 +13,9 @@ pub fn decode_unknown_string(bytes: &[u8]) -> Result<Cow<str>> {
     } else {
         let (enc, confidence, _) = detect(bytes);
         if confidence < 0.5 {
-            fail!("cannot detect language with sufficient confidence");
-        }
-        if_chain! {
-            then {
-                Ok(Cow::Owned(enc))
-            } else {
-                fail!("unknown encoding for string");
-            }
+            fail!("unknown encoding for string");
+        } else {
+            Ok(Cow::Owned(enc))
         }
     }
 }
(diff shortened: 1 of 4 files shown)


jan-auer:
Thanks a lot for the work you put into this (and also your other pull requests), @kirkins. I'd love to merge and release this soon. Replacing `uchardet` has been on our list for a while, so we were very happy when we saw this PR. 

Before merging, we will have to make sure that there are no regressions with the new library. @mitsuhiko do you maybe still have the JS files that were causing the encoding issues?

jan-auer:
Ran some tests and you need to revert back to `confidence >= 0.5`. Unfortunately, latin1 based encodings only report with a confidence of `0.73`. Other than that, fine with me. 

Could you please `git reset --hard HEAD^ && git push origin master`? I'll merge afterwards.

src/utils/enc.rs
```diff
@@ -13,14 +13,16 @@ pub fn decode_unknown_string(bytes: &[u8]) -> Result<Cow<str>> {
     if let Ok(s) = str::from_utf8(bytes) {
         Ok(Cow::Borrowed(s))
     } else {
+        let (label, confidence, _) = detect(bytes);
         if_chain! {
-            if let Ok(enc) = detect_encoding_name(bytes);
-            if let Some(enc) = encoding_from_whatwg_label(&enc);
+            if confidence >= 0.95;
```
jan-auer: Needs to be 0.5, sorry about the confusion.

kirkins:
Ah good catch. I rolled back the commit with 0.95 confidence.

Issue #191, opened 2017-12-20
Add IP Address to User


kirkins:
Our organization is used to seeing Sentry on a front-end application. For events they can always see the IP of the device under user:

![untitled](https://user-images.githubusercontent.com/9503662/34215769-7115368c-e574-11e7-8b1e-1b9a1eb050c5.png)

Where as `sentry-cli` doesn't track this. I was thinking to make a PR fixing this as I think it would be very similar to this line in `events.rs`:

    event.user.insert("username".into(), get_user_name().unwrap_or("unknown".into()));

The problem is all the methods I know of, require pinging (and thus relying on) an external server. For example `curl -s http://whatismyip.akamai.com/`.

The most efficient solution seems like it would be to have the server part of Sentry automatically append IP based on the location of the reporting device.

Does anyone have insights into whether the other Sentry projects record the external IP on the client or server?

kirkins:
This crate might make the process easier if we want to add it on the client side:

https://github.com/equalsraf/whatsmyip

jan-auer:
This has been released in 1.27.0

PR #193, opened 2017-12-20
Get users IP address before sending event

kirkins:
This sends a `curl` request before logging events. 

I have compiled and tested _(on linux)_:

untitled

This requires that user has `curl` installed. I know `libcurl` is a dependency that means `curl` should be available in bash right?

https://github.com/getsentry/sentry-cli/issues/191

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -132,6 +132,7 @@ impl Event {
         }).collect()));
 
         event.user.insert("username".into(), get_user_name().unwrap_or("unknown".into()));
+        event.user.insert("ip_address".into(), String::from("{{auto}}"));
 
         let mut device = HashMap::new();
         if let Some(hostname) = get_hostname() {


kirkins:
I see libcurl is used during build for the curl crate. So I should use that directly instead of calling shell and command prompt.

I can make the change but I will wait for feedback.

If there is a sentry endpoint to get IP that might easier then using the 4 random URLs.

kirkins:
I started searching Google for information on using curl with Rust.

When I started reading an article it mentioned Sentry, then I realized the article was by @mitsuhiko.

http://lucumr.pocoo.org/2016/7/10/rust-rest/

Very interesting. When I search the code base on Github I don't see any instances of "request(" is the function not used anymore? 
 
I'm having some trouble using the function here is the diff for what I'm trying:

https://github.com/getsentry/sentry-cli/compare/master...kirkins:request_function?expand=1

I'm getting back:

```
141 |         let ip = api.request(Api::Method::Get, url);
    |                              ^^^^^^^^^^^^^^^^ ambiguous associated type
```
Am I using it wrong? Still learning the basics on Rust.

mitsuhiko:
Sentry should automatically report the IP address of the client if the value is set to `{{auto}}`. That would simplify all of this.

kirkins:
What value?

mitsuhiko:
If you set the value of `ip_address` in the user to `"{{auto}}"` it should be set to the right value automatically.

kirkins:
@mitsuhiko it doesn't work on the server side:

untitled

alex-hofsteede:
not filling in the IP is a bug on the server side https://github.com/getsentry/sentry/pull/6818 will fix it

mitsuhiko:
Looks like a recent regression in Sentry. This is going to be fixed with this PR on the server: https://github.com/getsentry/sentry/pull/6818

PR #194, opened 2017-12-21
Check for model and family

kirkins:
This adds the enhancement https://github.com/getsentry/sentry-cli/issues/188.

I'm testing locally and it works. I tried follow the style in config.rs.

_Edit: combined commits and removed extra info to make easier for reviewers._

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,7 +12,7 @@ use url::Url;
 use ini::Ini;
 
 use prelude::*;
-use constants::{DEFAULT_URL, DEFAULT_FAMILY, VERSION, PROTOCOL_VERSION};
+use constants::{DEFAULT_URL, VERSION, PROTOCOL_VERSION};
 
 /// Represents the auth information
 #[derive(Debug, Clone)]
@@ -284,7 +284,7 @@ impl Config {
         } else if let Some(val) = self.ini.get_from(Some("device"), "family") {
             Some(String::from(val))
         } else {
-            Some(DEFAULT_FAMILY.to_owned())
+            None
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,7 +12,7 @@ use url::Url;
 use ini::Ini;
 
 use prelude::*;
-use constants::{DEFAULT_URL, VERSION, PROTOCOL_VERSION};
+use constants::{DEFAULT_URL, DEFAULT_FAMILY, VERSION, PROTOCOL_VERSION};
 
 /// Represents the auth information
 #[derive(Debug, Clone)]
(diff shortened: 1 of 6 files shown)


kirkins:
Ok I tried to follow the style in `config.rs`, let me know what you think.

I guess we might want to add some defaults to `constants.rs` but I'm not sure what the default should be. Something like "Linux device"?

src/config.rs
```diff
@@ -375,6 +401,25 @@ fn get_default_auth(ini: &Ini) -> Option<Auth> {
...
+    }
+}
+
+fn get_default_family(ini: &Ini) -> String {
+    if let Some(ref val) = env::var("DEVICE_FAMILY").ok() {
+        val.to_owned()
+    } else if let Some(val) = ini.get_from(Some("defaults"), "family") {
+        val.to_string()
+    } else {
+        DEFAULT_FAMILY.to_owned()
```
kirkins: The above 2 cases work but not the constant `DEFAULT_FAMILY` here. I'm not sure why.

src/utils/system.rs
```diff
@@ -217,5 +217,5 @@ pub fn get_model(config: &Config) -> Option<String> {
 
 #[cfg(not(target_os="macos"))]
 pub fn get_family(config: &Config) -> Option<String> {
-    config.get_family()
+    Some(config.family.to_owned())
```
kirkins: This fixes my earlier comment. But the fact that `config.get_family()` and `config.family` don't always return the same value probably means I'm doing this wrong.

src/constants.rs
```diff
@@ -9,6 +9,12 @@ pub const APP_INFO: &'static AppInfo = &AppInfo {
 
 /// The default API URL
 pub const DEFAULT_URL: &'static str = "https://sentry.io/";
+/// The default device family
+#[cfg(windows)]
+pub const DEFAULT_FAMILY: &'static str = "Windows device";
+#[cfg(not(windows))]
+pub const DEFAULT_FAMILY: &'static str = "Unix device";
```
jan-auer: These default values do not really make sense as this information is already present in the `"os"` field. Just remove the defaults completely please.

src/utils/system.rs
```diff
@@ -190,14 +192,14 @@ pub fn get_model() -> Option<String> {
 }
 
 #[cfg(target_os="macos")]
-pub fn get_family() -> Option<String> {
+pub fn get_family(config: &Config) -> Option<String> {
```
jan-auer: It would be great if the macOS implementation also tries to read the overrides from `config` and then falls back to the current implementation if not set. Same goes for `get_model`.

src/config.rs
```diff
@@ -264,6 +266,28 @@ impl Config {
...
+    }
+
+    /// Return device family
+    pub fn get_family(&self) -> Option<String> {
+        if env::var_os("DEVICE_FAMILY").is_some() {
+            env::var("DEVICE_FAMILY").ok()
+        } else if let Some(val) = self.ini.get_from(Some("device"), "family") {
+            Some(String::from(val))
+        } else {
+            Some(DEFAULT_FAMILY.to_owned())
```
jan-auer: Just return `None` here.

jan-auer:
Great, thanks for the update. Merging this now. 🎉 

Just as a suggestion, you could've also used this syntax which is pretty neat for situations like this:

```rust
if let Some(model) = config.get_model() {
  return model;
}

...
```

kirkins:
@jan-auer thanks that does look better.

Feel free edit my PRs to improve to improve style. I'm learning as I go so feedback is great.

kirkins:
Just wanted to note related to this. By default Sentry GUI shows "undefined" for `sentry-cli` that hasn't set the model:

untitled


https://github.com/getsentry/sentry/issues/6830

jan-auer:
Thanks for mentioning that. In near future we are planning to rework this part of the UI to be a bit more dynamic wrt missing data. Up until very recently this was only used by our iOS and ReactNative SDKs which always included that data.

PR #197, opened 2017-12-27
js: Add support for setting binary's path.

roughwin:
usage:
```
export SENTRY_LOCAL_DOWNLOAD_URL=https://my.cdn.com/ss-static/binary/sentry/2.6.1
yarn
```


jan-auer:
Hi @roughwin , thanks for submitting this pull request. This is actually a great idea and we chose to extend your solution a little bit. Therefore this was closed by #201

roughwin:
good

PR #200, opened 2018-01-02
Use Info.plist from build with xcode

### LINKED ISSUE #198: Dynamic app version with react-native
Trying out `react-native-sentry` now on our project. While building in xcode it fails with `Could not parse Info.plist file`. We set the version number based on git tag in an xcode build phase (before react-native bundling), but I guess the sentry-cli isn't reading from the "build" Info.plist, but from the "project" Info.plist so it fails.

This is how we set version number in an xcode build phase:
```
defaults write "${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH%.*}" "CFBundleShortVersionString" "${VERSION}"
defaults write "${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH%.*}" "CFBundleVersion" "${BUILD}"
``` 

Also, to work with codepush the Info.plist will not contain the version number so we would need some additional argument to specify what the version number is.

adbl:
Fixes #198 

Tested for the use case described in the issue. I have every reason to believe this doesn't affect the existing functionality since xcode copies all files to the build directory, but didn't test this. I believe this is the recommended way to look at `Info.plist` during build. See e.g. https://github.com/facebook/react-native/blob/master/scripts/react-native-xcode.sh#L99 (but they use another variable to get the directory 🤔).

Also never touched rust before so feel free to make it more rusty 😉

### DIFF
--- a/src/utils/xcode.rs
+++ b/src/utils/xcode.rs
@@ -179,13 +179,16 @@ impl InfoPlist {
         // if we are loaded directly from xcode we can trust the os environment
         // and pass those variables to the processor.
         if env::var("XCODE_VERSION_ACTUAL").is_ok() {
-            let vars: HashMap<_, _> = env::vars().collect();
-            if let Some(filename) = vars.get("INFOPLIST_FILE") {
-                let base = vars.get("PROJECT_DIR").map(|x| x.as_str()).unwrap_or(".");
-                let path = env::current_dir().unwrap().join(base).join(filename);
-                Ok(Some(InfoPlist::load_and_process(&path, &vars)?))
-            } else {
-                Ok(None)
+            match (
+                env::var("BUILT_PRODUCTS_DIR"),
+                env::var("INFOPLIST_PATH")
+            ) {
+                (Ok(dir), Ok(filepath)) => {
+                    let path: PathBuf = [dir, filepath].iter().collect();
+                    let vars: HashMap<_, _> = env::vars().collect();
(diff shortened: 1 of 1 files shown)


adbl:
Changed to `TARGET_BUILD_DIR` as it seems more appropriate even though they seem to contain the same value.

https://developer.apple.com/legacy/library/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW88

HazAT:
@mitsuhiko LGTM and it also makes sense to use the plist of the built app instead of the project which could be old.
Just need a rust code check (I guess it's fine) 😅

HazAT:
Thx @adbl 🍻

jan-auer:
@adbl FYI: We had to revert this as it started causing issues when `Preprocess Info.plist File` was set to `YES`. In the long run, we want to add this again, but for now it's easier to roll back to the old mechanism.

cc @HazAT

adbl:
Oh bummer, is there an issue or any more information about this problem?
Did the file not exist there if this flag set?

jan-auer:
The issue and follow-up PR are #226 and #232. 

Basically, we need to support sentry-cli running as an Xcode build step as well as outside of XCode. In order to support both cases reliably, we will have to split handling of the Info.plist and Symbol upload in two different steps.

Right now, we are quite busy with other issues, but we’d like to deal with this in February.

PR #219, opened 2018-01-16
ref: Remove explicit AsciiExt usage

jan-auer:
Since rust 1.23.0 this is no longer necessary and there's not really a point in supporting old rust versions for `sentry-cli`.


jan-auer:
Cancelled macOS jobs since the Travis queue is too long for them to be ever executed.

PR #221, opened 2018-01-16
build: Build a proper musl target

jan-auer:
This builds `sentry-cli` properly for musl using the [clux/muslrust](https://github.com/clux/muslrust) docker image. The image is rather large and builds take a while (almost 8 minutes on my machine), but the result is promising:

- **Before:** file size *18.1M* and requires `libcurl`, `llvm-libunwind`, and `libgcc` in alpine
- **After:** file size *15.6M* and requires no dependencies in alpine

Once we refactor our linux builds we can use this image to build and release a `x86_64-unknown-linux-musl` target directly to the CDN/GitHub.


jan-auer:
@mitsuhiko This does not build on Windows because of mitsuhiko/clicolors-control#4 and mitsuhiko/console#4. Once you release new versions there, this should build everywhere again.

jan-auer:
@mitsuhiko seems like this works (though macOS tests will take some more time). Mind if we merge this and trigger the next release?

PR #229, opened 2018-01-19
feat(upload): Implement Linux symbol upload

jan-auer:
This PR adds a command to upload ELF symbols, both from executables / libraries and from stripped `*.dbg` files:

```
sentry-cli-upload-sym
Upload Linux debug symbols to a project.

USAGE:
    sentry-cli upload-sym [OPTIONS] [--] [PATH]...

OPTIONS:
    -h, --help                 Prints help information
        --no-executables       Exclude executables and look for stripped symbols only.
        --no-reprocessing      Do not trigger reprocessing after uploading.
    -o, --org <ORG>            The organization slug
    -p, --project <PROJECT>    The project slug
        --require-all          Errors if not all UUIDs specified with --uuid could be found.
        --uuid <UUID>...       Search for specific UUIDs.

ARGS:
    <PATH>...    A path to search recursively for symbol files.
```

Other changes include:
 - Refactored common code from `upload-breakpad` to `utils::upload::*`.
 - Update wording for `upload-dsym` to make it more clear

**NOTE:** `upload-dsym` and `upload-proguard` are still separate. Also, `difutil` uses a different file walker implementation. I think it would be good to consolidate these in near future to avoid unintended different behavior.


src/utils/upload.rs
```diff
@@ -0,0 +1,445 @@
...
+    info!("Checking for missing debug symbols: {:#?}", &batch);
+
+    let missing_checksums = {
+        let checksums = batch.iter().map(|ref s| s.checksum.as_str()).collect();
+        context.api().find_missing_dsym_checksums(&context.org(), &context.project(), &checksums)?
+    };
+
+    let missing = batch
+        .into_iter()
+        .filter(|sym| missing_checksums.contains(&sym.checksum))
```
jan-auer: While rewriting I found that `upload-breakpad` code previously did not deduplicate identical checksums. For Breakpad and ELF, this doesn't matter as each object is in its own file. However, to support Mach-O fat objects, we need to either change this here or change the way objects are Zipped completely.

Issue #228, opened 2018-01-18
Install Script Fails When Behind Proxy


edcs:
The `bin/install-sentry-cli` script uses Node's HTTP and HTTPS libraries to download the Sentry CLI binary. Node.js doesn't have a way of globally applying proxy settings meaning that you need to modify the install script to apply them - this isn't very practical because there's probably lots of people who aren't trapped behind a corporate proxy.

Would it be feasible to change this script to use `curl` if available to do the download? Or is there another way to grab the file? I'm using Sentry in a Rect Native project and trying to install using Yarn.

edcs:
Or as an alternative, could the script skip installation if the Sentry CLI has already been installed via Homebrew or another alternative method?

jan-auer:
Hi @edcs and thanks for bringing this up! To be fair, our installation script has been pretty rudimental so far and I guess it's time to change this now.

Proxy: Happy to accept a PR for this. However, it would be good to find a solution that ideally works the same on Windows, Linux and macOS. After reading up on this a little bit, you must be having the same issue with `npm` itself as well. Do people behind proxies usually configure npm's proxy setting? If so, we could leverage this, migrate to [node-fetch](https://npmjs.org/package/node-fetch) and add the proxy in the download script like [here](https://github.com/bitinn/node-fetch/issues/79#issuecomment-184594701). That way, all systems are supported, even Windows and those without curl.

Homebrew: This is a good idea, but I'm looping in @HazAT here for a second opinion. In general, I'd be fine if you **copy** the exectuable to the right location iff the version matches exactly. Even though we try to not break between major releasees, we had to make some adaptions in the past and do not test all version combinations when we push out a release. From my side, PR for this is more than welcome anytime.

edcs:
Hi @jan-auer - To get Yarn or NPM to work behind a proxy, you set the config options you linked to and they just work. Unfortunately Node has no such setting.

You can set a proxy agent in `http` and `https`, `node-fetch` seems to use the same configuration. I did experiment with this idea yesterday but couldn't get it to work, hence why I thought CuRL would be a good alternative. However, if I can get Node configured properly that's possibly the best option. My idea was to include the agent if `HTTP_PROXY`/`HTTPS_PROXY` were defined in the environment.

jan-auer:
[...]

PR #230, opened 2018-01-22
ref(npm): Add support for proxies in the NPM installer

edcs:
When installing the CLI with NPM or Yarn behind a proxy server, Node.js fails to complete the download because no proxy is configured. This PR adds in `https-proxy-agent` (https://www.npmjs.com/package/https-proxy-agent) as a dependancy and updates the HTTPS request options so that the download can succeed.

Please could someone who is also working behind a proxy test this? I believe it to be working, but for some reason the final download from AWS `401`s. I think this is because of the proxy configuration at my end as the changes I've made are as per the documentation of the agent package.

References: https://github.com/getsentry/sentry-cli/issues/228


jan-auer:
Awesome, thanks for putting this together. The code looks good. 

I'll try to perform a test myself a little later. The 401 is indeed strange since it's not even mentioned in the S3 documentation.

jan-auer:
There seems to be some bug in `http-proxy-agent`. I wasn't able to reproduce `401 Unauthorized` but keep getting a `403 Forbidden`. It happens only when going through the HttpsProxyAgent, but not when using the same proxy via `curl`.

edcs:
Ok sure, I'll look at an alternative solution for this one. I think there's other packages/methods for defining a proxy server in Node.

jan-auer:
@edcs thanks for putting so much time in this. I was finally able to spend some time debugging the error. Turns out `https-proxy-agent` was indeed causing a not expected `Host` header in the target request. Created a fix for this here: https://github.com/TooTallNate/node-https-proxy-agent/pull/43

Besides, I think we can drop support for Node 0.12 and migrate to `node-fetch` to remove the custom (and incomplete) redirect logic. I played around with it a little bit and ended up with something like:

```js
  var HttpsProxyAgent = require('https-proxy-agent');
  var fetch = require('node-fetch');
  var getProxyForUrl = require('proxy-from-env').getProxyForUrl;

  var proxyUrl = getProxyForUrl(downloadUrl);
  var agent = proxyUrl ? new HttpsProxyAgent(proxyUrl) : null;
  fetch(downloadUrl, { redirect: 'follow', agent: agent }).then(function(response) {
    if (response.status < 200 || response.status >= 300)
      throw new Error('Received ' + response.status + ': ' + response.statusText);

    response.body.pipe(fs.createWriteStream(outputPath, { autoClose: true, mode: '0755' }));
  });
```

This obviously doesn't include the progress bar yet, but already demonstrates how much simpler the code gets. Let me know if you'd like me to edit your PR or do it yourself. In any case I will clean up (i.e. refactor to ES6) and update tests afterwards, so no need for you to spend too much effort on this.

edcs:
Amazing - thanks for picking this up! I've not had time to work on it recently. Should be able to do some work on it following from your feedback over the next few days :)

jan-auer:
@edcs Heads up, I took the liberty to rewrite the installer based on our discussion since I had some time. Let me know what you think and feel free to edit. 

Also looping in @HazAT for a second opinion. FYI - the `server.close()` is now refactored into a process `exit` event so that we can be sure it always gets called. 

Let's also keep in mind that we will need to wait for TooTallNate/node-https-proxy-agent#43 until we can merge this.

HazAT:
@jan-auer Good job!
Looks much cleaner now. 
Fix CI then I'll hit approve!

jan-auer:
Ugh, just noticed that I killed the `--version` check after downloading

jan-auer:
@HazAT updated. Should have tried it at least once before committing :)

When implementing the version-check I noticed [this `console.log`](https://github.com/getsentry/sentry-cli/blob/master/js/helper.js#L78) - is it really necessary?

jan-auer:
@edcs could you please confirm that this works in your setup now?

edcs:
That's working for me now! 🎉 

screen shot 2018-01-30 at 09 03 31

jan-auer:
Thank you so much for the bug report and the PR!

Issue #226, opened 2018-01-17
Error parsing Info.plist "Invalid Data"


gilesvangruisen:
Hi there! Thanks for this project, setup process is pretty smooth. I am, however, seeing an issue with the dSYM upload script phase. The script runs fine, but my Info.plist file fails to parse, with the reason "invalid data".

My info.plist is not terribly crazy. I had one env var in there that I removed, but the problem seems to persist. Here's a somewhat obfuscated info.plist (nothing important missing):

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>CFBundleDevelopmentRegion</key>
	<string>en</string>
	<key>CFBundleDisplayName</key>
	<string>Acme</string>
	<key>CFBundleExecutable</key>
	<string>$(EXECUTABLE_NAME)</string>
	<key>CFBundleIdentifier</key>
	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
	<key>CFBundleInfoDictionaryVersion</key>
	<string>6.0</string>
	<key>CFBundleName</key>
	<string>$(PRODUCT_NAME)</string>
	<key>CFBundlePackageType</key>
	<string>APPL</string>
	<key>CFBundleShortVersionString</key>
	<string>1.0</string>
	<key>CFBundleSignature</key>
	<string>????</string>
	<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>obfuscated FB token</string>
			</array>
		</dict>
	</array>
	<key>CFBundleVersion</key>
	<string>25</string>
	<key>Fabric</key>
	<dict>
		<key>APIKey</key>
		<string>obfuscated fabric token</string>
		<key>Kits</key>
		<array>
			<dict>
				<key>KitInfo</key>
				<dict/>
				<key>KitName</key>
				<string>Crashlytics</string>
			</dict>
		</array>
	</dict>
	<key>FacebookAppID</key>
	<string>obfuscated FB app ID</string>
	<key>FacebookDisplayName</key>
	<string>obfuscated FB display name</string>
	<key>ITSAppUsesNonExemptEncryption</key>
	<false/>
	<key>LSApplicationQueriesSchemes</key>
	<array>
		<string>fbapi</string>
[...]

PR #232, opened 2018-01-24
[WIP] fix: Xcode

HazAT:
Don't build plist file by our self, use Xcode result instead.

Fixes https://github.com/getsentry/sentry-cli/issues/226
Fixes https://github.com/getsentry/react-native-sentry/issues/333

thx @jan-auer


jan-auer:
Quick update on this: The above fix does not use the correct target build directory if the project was built in Xcode. 

The correct way would be to move out the preprocessed plist or extracted metadata to a well-known location and then load it from there when uploading dsyms. This would essentially be the first step to a “release archive”

PR #235, opened 2018-01-31
ref(js): Use ES6 language level

jan-auer:
Since we dropped support for Node 0.12 recently, we can now use a considerable subset of the ES6 spec. This PR updates the linter and formatter and applies the new style to all files.

Most of the changeset was generated automatically. See comments below for exceptions.

### DIFF
--- a/js/__tests__/helper.test.js
+++ b/js/__tests__/helper.test.js
@@ -1,38 +1,38 @@
 /* eslint-env jest */
 
-var helper = require('../helper');
+const helper = require('../helper');
 
-var SOURCEMAPS_OPTIONS = require('../releases/options/uploadSourcemaps');
+const SOURCEMAPS_OPTIONS = require('../releases/options/uploadSourcemaps');
 
-describe('SentryCli helper', function() {
-  test('call sentry-cli --version', function() {
+describe('SentryCli helper', () => {
+  test('call sentry-cli --version', () => {
     expect.assertions(1);
-    return helper.execute(['--version']).then(function() {
-      expect(true).toBe(true);
-    });
+    return helper
+      .execute(['--version'])
+      .then(version => expect(version.trim()).toBe('sentry-cli DEV'));
   });
 
-  test('call sentry-cli with wrong command', function() {
+  test('call sentry-cli with wrong command', () => {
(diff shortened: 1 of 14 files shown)


js/__tests__/helper.test.js
```diff
@@ -1,38 +1,38 @@
...
-  test('call sentry-cli --version', function() {
+describe('SentryCli helper', () => {
+  test('call sentry-cli --version', () => {
     expect.assertions(1);
-    return helper.execute(['--version']).then(function() {
-      expect(true).toBe(true);
-    });
+    return helper
+      .execute(['--version'])
+      .then(version => expect(version.trim()).toBe('sentry-cli DEV'));
```
jan-auer: Added a check for the output. Note the `trim()` - we could consider trimming in `helper.execute()`

js/helper.js
```diff
@@ -1,84 +1,90 @@
...
+        resolve(stdout);
+      }
+    });
+  });
+}
+
+module.exports = {
+  normalizeOptions,
+  prepareCommand,
+  getPath,
```
jan-auer: In `helper.js` all functions were moved to the top scope so that no more `this` is necessary.

js/index.js
```diff
@@ -1,23 +1,23 @@
...
+const helper = require('./helper');
 
-function SentryCli(configFile) {
-  if (typeof configFile === 'string') process.env.SENTRY_PROPERTIES = configFile;
-}
-
-SentryCli.prototype.getConfigStatus = function() {
-  return helper.execute(['info', '--config-status-json']);
-};
+class SentryCli {
```
jan-auer: The CLI is a `class` now.

js/helper.js
```diff
@@ -1,84 +1,90 @@
...
-        }
-      });
-    });
+
+    return newOptions.concat(paramName, paramValue);
+  }, []);
+}
+
+function prepareCommand(command, commandOptions, options) {
+  return command.concat(normalizeOptions(commandOptions || {}, options || {}));
```
jan-auer: Note how it is no longer necessary to call `this.normalizeOptions`

PR #253, opened 2018-02-15
ref(upload): Split multi-arch binaries before uploading

jan-auer:


jan-auer:
Waiting for https://github.com/getsentry/sentry/pull/7270

PR #254, opened 2018-02-17
feat(upload): Create a unified upload-dif command

jan-auer:
This creates a unified command to upload Debug Information Files (DIFs): `sentry-cli upload-dif`. Its command line options are the same as `upload-dsym`, plus:

```
        --no-debug              Exclude files containing only stripped debugging info.
        --no-bin                Exclude executables and libraries and look for debug symbols only.
    -t, --type <TYPE>...        Only consider debug information files of the given type.  By
                                default, all types are considered. [values: dsym, elf, breakpad]
```

Also, the former `upload-dsym` command now builds completely on `upload-dif` and has no more own source code. It sets the `legacy` flag, which simulates `--type dsym --no-bin`. To promote the new command and prepare an upcoming deprecation, `upload-dsym` also outputs the following notice:

```
Note: Use the upload-dif command for improved options. For more information, run:
$ sentry-cli upload-dif --help
```

Finally, the commands `upload-breakpad` and `upload-dbg` are no longer needed and have been removed. Since they were never documented officially, there is no need for a deprecation period or breakage notice except for select test users.

**Remarks**:
 - We need to update docs when releasing this
 - Since this is a new command, it would make sense to only support the new debug identifiers and no more UUIDs for `upload-dif`.


jan-auer:
As discussed offline: We should call it `dsym` and not `mach`.

PR #266, opened 2018-03-08
fix: stripPrefix is a string

HazAT:


HazAT:
This is a array not a string, there error must be somewhere in our webpack plugin.

PR #276, opened 2018-03-20
ref: Use DebugId in place of UUIDs

jan-auer:
This PR uses `ObjectId` for debug information files in place of the former `UUID`. Since serialization and deserialization are compatible with UUIDs for all cases where UUIDs are currently being used, this change should not break current API behavior.

The `difutil uuid` command has been renamed `difutil id`, but there is a hidden alias to `difutil uuid` to remain backward compatible.

Likewise, the `difutil find --uuid` option has been renamed `--id` with an alias `--uuid`.

From now on, all API responses returning a `DebugInfoFile` can contain a full object ID in the `debug_id` field in addition to a UUID in `uuid` now. These are `<...>/files/dsyms` (legacy archive upload), `<...>/files/difs/assemble` (new chunk assemble), and `<...>/files/dsyms/associate` (build association for dSYMs).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,7 +189,7 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "2.31.1"
+version = "2.31.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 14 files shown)


jan-auer:
Note that we have to release `symbolic` before merging this. Also, this needs more testing.

PR #283, opened 2018-04-06
doc: Add a description for PDB files

jan-auer:
screen shot 2018-04-06 at 15 33 39


jan-auer:
Updated wording to
> Microsoft PDB files are not yet supported directly by Sentry. Until we provide official support, you can convert them to Breakpad symbols and upload **those** instead:

PR #291, opened 2018-04-27
ref: Fix platform warnings about unused imports

tonyo:


.gitignore
```diff
@@ -4,6 +4,7 @@ target
 .cargo
 node_modules
 coverage
+.*.swp
```
jan-auer: That should go in a local .gitignore please.

tonyo: Fixed

Issue #227, opened 2018-01-18
Support Alpine


mrmckeb:
As the CLI is a dependency of https://github.com/getsentry/sentry-webpack-plugin, we've been pulling the binary from your docker container, hosting that, and then passing in our host with the new `SENTRYCLI_CDNURL` environment variable. We've also used Yarn's resolutions feature to pin to the latest release for now.

This is fine, but it makes updates a little more painful, and I think other users are likely to face this in future.

Could the binary from that container be automatically attached to releases here?

CC @TheDen

jan-auer:
Hi @mrmckeb, it's definitely on our roadmap!

TheDen:
@jan-auer 

When using `SENTRYCLI_CDNUR`, we use a script that copies out the `sentry-cli` alpine binary from the sentry's docker hub repo. Perhaps it may be useful for your automated releases.

```#!/bin/bash
# tag is either a specific version or "latest"
tag=1.28.3
docker pull getsentry/sentry-cli:$tag
ID=$(docker create getsentry/sentry-cli:$tag)
docker cp $ID:/bin/sentry-cli .
```

jan-auer:
So, we have recently changed the alpine build and want to completely replace the current linux builds with the alpine one. However, I still haven't looked into cross-building 32bit in the musl container. 

After that, both the NPM package and GitHub releases for linux will work on alpine just like any other currently supported distribution.

TheDen:
awesome 👍

mwarkentin:
Getting sentry-cli added to the alpine package repository at that point would be amazing too... `apk add sentry-cli`

<3

JalilArfaoui:
This seems to block https://github.com/getsentry/sentry-cli/issues/281 and https://github.com/getsentry/sentry-webpack-plugin/issues/44

byCedric:
This is sort of blocking for us too since we moved to lightweight pipeline images. Is there an existing workaround e.g. prebuilt alpine 3.7 binary?

Kamshak:
@byCedric you can do 

in package.json
```
"resolutions": {
    "@sentry/cli": "1.30.4"
  },
```

in your dockerfile/CI
``` 
SENTRYCLI_CDNURL=https://cdn.rawgit.com/Kamshak/packages/0cd95c30 yarn
```

tonyo:
Hey everyone,
Thanks for your input.
[...]

PR #294, opened 2018-05-02
ci: Refactor and optimize Linux CI builds

tonyo:
This PR changes a base image for our Docker builds (Linux only), as a result we get statically-linked binaries for 32 and 64 bits.
I also cleaned up `.travis.yml` and `.ci/` a bit.


tonyo:
Looks like it fixes #227.

PR #298, opened 2018-05-04
feat(releases): Detect HEROKU_SLUG_COMMIT in propose-version

### LINKED ISSUE #289: HEROKU_SLUG_COMMIT not used with releases propose-version
According to the docs here: https://docs.sentry.io/integrations/heroku/#configure-releases it states:

> Whether use the add-on or configure Sentry yourself, you’ll also likely want to grant quick access to your dyno metadata, which will allow Sentry to automatically pick up the git version of your application.
>
> To do this, enable the runtime-dyno-metadata feature:
>
> heroku labs:enable runtime-dyno-metadata -a <app name>
> This exposes the HEROKU_SLUG_COMMIT environment variable, which most Sentry SDKs will automatically detect and use for configuration.

Running `HEROKU_SLUG_COMMIT=e123445 npx sentry-cli releases propose-version` on a exported git repo doesn't result in `e123445` being the proposed name, but rather `error: Could not automatically determine release name`. 

The docs state that "most SDK's" pick this up, I can't find any information outside of the heroku page which discusses this. Please help :)

I'm using Webpack for my setup, I've worked out that I need to do:

```
new SentryCliPlugin({
  configFile: "sentry.properties",
  include: cwd + "/dist/",
  release: process.env.HEROKU_SLUG_COMMIT,
  ignore: ["node_modules"]
})
```

jan-auer:
This detects `HEROKU_SLUG_COMMIT` in `sentry-cli releases propose-version`,
as users would expect: https://docs.sentry.io/integrations/heroku/#configure-releases

Fixes #289

### DIFF
--- a/src/utils/releases.rs
+++ b/src/utils/releases.rs
@@ -88,6 +88,13 @@ pub fn detect_release_name() -> Result<String, Error> {
         return Ok(release);
     }
 
+    // try heroku: https://docs.sentry.io/integrations/heroku/#configure-releases
+    if let Ok(release) = env::var("HEROKU_SLUG_COMMIT") {
+        if !release.is_empty() {
+            return Ok(release);
+        }
+    }
+
     // for now only execute this on macs.  The reason is that this uses
     // xcodebuild which does not exist anywhere but there.
     if_chain! {


cyrilchapon:
Any chance we can also support `SOURCE_VERSION` as this is the [heroku native way](https://devcenter.heroku.com/changelog-items/630) to detect version (without the need of a buildpack)

jan-auer:
This would then need to be by the respective SDK(s) as well, which is ruby AFAIK. I will try to have a look at what this would take some time soon.

cyrilchapon:
> This would then need to be added to the respective SDK(s) as well

Why ? Do SDK's also include some release auto finding ?

> which is ruby AFAIK

I don't think so. I think this is common to every Heroku apps

jan-auer:
> Do SDK's also include some release auto finding ?

Yes, they have to so the events they are sending match the release created with sentry-cli. You can see an example [here](https://github.com/getsentry/raven-ruby/blob/34c0fbc19ed662ca306ba75241e0fa1e0a9c736b/lib/raven/configuration.rb#L355). 

> I think this is common to every Heroku apps

Yes, you are right. However, at the moment only our ruby and python SDKs support reading this environment variable. You can of course specify the release manually in every other SDK.

cyrilchapon:
Hehe right 😁 I was asking because I'm a js guy and we're forced to kinda reproduce the same mechanism use by Webpack-plugin in our frontend build (raven call) in order to match the same release for sending sources and poping events (using sentry-cli from code, which works great btw)

In the meantime, I was able to activate Heroku dyno metadata, but the Webpack plugin is still not able to parse it apparently I'm getting errors. I'll digg into it and reopen elsewhere if needed.

In the meantime, I still think SOURCE_VERSION is a good idea. I suck at ruby and python but I'll give a try to PR something

jan-auer:
FWIW, the webpack plugin injects `global.SENTRY_RELEASE` into your bundle which you can refer to in your javascript source code. But you'd still need to call `sentry-cli releases propose-version` and put that in your other code. Or you simply use `release: process.env.SOURCE_VERSION` when configuring your SDK(s).

Anyway, PRs welcome, although I'm not sure how much we actually want to support in the SDKs. I think for best compatibility, it would be better to infer the release name at build time and then configure it explicitly as described above.

cyrilchapon:
I agree with the idea of challenging the whole concept of sentry release retrieval dynamically from source code (especially since in frontend, the runtime may not be the env than build time, and we need to refer to build time release). We build something in a staging env, and promote the compiled slug to production in here; the sentry release is therefore the staging one, and must stick to it.

Actually, we build a "process.env"-like for frontend, through Webpack, using sentry-cli along with the plugin, and use process.env.something inside our app, to ensure consistency.

However, since something is already implemented, both here and in SDK's, regarding Heroku release retrieval, I'd vote for upgrading it to be SOURCE_VERSION aware by design (without killing HEROKU_SLUG_COMMIT).

I'd also argue that SDKs matter is not the same concern as sentry-cli one, and finally I'd probably vote for a cross platform, cross language externalization of the "proposeRelease" tool in another separate open source repo

PR #306, opened 2018-05-22
Use installation method specific update prompts

### LINKED ISSUE #305: Give correct upgrade nag message when installed via homebrew or NPM
If Sentry CLI is installed via Homebrew or NPM, then running any command will return 

```
$ sentry-cli info
Sentry Server: https://sentry.io
...

sentry-cli update to 1.31.0 is available!
run sentry-cli update to update
```

If you then run `sentry-cli update` you get:

```console
$ sentry-cli update
This installation of sentry-cli is managed through homebrew
Please use homebrew to update sentry-cli:

  $ brew upgrade sentry-cli

sentry-cli update to 1.31.0 is available!
run sentry-cli update to update
```

Instead of the nag message saying "run sentry-cli update to update" for all install types, under NPM or Homebrew it should probably say something like "run brew upgrade sentry-cli to update".

https://github.com/getsentry/sentry-cli/blob/e79e68d1933dc679ec48351b59b35eab18d14491/src/utils/update.rs#L236

danielcompton:
I wasn't quite sure what to do about the `is_npm_install` branch. It looks like it has been overridden in the update nagger, which would mean that it would never run here.

Fixes #305

### DIFF
--- a/src/utils/update.rs
+++ b/src/utils/update.rs
@@ -153,7 +153,7 @@ impl SentryCliUpdateInfo {
         }
         if is_npm_install() {
             println!("This installation of sentry-cli is managed through npm/yarn");
-            println!("Please use npm/yearn to update sentry-cli");
+            println!("Please use npm/yarn to update sentry-cli");
             return Err(QuietExit(1).into());
         }
         if self.latest_release.is_none() {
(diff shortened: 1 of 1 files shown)


src/utils/update.rs
```diff
@@ -233,7 +233,13 @@ fn update_nagger_impl() -> Result<(), Error> {
                 check.latest_version()
             )).yellow()
         );
-        println_stderr!("{}", style("run sentry-cli update to update").dim());
+        if is_homebrew_install() {
+            println_stderr!("{}", style("run brew upgrade sentry-cli to update").dim());
+        } else if is_npm_install() {
+            println_stderr!("{}", style("Please use npm/yarn to update sentry-cli").dim())
```
danielcompton: Currently this line will never actually run because of https://github.com/getsentry/sentry-cli/pull/306/files#diff-5f099b4f4eb2e9d89a0a2ed0f393f9c2R267. Either this branch should be removed, or the `is_npm_install` noop below.

PR #315, opened 2018-05-31
feat(api): Support gzipped chunk upload

jan-auer:
Debug Information File upload has been pretty slow after we've introduced the
new chunk upload since we did not compress uploads anymore. This PR introduces
compression into the chunk upload if the server can handle it.


jan-auer:
@mitsuhiko this should be final now

PR #316, opened 2018-06-01
ref(event): Use sentry to send events

jan-auer:
Switches implementation of `sentry-cli send-event` to the `sentry` crate. This comes with a couple of differences in the events sent to sentry now:

 - `contexts.device.name` is no longer set to the hostname. The `server_name` tag is still included. This is due to the way, `sentry` constructs the default device context.
 - `contexts.rust` now contains the Rust version used to compile sentry-cli, which was not present before
 - `sdk_info.version` was incorrectly set to the build architecture and now contains the sentry-cli version.


jan-auer:
@mitsuhiko This is ready for review now. Will still run some manual tests, though.

jan-auer:
@mitsuhiko Updated. Please also see the updated PR description.

PR #321, opened 2018-06-12
fix(api): Generate proper errors for project renames

jan-auer:
```
$ sentry-cli releases -o sentry-test -p rust2 list
error: project not found
  caused by: project was renamed to 'rust'
```


HazAT:
🎉

Issue #218, opened 2018-01-15
Add support for App Center CLI in addition to CodePush


bslayerw:
```$ sentry-cli --version
sentry-cli 1.28.1
```

Following the documentation located here: https://docs.sentry.io/clients/react-native/codepush/. I cannot get sentry-cli to work without me specifying parameters explictly. They don't seem to be picked up from the sentry.properties file I specify with `ENTRY_PROPERTIES=./ios/sentry.properties`.

When I specify the parameters I get the following error: 

```


sentry-cli react-native codepush --org myorg --project myproject  myCodePushProjectName ./build
> Fetching latest code-push package info
error: No such file or directory (os error 2)
```

jan-auer:
Hi @bslayerw, 

 - Is this just a typo here or did you declare the environment variable as `ENTRY_PROPERTIES`? 
 - Please make sure that the `code-push` binary is on the `PATH`. Assuming it is somewhere in `node_modules`, we will update `sentry-cli` to check there, too.
 - FWIW, the command you posted here is missing the platform argument before the path (`android` or `ios`). But judging by the output you used it when you ran locally.

bslayerw:
@jan-auer code push is definetly in my path or I would not get the following output `> Fetching latest code-push package info`. Here's another command I tried with the same error: 

`SENTRY_PROPERTIES=./ios/sentry.properties ./node_modules/@sentry/cli/bin/sentry-cli react-native codepush  myCodePushProjectName ios ./build`

`> Fetching latest code-push package info
error: No such file or directory (os error 2)`

bslayerw:
okay I think I figured it out. Code push is probably not in my path because I'm using the new appcenter cli. I guess sentry-cli doesn't support the new appcenter cli yet? Making a note here for others who might bump into this.

jan-auer:
Sorry, did not mean to close this via the issue. You are correct, we do not support the App Center CLI just yet. Sorry for the inconvenience and let us keep this open until we've added support for App Center.

In the meanwhile, the linked PR (#220) tries to improve the error messages to make this more clear. "Fetching latest code-push package info" simply meant that `sentry-cli` is trying to invoke codepush.

HAMPLENI:
Same situation for me.  I would prefer to use the App Center CLI except I cannot.  When I try to use the CodePush CLI, I still have not been able to resolve the error unlike bslayerw.

* OS: Windows 10
* Using nvm-windows
* npm@3.10.10
* node@6.11.1
[...]

PR #325, opened 2018-06-18
Switch from codepush to App Center CLI

ianjsikes:
The `sentry-cli react-native codepush ...` command currently uses the `codepush` CLI, which is deprecated in favor of the App Center CLI. This PR moves over to the new App Center CLI.

The main difference is the output of the CLIs. App Center's output for `deployment list ...` is something like this:
```json
[["Staging", "123abc"], ["Release", "456def"]]
```
So I mapped that data into the array of structs that the old CodePush CLI outputted.

I'm not super familiar with Rust so if my code is ugly or not idiomatic let me know and I'll try to clean it up.

Addresses #218

### DIFF
--- a/src/utils/codepush.rs
+++ b/src/utils/codepush.rs
@@ -12,8 +12,8 @@ use serde_json;
 use utils::releases::{get_xcode_release_name, infer_gradle_release_name};
 use utils::xcode::{InfoPlist, XcodeProjectInfo};
 
-static CODEPUSH_BIN_PATH: &'static str = "code-push";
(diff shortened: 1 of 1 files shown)


jan-auer:
Sorry, we had to roll this back since I realized that appcenter CLI works a bit differntly. The second string in the response is not the package label, but just the deployment key. We will put together a follow-up PR. Still, many thanks for submitting your PR!

jan-auer:
See https://github.com/getsentry/sentry-cli/pull/327

PR #327, opened 2018-06-21
feat(react_native): Add a command for appcenter-cli

jan-auer:
Introduces the command `sentry-cli react-native appcenter` to manager codepush deployments with AppCenter CLI. It uses `appcenter codepush deployment history` to get a history of recent releases and fetches the package label from the last entry in the history.

This also deprecates the `sentry-cli react-native codepush` command and hides it in the command overview. However, it will still work as expected.

The appcenter command is a simple copy/paste from the old codepush command with changed strings and better help messages. Apart from `get_appcenter_package`, the code is identical.

Since we only require a deployment's package label, I've implemented a custom visitor that discards all other values returned by AppCenter CLI. This theoretically also makes us more robust against changes in the output format, if more parameters are added in the future. We could switch to tuple deserialization if we need to extract more info in the future.

Ref https://github.com/getsentry/sentry-cli/issues/218

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -69,6 +69,7 @@ pub mod upload_dsym;
 pub mod upload_proguard;
 
 pub mod react_native;
+pub mod react_native_appcenter;
 pub mod react_native_codepush;
 pub mod react_native_gradle;
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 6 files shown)


jan-auer:
@HazAT could you test this with a new version of the react-native SDK please? Once done, we can release both at the same time.

PR #342, opened 2018-07-16
fix: Prevent freeze during process exit after symbol upload

jan-auer:
Our `thread_local` shared API client (`Api::get_current()`) is not properly cleaned up for spawned threads on older Windows versions. This only occurs during DIF upload, when we spawn multiple threads to upload concurrently.

This PR switches the threadpool implementation to _rayon_ and manually replaces the curl handle 
(causing the old one to drop and clean up) when exiting each spawned thread. The new handle is never initialized.

Fix tested with Windows Server 2008 R2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -298,11 +298,34 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "crossbeam-deque"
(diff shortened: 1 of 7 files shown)


jan-auer:
@mitsuhiko This is ready for final review now.

PR #353, opened 2018-08-09
fix: Reset shared handle on Drop in API

jan-auer:


jan-auer:
@mitsuhiko works now 🤷‍♂️

PR #363, opened 2018-08-29
fix(dif_upload): Also show errors from previous uploads

jan-auer:
With this change, sentry-cli will always show errors for DIFs during the
assemble poll, even if all their chunks had already been uploaded. Previously,
such an upload would just show:

```
> Nothing to upload, all files are on the server
```

This made it hard to track and reproduce DIF upload and assembly errors.

### DIFF
--- a/src/utils/dif_upload.rs
+++ b/src/utils/dif_upload.rs
@@ -809,7 +809,7 @@ fn process_symbol_maps<'a>(
 fn try_assemble_difs<'data>(
     difs: &'data [ChunkedDifMatch<'data>],
     options: &DifUpload,
-) -> Result<Option<MissingDifsInfo<'data>>, Error> {
+) -> Result<MissingDifsInfo<'data>, Error> {
     let api = Api::get_current();
     let request = difs.iter().map(ChunkedDifMatch::to_assemble).collect();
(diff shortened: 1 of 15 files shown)


jan-auer:
@mitsuhiko Most of the changes are just rustfmt, for a good diff just look at 4af890e.

src/utils/dif_upload.rs
```diff
@@ -1101,16 +1105,14 @@ fn upload_difs_chunked(
...
 
-    // Upload until all chunks are present on the server
-    let mut initially_missing = None;
-    while let Some(missing_info) = try_assemble_difs(&chunked, options)? {
-        upload_missing_chunks(&missing_info, chunk_options)?;
-        initially_missing.get_or_insert(missing_info);
-    }
+    // Upload missing chunks to the server and remember incomplete difs
+    let missing_info = try_assemble_difs(&chunked, options)?;
+    upload_missing_chunks(&missing_info, chunk_options)?;
```
jan-auer: We previously looped here to be defensive, but there is no real reason. Since we're uploading all chunks, there won't be any missing afterwards. Upload errors will exit immediately in both cases.

Issue #371, opened 2018-09-06
error: For non git repositories explicit revisions are required


tjg37:
Trying to setup releases as described in docs docs: https://docs.sentry.io/learn/releases/

Following the steps: 

```
# Assumes you're in a git repository
export SENTRY_AUTH_TOKEN=...
export SENTRY_ORG=my-org
VERSION=$(sentry-cli releases propose-version)

# Create a release
sentry-cli releases new -p project1 -p project2 $VERSION

# Associate commits with the release
sentry-cli releases set-commits --auto $VERSION
```

All works well except for the last command `sentry-cli releases set-commits --auto $VERSION` which returns:
`error: For non git repositories explicit revisions are required`

```
sentry-cli --version
sentry-cli 1.35.4
```

This issue seems to be similar to https://github.com/getsentry/sentry-cli/issues/115 and https://github.com/getsentry/sentry-cli/issues/300, but unlike both of those issues, my repo is a Github repo: 

```
sentry-cli repos list
+----------------------------------+----------+-----------------------------------------------------+
| Name                             | Provider | URL                                                 |
+----------------------------------+----------+-----------------------------------------------------+
| getslideapp/slide-django-adaptor | GitHub   | https://github.com/getslideapp/slide-django-adaptor |
+----------------------------------+----------+-----------------------------------------------------+
```
One additional piece of information is that since I have multiple Github accounts on my machine, I use a configuration as described in this gist: https://gist.github.com/jexchan/2351996. As such, my github hostname is `github.com-tjg37` instead of `github.com`. I'm not sure if this is the source of the issue, but if it is, would it be possible to allow for hostnames of this form? 

```
git config --get remote.origin.url
git@github.com-tjg37:getslideapp/slide-django-adaptor.git
```

Also, as suggested in https://github.com/getsentry/sentry-cli/issues/115 and https://github.com/getsentry/sentry-cli/issues/300, if I manually add the previous commits `sentry-cli releases set-commits $VERSION --commit "$REPO@$VERSION"` it works.

Zabanya:
same problem,  i'm using Bitbucket Pipelines and V2 connector
```
        - step:
            name: Sentry Release
            image: getsentry/sentry-cli
            script:
              - sentry-cli releases -o abc -p xyz new "$BITBUCKET_COMMIT"
              - sentry-cli releases -o abc -p xyz set-commits --auto "$BITBUCKET_COMMIT"
```

```
[...]

PR #378, opened 2018-09-13
Handle API change in repository provider IDs

nijel:
The ID now might contain integrations: prefix.

I'm not really sure this is a correct fix, it's just wild guess and I currently can't compile the code locally to test it.

See https://github.com/getsentry/sentry-cli/issues/371

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -169,7 +169,7 @@ fn find_reference_url(repo: &str, repos: &[Repo]) -> Result<String, Error> {
         }
 
         match configured_repo.provider.id.as_str() {
-            "github" | "bitbucket" | "visualstudio" | "git" => {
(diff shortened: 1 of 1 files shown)


nijel:
I have no clue if I can anyhow influence the test failures:

```
error: API request failed
  caused by: sentry reported an error: Authentication credentials were not provided. (http status: 401)
```

tonyo:
@nijel Don't worry about the test failures, currently those commands will fail for any PR from forked projects. We'll fix this.
Regarding the issue: should be fixed soon via https://github.com/getsentry/sentry-cli/pull/381

jan-auer:
@nijel Sorry for lacking behind, and thank you so much for providing a fix for this issue. 🎉  I just noticed this PR right now and had already opened mine in the meanwhile.

Since we need some more cleanup (e.g. support for GH Enterprise) I am going to close this PR and continue on the other one. Again, thanks for taking the time to fix this!

PR #379, opened 2018-09-13
fix: correct bin path

stephan-nordnes-eriksen:
This seems, at least on my CI system, to fix the issue discussed here; https://github.com/getsentry/react-native-sentry/issues/401

This change might have some unknown consequences which I have not yet seen. Please review thoroughly before a possible merge.


stephan-nordnes-eriksen:
Yeah, this PR seems to be breaking stuff. Sorry for that!

PR #400, opened 2018-09-28
ci: Update "gcs" target options, make zeus upload fail for releases

tonyo:
`|| [[ ! "$TRAVIS_BRANCH" =~ ^release/ ]]` will only fail the build if the command before it fails *and* it's a build on the release branch.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,5 +1,5 @@
 ---
-minVersion: '0.5.1'
+minVersion: '0.6.0'
 github:
   owner: getsentry
   repo: sentry-cli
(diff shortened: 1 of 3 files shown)


.craft.yml
```diff
@@ -30,12 +30,14 @@ targets:
     bucket: sentry-sdk-assets
     paths:
       - path: /sentry-cli/{{version}}/
-        maxCacheAge: 2592000
+        metadata:
+          cacheControl: 'public, max-age=2592000'
```
jan-auer: It's already a bit late for this, but shouldn't this be a setting configured more centrally in GCS rather than in each repository? Not that we're going to make many changes on this, but it feels wrong (and especially too verbose) to have this here.

.craft.yml
```diff
@@ -30,12 +30,14 @@ targets:
...
       - path: /sentry-cli/latest/
-        maxCacheAge: 600
+        metadata:
+          cacheControl: 'public, max-age=600'
   - name: registry
     type: app
     urlTemplate: "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/{{file}}"
-    includeNames: /^.*sentry-cli.*$/
+    includeNames: /^sentry-cli-.*$/
```
jan-auer: Seeing this change, this would be a good point to consider tagging artifacts or relying on the MIME type, maybe? Should not block this PR, but I highly encourage to consider this. Otherwise, we might get false-positives if we coincidentally add another artifact starting with `sentry-cli-`, which is not so unlikely.

.craft.yml
tonyo: Different projects (and more interestingly, different paths in the same bucket) may have different metadata values. I don't see a nicer way to store this kind of information anywhere else without confusing people who push new releases.

.craft.yml
tonyo: You're totally right, this should be definitely implemented at some point.

PR #403, opened 2018-10-03
fix: Prepare for making --rewrite the default

untitaker:
- Write out warning if the option is not set in the js helper or not specified
as CLI flag.
- Make --rewrite and --no-rewrite actually mutually exclusive.
- If `rewrite` in the js helper is explicitly set to false, pass `--no-rewrite` to CLI

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -48,12 +48,13 @@ function mockBinaryPath(mockPath) {
  * @returns {string[]} An arguments array that can be passed via command line.
  */
 function serializeOptions(schema, options) {
-  return Object.keys(schema).reduce((newOptions2, option) => {
+  return Object.keys(schema).reduce((newOptions, option) => {
     const paramValue = options[option];
     if (paramValue === undefined) {
-      return newOptions2;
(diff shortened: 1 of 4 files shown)


js/helper.js
```diff
@@ -48,42 +48,42 @@ function mockBinaryPath(mockPath) {
  * @returns {string[]} An arguments array that can be passed via command line.
  */
 function serializeOptions(schema, options) {
-  return Object.keys(schema).reduce((newOptions, option) => {
+  return Object.keys(schema).reduce((newOptions2, option) => {
     const paramValue = options[option];
     if (paramValue === undefined) {
-      return newOptions;
+      return newOptions2;
```
untitaker: Not sure why I had to rename this, but `npm test` would fail otherwise

jan-auer: Because you're otherwise shadowing that name in line 57, which in JavaScript is a common bug and thus linted against.

untitaker: Amazing, I would've expected that thing to make a new scope

Issue #279, opened 2018-04-02
Offline Install


oatkiller:
Hello,
I am trying to use the sentry webpack plugin (https://github.com/getsentry/sentry-webpack-plugin) but my build environment is not connected to the Internet. I can manually download the release and provide a location to it in my file system in advance. I see that configuring the CDN location is possible, but how would you feel about having an option to use the release from disk instead of downloading it at all? As things are now, my only option would be to run a temporary HTTP server (like the install script does during testing.) Please advise. Thanks!

jan-auer:
I *think* we could add support for `file:///` URLs to solve this, despite for some concerns here: https://github.com/bitinn/node-fetch/issues/75. 

However, I assume you are hosting Sentry somewhere in your infrastructure. Wouldn't it be easier to just put the binary there?

oatkiller:
@jan-auer Thank you very much for the reply. We do have artifactory which we use to host yum, npm and other package management repositories. I'm not sure how I would use artifactory to make the sentry CLI binary available. If I found a way to do so, I would need to make sure that the correct URL pattern is supported. I've pasted a snippet from the docs to help explain what I mean:

```
By default, this package will download sentry-cli from releases. This should work fine for most people. If you are experiencing issues with downloading from GitHub, you may need to use a different download mirror. To use a custom CDN, set the npm config property sentrycli_cdnurl. The downloader will append "/<version>/sentry-cli-<dist>".
```

In the mean time, I have worked around this issue by keeping the linux CLI binary in my git repo and spawning a temporary HTTP server while running `yarn install`. I got the idea from the tests here: https://github.com/getsentry/sentry-cli/blob/dd1277280d32c7b25ab3e58f9302ce61426b325c/scripts/install.js#L123

tonyo:
We partially addressed the issue in https://github.com/getsentry/sentry-cli/pull/425
Starting from the next version binaries will be cached in `~/.npm/sentry-cli`. 
Probably this is not ideal for your specific use case, but at least it opens some room for automation.

Issue #273, opened 2018-03-16
Binary file caching


mike1808:
Please implement caching for your binary file which is downloaded from `scripts/install.js`. It is pretty big and when I regularly switch between my different branches where the version of `@sentry/cli` is different. And it downloads the same versions.

tonyo:
Fixed in https://github.com/getsentry/sentry-cli/pull/425

PR #425, opened 2018-11-13
feat: Cache binaries on npm/yarn install

timfish:
Cache binaries locally which helps when using a poor or intermittent connection but also allows the cli to support offline yarn install.

Fixes: #279 and #273

Also fixes an issue where a failed partial download could only be remedied by removing the partial binary or nuking `node_modules`.

- Cache location and temp file logic comes from [`prebuilt-install`](https://github.com/prebuild/prebuild-install/blob/0da7f4e7d838dfcf4f197350a3de960ee66b20fa/util.js#L79) which is robust and well tested
  - Addition of `npm_config_yarn_offline_mirror` as cache location
- Fixed `package.json` scripts so they work on Windows

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -2,7 +2,7 @@
 
 'use strict';
 
-const fs = require('fs-extra');
+const fs = require('fs');
 const http = require('http');
 const os = require('os');
 const path = require('path');
@@ -12,6 +12,8 @@ const HttpsProxyAgent = require('https-proxy-agent');
 const fetch = require('node-fetch');
 const ProgressBar = require('progress');
 const Proxy = require('proxy-from-env');
+const copyFileSync = require('fs-copy-file-sync');
+const mkdirp = require('mkdirp');
 
 const helper = require('../js/helper');
 const pkgInfo = require('../package.json');
@@ -114,7 +116,7 @@ function downloadBinary() {
 
   const cachedPath = getCachedPath(downloadUrl);
   if (fs.existsSync(cachedPath)) {
-    fs.copyFileSync(cachedPath, outputPath);
+    copyFileSync(cachedPath, outputPath);
     return Promise.resolve();
   }
 
@@ -136,7 +138,7 @@ function downloadBinary() {
     const progressBar = createProgressBar(name, total);
 
     const tempPath = getTempFile(cachedPath);
-    fs.ensureDirSync(path.dirname(tempPath));
+    mkdirp.sync(path.dirname(tempPath));
 
     return new Promise((resolve, reject) => {
       response.body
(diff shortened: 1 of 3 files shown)


scripts/install.js
```diff
@@ -65,6 +66,39 @@ function createProgressBar(name, total) {
...
+function getCachedPath(url) {
+  const digest = crypto
+    .createHash('md5')
+    .update(url)
+    .digest('hex')
+    .slice(0, 6);
+
+  return path.join(
+    npmCache(),
+    `${digest}-${path.basename(url).replace(/[^a-zA-Z0-9.]+/g, '-')}`
```
tonyo: What do you think about adding an additional intermediate directory (e.g. `sentry-cli`) to place the binaries there?

scripts/install.js
```diff
@@ -65,6 +66,39 @@ function createProgressBar(name, total) {
   return { tick: () => {} };
 }
 
+function copyFileSync(src, dest) {
+  const data = fs.readFileSync(src);
+  fs.writeFileSync(dest, data);
+}
```
tonyo: This copy function does not preserve file mode, and as a result, the output file does not have the execution bits set. Because of that we get the following CI failure: https://travis-ci.com/getsentry/sentry-cli/jobs/158100653

Additionally, `sentry-cli` binaries are (sadly) not that small at the moment (~12-20 MB), so it would be a bit nicer not to read the entire file into memory and use read/write streams instead.

scripts/install.js
timfish: yep, good idea!

scripts/install.js
timfish: oh no 🤦.

ok just to use this?
https://www.npmjs.com/package/fs-copy-file-sync

timfish: Scrap that. Went with `fs-extra` because we also need to use `mkdirp` or similar.

timfish:
@tonyo Looks like `fs-extra` doesn't support node v4. I can attempt to find separate packages that support node that old or we could update the minimum supported node version to v6? v4 has been End-of-life for over 6 months.

tonyo:
@timfish we would like to keep compatibility with v4 for a bit longer, sorry :(
If you don't have any compatible packages in mind, I guess we can borrow implementations from e.g. `fs-extra` into a separate file and add a license/copyright notice there.

timfish:
@tonyo no worries at all. The Sentry people have some serious LTS respect and I like that 😄

I'm tempted to play a short game of dependency vs CI whack-a-mole before venturing into copy-paste.

timfish:
@tonyo, that appears to have fixed it!

jan-auer:
Oh wait, I might have approved a little early. We're not including sentry-cli's version in the cache path. That means we will have to check the version of the cached binary and replace old versions when downloading newer ones (whereas we can always use a newer binary).

tonyo:
@timfish  🎉  Thanks! This will certainly improve the installation experience.

PR #438, opened 2018-12-05
ref: Use release registry for update checks

tonyo:
Also: fix check that the installation was done via NPM.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -100,10 +100,7 @@ function getTempFile(cached) {
 function downloadBinary() {
   const arch = os.arch();
   const platform = os.platform();
-  const outputPath = path.resolve(
-    __dirname,
-    platform === 'win32' ? '../sentry-cli.exe' : '../sentry-cli'
-  );
+  const outputPath = helper.getPath();
 
   const downloadUrl = getDownloadUrl(platform, arch);
   if (!downloadUrl) {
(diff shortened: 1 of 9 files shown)


src/commands/update.rs
```diff
@@ -21,11 +21,17 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
 }
 
 pub fn execute<'a>(matches: &ArgMatches<'a>) -> Result<(), Error> {
+    // Disable update check in case of errors
+    env::set_var("SENTRY_DISABLE_UPDATE_CHECK", "true");
+
+    // Aborts with an error if this installation is not updatable.
+    assert_updatable()?;
```
tonyo: I think it makes sense to check this before hitting the registry to save some requests.

tonyo: @jan-auer We also want to avoid re-running the update when we have any kind of errors:

https://github.com/getsentry/sentry-cli/blob/31aacf105125f7797fb29ac5c468e6da7d7ec699/src/commands/mod.rs#L255

scripts/install.js
```diff
@@ -102,7 +102,7 @@ function downloadBinary() {
   const platform = os.platform();
   const outputPath = path.resolve(
     __dirname,
-    platform === 'win32' ? '../bin/sentry-cli.exe' : '../sentry-cli'
+    platform === 'win32' ? '../sentry-cli.exe' : '../sentry-cli'
```
tonyo: Let's update it, so the binaries are downloaded to the same directory. It's important for `is_npm_install` check.

PR #443, opened 2018-12-07
Adding snap package

joedborg:


tonyo:
@joedborg could you elaborate on your use case? Does any of the existing installation methods not work for you?

joedborg:
Hey @tonyo,

It makes installation on most popular Linux distributions much easier than, say, starting a Docker container each time.  It'd be as simple as `snap install sentry-cli` and then you'd have the `sentry-cli` executable.

kamilogorek:
Closing, as it's from almost 2y ago. Will reopen if necessary.

PR #456, opened 2018-12-18
feat: Retry on upload-related operations (chunk-upload, assemble)

tonyo:
Adds `with_retry` method on `ApiRequest` that retries the request with exponential (but constrained) backoff.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 [dependencies]
 anylog = "0.3.0"
 app_dirs = "1.2.1"
+backoff = "0.1.2"
 backtrace = "0.3.9"
 brotli2 = "0.3.2"
 chardet = "0.2.4"
--- a/src/api.rs
+++ b/src/api.rs
@@ -905,8 +905,7 @@ impl Api {
     pub fn get_chunk_upload_options(&self, org: &str) -> ApiResult<Option<ChunkUploadOptions>> {
         let url = format!("/organizations/{}/chunk-upload/", org);
         match self
-            .request(Method::Get, &url)?
-            .send()?
+            .get(&url)?
             .convert_rnf(ApiErrorKind::ChunkUploadNotSupported)
         {
             Ok(options) => Ok(Some(options)),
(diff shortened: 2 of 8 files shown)


tonyo:
Discussed this with @jan-auer. Moving the retry code to `send()` sounds like a good idea, will update the PR.

src/api.rs
```diff
@@ -616,6 +639,14 @@ impl Api {
...
             .request(Method::Post, &path)?
             .with_form_data(form)?
+            .with_retry(
+                self.config.get_max_retry_count().unwrap(),
+                &[
+                    HTTP_STATUS_502_BAD_GATEWAY,
+                    HTTP_STATUS_503_SERVICE_UNAVAILABLE,
+                    HTTP_STATUS_504_GATEWAY_TIMEOUT,
+                ],
+            )?
```
tonyo: Just a note here, the file uploads are not really idempotent ("File already present" might be returned on a retry).

Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2018"
 [dependencies]
 anylog = "0.3.0"
 app_dirs = "1.2.1"
+backoff = "0.1.2"
```
jan-auer: ❤️ for alphabetical order

src/config.rs
```diff
@@ -325,6 +325,16 @@ impl Config {
             .unwrap_or(35 * 1024 * 1024))
     }
 
+    pub fn get_max_retry_count(&self) -> Result<u32, Error> {
+        if env::var_os("SENTRY_REQUEST_MAX_RETRY").is_some() {
+            Ok(env::var("SENTRY_REQUEST_MAX_RETRY")?.parse()?)
+        } else if let Some(val) = self.ini.get_from(Some("http"), "max_retry_count") {
```
jan-auer: I know that we've been inconsistent with these, but can we make them more alike? For example `export SENTRY_HTTP_UPLOAD_RETRIES=2` and `http.upload_retries = 2`.

On a side note, after releasing this we should also document the config [here](https://docs.sentry.io/cli/configuration/#configuration-values)

tonyo: Updated!

PR #467, opened 2019-01-16
feat: Allow to hide progress bars

### LINKED ISSUE #282: Add a `silent` or `quiet` flag for `releases files upload-sourcemaps`
We use this excellent tool as part of our CI.

The `upload-sourcemaps` tool produces insanely verbose output, so much that the CI tool we use (circleci) truncates the output, preventing us from seeing the important parts (did the deployment work)? that happens after sourcemap upload.

It would be very helpful if we could have an option to silence all non-error output from the cli tool.

Example command:

```bash
sentry-cli --auth-token <token> releases -o <organisation> -p <project> files <hash> upload-sourcemaps --validate dist/
```

tonyo:
If `SENTRY_NO_PROGRESS_BAR` is set to "1", all progress bar operations should be noops now.

Update of "indicatif" was not really required, but why not.

Fixes #282.


dimaqq:
Is there an equivalent command-line option? 🤔

kamilogorek:
No, it's only available through env configuration.

PR #474, opened 2019-02-04
feat(proguard): Log mapping file used and hash

bruno-garcia:


bruno-garcia:
@tonyo I had it tested with the new version of deps as I had to point to patched version of symbolic but I could rollback and test it again.

tonyo:
@bruno-garcia If you didn't test a bunch of other commands, I wouldn't do the update in the same PR. Just to avoid the unnecessary risk 🤷‍♂️

bruno-garcia:
@tonyo fixed

PR #484, opened 2019-02-13
fix(upload): Flush archives before uploading them

jan-auer:
Fixes incomplete DIF and Proguard uploads on Windows. 

Instead of Cloning the file handle, we now return a mutable reference to the actual file, instead. Additionally, this flushes all data to **disk** before passing the path on to libcurl. Without that, libcurl on Windows seems to pick up an incomplete state of the file.

I yet have to find out why Sentry did not recognize the form field at all before.


jan-auer:
Verified that flushing is **required** on Windows.

PR #525, opened 2019-05-06
fix(releases) Accept more Azure Devops URLs

markstory:
Parse SSH URLs from Azure DevOps that use the project name as the SSH username.

Fixes #11055


markstory:
This was related to getsentry/sentry#11055

PR #527, opened 2019-05-12
ref: Expose function for unpacking a ram bundle

reaganboy916:
S


tonyo:
Thanks, but we'll open a PR when the feature is ready :)

PR #528, opened 2019-05-14
feat: Implement file RAM bundles

tonyo:


tonyo:
Renamed "ram-bundle" arguments of upload-sourcemap to "bundle", so that we also support regular (javascript) ones.

PR #551, opened 2019-06-17
feat(releases): Allow port in url

### LINKED ISSUE #539: set-commits --auto throws error saying I'm not in a repository
Version: 1.44.1

When I try to do the `sentry-cli releases set-commits --auto $VERSION` command from within a repo folder, I get the following error:

```
error: For non git repositories explicit revisions are required
```

I'm in a git repository, so I don't understand why this is giving me trouble. This fails locally and on CI. I could be doing something else wrong.

I am using these sources as references:
- https://docs.sentry.io/cli/releases/#sentry-cli-commit-integration
- https://docs.sentry.io/workflow/releases/?_ga=2.225371167.422867857.1559060054-1248220087.1552929159&platform=javascript#create-release

icfr:
fix #539


markstory:
Thank you 👍

PR #552, opened 2019-06-18
Allow options to pass to releases for `silent` support.

roelvanhintum:


js/releases/index.js
```diff
@@ -18,7 +18,16 @@ const SOURCEMAPS_SCHEMA = require('./options/uploadSourcemaps');
...
  */
-module.exports = {
+class Releases {
+  /**
+   * Creates a new `Releases` instance.
+   *
+   * @param {Object} [options] More options to pass to the CLI
+   */
+  constructor(options) {
+    this.options = options;
```
kamilogorek: Can we default to `{}` in params here just in case? It will be a very soft guard for `this.options.silent` accessing undefined.

jan-auer:
Thanks!

PR #557, opened 2019-06-19
fix(appcenter): Do not require version name

### LINKED ISSUE #554: [Breaking change] React Native symbol upload iOS now requires version name parameter
Previously only `--bundle-id` was required when running in a non macOS environment, now `--version-name` is also required.

Error after upgrading:
```
$ sentry-cli "react-native" "appcenter" "-o" "storr" "-p" "mobile" "--deployment" "master" "--bundle-id" "app.storr" "storr/app.storr-ios" "ios" "./build/ios/CodePush"

> Fetching latest AppCenter deployment info
error: AppCenter codepush releases for iOS require macOS if no bundle ID is specified
```

Code where this is checked
https://github.com/getsentry/sentry-cli/blob/585f41ea4b7c0e6f1a1a591c56fffa649f6a6070/src/utils/appcenter.rs#L138-L143

Change introduced in https://github.com/getsentry/sentry-cli/commit/585f41ea4b7c0e6f1a1a591c56fffa649f6a6070#diff-8e4c8496f8f401a0cebfddf00bcdc705R138

The following line is now being triggered because the above condition is no longer being met:
https://github.com/getsentry/sentry-cli/blob/585f41ea4b7c0e6f1a1a591c56fffa649f6a6070/src/utils/appcenter.rs#L147

At a minimum, the above error should be updated; and this breaking change should be added to the changelog (version 1.44.4), to prevent future confusion.

tonyo:
Fixes #554

### DIFF
--- a/src/utils/appcenter.rs
+++ b/src/utils/appcenter.rs
@@ -135,11 +135,15 @@ pub fn get_react_native_appcenter_release(
     let bundle_id_ovrr = bundle_id_override.unwrap_or("");
     let version_name_ovrr = version_name_override.unwrap_or("");
 
-    if bundle_id_ovrr != "" && version_name_ovrr != "" {
-        return Ok(format!(
-            "{}-{}-codepush:{}",
-            bundle_id_ovrr, version_name_ovrr, package.label
-        ));
+    if bundle_id_ovrr != "" {
+        if version_name_ovrr == "" {
+            return Ok(format!("{}-codepush:{}", bundle_id_ovrr, package.label));
+        } else {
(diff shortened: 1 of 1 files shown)


tonyo:
We probably don't want this, see the discussion in https://github.com/getsentry/sentry-cli/pull/506

pvinis:
Why was this closed? Will this not be fixed for some reason?

jan-auer:
This was closed because the fix was actually invalid. The proper solution will be to revert the change that introduced the regression, but we have not fully assessed the problem just yet.

PR #558, opened 2019-06-25
feat(api): Implement source bundle uploads

jan-auer:


jan-auer:
The changes for release file uploads have not been tested yet.

PR #574, opened 2019-07-26
build: Update Craft config (checksums, required files)

tonyo:
Here we use a couple of new Craft features:
* `requireNames` will check that every pattern has a matching file present. This will help to avoid situations when Zeus doesn't know anything about Appveyor some builds (e.g. those that were not started yet) and reports build as green.
* `checksums` will compute and add file checksums to the release registry.


tonyo:
> might be a good time to move on

Zeus will be like


image

PR #583, opened 2019-08-19
fix: log missing if errtrace is used

jsecchiero:
How to reproduce:

```
#!/bin/bash
set -E

>&2 echo error
$(false)
echo no error
```

this gives:
```
error: Could not open logfile
  caused by: No such file or directory (os error 2)
```

Actually i'm using this workaround:
```
eval "$(sentry-cli bash-hook | sed 's/: > "\$_SENTRY_LOG_FILE"//g' | sed 's/export SENTRY_LAST_EVENT/: > "\$_SENTRY_LOG_FILE"\n  export SENTRY_LAST_EVENT/g')"
```

### DIFF
--- a/src/bashsupport.sh
+++ b/src/bashsupport.sh
@@ -28,6 +28,8 @@ _sentry_err_trap() {
   _sentry_traceback 1
   echo "@command:${_command}" >> "$_SENTRY_TRACEBACK_FILE"
   echo "@exit_code:${_exit_code}" >> "$_SENTRY_TRACEBACK_FILE"
+
+  : >> "$_SENTRY_LOG_FILE"
   export SENTRY_LAST_EVENT=$("___SENTRY_CLI___" bash-hook --send-event --traceback "$_SENTRY_TRACEBACK_FILE" --log "$_SENTRY_LOG_FILE")
   rm -f "$_SENTRY_TRACEBACK_FILE" "$_SENTRY_LOG_FILE"
 }


jsecchiero:
the lint test fail seems have nothing to do with the PR

src/bashsupport.sh
```diff
@@ -28,6 +28,8 @@ _sentry_err_trap() {
   _sentry_traceback 1
   echo "@command:${_command}" >> "$_SENTRY_TRACEBACK_FILE"
   echo "@exit_code:${_exit_code}" >> "$_SENTRY_TRACEBACK_FILE"
+
+  : > "$_SENTRY_LOG_FILE"
```
jan-auer: Seems like this should append instead of overwriting? @tonyo could you double-check this please?

tonyo: Agreed, otherwise it looks like we'll be sending empty _SENTRY_LOG_FILE all the time.
At the same time, I'd also leave `: > "$_SENTRY_LOG_FILE"` at the end of the file (currently removed), just to make sure that we're clearing the log file while doing the initial "eval".
What do you think @jsecchiero ?

jsecchiero: https://github.com/getsentry/sentry-cli/blob/5c2b9e894ba14934a8be813ae6f8129a11cf57c2/src/bashsupport.sh#L44
I don't think to have enough sentry-cli knowledge to say if it's the best solution, but this modification just Imitate the behavior of _sentry_traceback func and _SENTRY_TRACEBACK_FILE. Where there is not a initial creation of the file _SENTRY_TRACEBACK_FILE and no append in the _sentry_traceback func.
Let me know if there is no any relationship between the two so i can rebase based on your suggestions.

tonyo: The difference between `*TRACEBACK_FILE` and `*LOG_FILE` is that as part of the hook, we redirect standard input and error to `*LOG_FILE`: https://github.com/getsentry/sentry-cli/blob/5c2b9e894ba14934a8be813ae6f8129a11cf57c2/src/bashsupport.sh#L56-L57
...so that lines capture there can be used as breadcrumbs. So essentially it is your script (not sentry-cli) who's writing data there.

And `*TRACEBACK_FILE` is generated by explicitly by `bash-hook` here: 
https://github.com/getsentry/sentry-cli/blob/5c2b9e894ba14934a8be813ae6f8129a11cf57c2/src/bashsupport.sh#L37-L52

That's why they are handled a bit differently (the contents of `*LOG_FILE` is not really something we want to control).

jsecchiero: thanks for the explanation.
Rebased

tonyo:
@jsecchiero No problem, thanks for the fix!

PR #584, opened 2019-08-22
ref: Remove short version

HazAT:


src/commands/releases.rs
```diff
@@ -136,9 +136,6 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                  .help("Set the release time. [defaults to the current time]")))
         .subcommand(App::new("list")
             .about("List the most recent releases.")
-            .arg(Arg::with_name("no_abbrev")
```
HazAT: How do we make this non breaking?
(How did we handle this in the past?)

PR #598, opened 2019-09-18
fix(bug): Using annotated tags now returns the correct commit SHA

iProgramStuff:
A customer was experiencing an issue when trying to specify commits using an annotated tag. The `find_matching_rev` function was returning the SHA of the tag instead of getting the SHA of it's associated commit.

This PR fixes that, and adds two tests (one against a lightweight tag, and one against an annotated tag) to ensure it's correctness and keep it correct.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -248,25 +248,42 @@ fn find_matching_rev(
         if let Ok(remote) = repo.find_remote("origin");
         if let Some(url) = remote.url();
         then {
-            debug!("if_chain macro! {:?}, {:?}, {:?}", url, &reference_url, is_matching_url(url, &reference_url));
             if !discovery || is_matching_url(url, &reference_url) {
                 debug!("  found match: {} == {}, {:?}", url, &reference_url, r);
                 let head = repo.revparse_single(r)?;
-                debug!("  head: {:?}", head);
                 if let Some(tag) = head.as_tag(){
                     debug!("  tag.target(): {:?}", tag.target());
                     if let Ok(tag_commit) = tag.target() {
                         return Ok(Some(log_match!(tag_commit.id().to_string())));
                     }
                 }
-                debug!("  Not a tag");
+                debug!("  not a tag");
+
                 return Ok(Some(log_match!(head.id().to_string())));
             } else {
                 debug!("  not a match: {} != {}", url, &reference_url);
             }
         }
     }
+    if let Ok(submodule_match) = find_matching_submodule(r, reference_url, repo) {
+        return Ok(submodule_match);
+    }
+    info!("  -> no matching revision found");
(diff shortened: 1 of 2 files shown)


src/utils/vcs.rs
```diff
@@ -377,6 +388,56 @@ pub fn find_heads(specs: Option<Vec<CommitSpec>>, repos: &[Repo]) -> Result<Vec<
...
+            id: String::from("1"),
+            name: String::from("getsentry/sentry-cli"),
+            url: Some(String::from("https://github.com/getsentry/sentry-cli")),
+            provider: RepoProvider { id: String::from("integrations:github"), name: String::from("GitHub") },
+            status: String::from("active"),
+            date_created: chrono::Utc::now()
+        }
+    ];
+
+    let res_with_lightweight_tag = find_matching_rev(reference, &spec, &repos, false);
```
bruno-garcia: I'd suggest having a test for each test-case:
`test_res_with_lightweight_tag`
`test_res_with_annotated_tag`

iProgramStuff: Split them up :)

src/utils/vcs.rs
```diff
@@ -249,15 +249,41 @@ fn find_matching_rev(
...
-                debug!("  found match: {} == {}", url, &reference_url);
+                debug!("  found match: {} == {}, {:?}", url, &reference_url, r);
                 let head = repo.revparse_single(r)?;
+                if let Some(tag) = head.as_tag(){
+                    debug!("  tag.target(): {:?}", tag.target());
+                    if let Ok(tag_commit) = tag.target() {
+                        return Ok(Some(log_match!(tag_commit.id().to_string())));
+                    }
+                }
+                debug!("  not a tag");
```
jan-auer: This logging is a bit verbose, even for `debug`. Mind if we remove these two additional logs again or make them `trace!`?

src/utils/vcs.rs
```diff
@@ -249,15 +249,41 @@ fn find_matching_rev(
...
         }
     }
+    if let Ok(submodule_match) = find_matching_submodule(r, reference_url, repo) {
+        return Ok(submodule_match);
+    }
+    info!("  -> no matching revision found");
+    Ok(None)
+}
 
+fn find_matching_submodule(
```
jan-auer: Was this factored out for a specific reason or just to restructure?

If you prefer to keep it that way, please move the macro out of the functions to prevent duplication.

iProgramStuff: I took it out because the linter was saying the function was too cognitively complex (31). 
Macro function moved out :)

src/utils/vcs.rs
iProgramStuff: Removed.

PR #612, opened 2019-10-29
release list: Show a column with each release project's slug.

kAworu:
Before this patch, the release list command would not show the
project(s) associated to each release.

After this patch the "Projects" column show the all the release's
project slugs.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1830,6 +1830,7 @@ pub struct ReleaseInfo {
     pub last_event: Option<DateTime<Utc>>,
     #[serde(rename = "newGroups")]
     pub new_groups: u64,
+    pub projects: Vec<ProjectSlugAndName>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -1958,6 +1959,12 @@ pub struct AssociateDsymsResponse {
 #[derive(Deserialize, Debug)]
 pub struct Team {
     pub id: String,
+    pub slug: String,
+    pub name: String,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct ProjectSlugAndName {
     pub slug: String,
     pub name: String,
 }
--- a/src/api.rs
+++ b/src/api.rs
@@ -1830,6 +1830,7 @@ pub struct ReleaseInfo {
     pub last_event: Option<DateTime<Utc>>,
     #[serde(rename = "newGroups")]
     pub new_groups: u64,
+    #[serde(default)]
     pub projects: Vec<ProjectSlugAndName>,
 }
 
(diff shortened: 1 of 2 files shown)


kAworu:
The `release list` command is becoming useless for us with only a handful of projects having their version based on git commit hash. If there is no interest for that feature feel free to close this PR.

About the implementation, having to define the `ProjectSlugAndName` is not ideal but I don't know a cleaner way.

jan-auer:
Hi @kAworu, thanks for the patch. This is a useful addition. Before merging this, I have a few suggestions and was wondering what you think about this:

The table can get quite big when adding projects, to a point where it line-wraps in a terminal. I played around with this and found that joining with newlines could look better in such cases:

image

Also, we could hide this behind a flag, like `sentry-cli releases list --show-projects` and `-P`. Then, by default, the table is small, but if someone needs more information they can display it.

src/api.rs
```diff
@@ -1830,6 +1830,7 @@ pub struct ReleaseInfo {
     pub last_event: Option<DateTime<Utc>>,
     #[serde(rename = "newGroups")]
     pub new_groups: u64,
+    pub projects: Vec<ProjectSlugAndName>,
```
jan-auer: Could you please add `#[serde(default)]` here, to be safe? IIRC, old versions of this endpoint did not return a list of projects.

src/commands/releases.rs
```diff
@@ -539,6 +540,16 @@ fn execute_list<'a>(ctx: &ReleaseContext<'_>, _matches: &ArgMatches<'a>) -> Resu
             row.add("(unreleased)");
         }
         row.add(&release_info.version);
+        let project_slugs = release_info
+            .projects
+            .iter()
+            .map(|p| p.slug.clone())
```
jan-auer: Cloning can be avoided here:

```rs
.into_iter()
.map(|p| p.slug)
```

src/commands/releases.rs
```diff
@@ -539,6 +540,16 @@ fn execute_list<'a>(ctx: &ReleaseContext<'_>, _matches: &ArgMatches<'a>) -> Resu
...
+            .collect::<Vec<_>>();
+        if !project_slugs.is_empty() {
+            row.add(project_slugs.join(","));
```
jan-auer: Depending on which version you like more, please use `"\n"` or `", "` (with a space).

kAworu:
Hi @jan-auer,

Thanks for taking the time to review and test this patch!

I agree about both hiding the column behind a flag to keep the default table short, and that newline is a cleaner separator. I've implemented your suggestions and feedback and look forward to hear from you.

jan-auer:
Excellent, thank you so much. Can you just run `cargo fmt` to fix the style errors? Then this is good to go!

kAworu:
Of course.

I've been building / running `sentry-cli` with Docker in order to work on this patch. I was wondering if there is a way to run the format check / lint from Docker too?

mitsuhiko:
You should be able to run `rustup component add rustfmt` and then run `cargo fmt` in docker.

PR #627, opened 2019-11-16
feat: Allow overriding dist in Xcode

### LINKED ISSUE #609: Allow setting/overriding Bundle ID for react native uploads.
Currently all of the react-native platforms with the exception of gradle don't allow you to override the generated/parsed bundle id when uploading source maps.

dotconnor:
Closes #609
Closes https://github.com/getsentry/sentry-react-native/issues/676

### DIFF
--- a/src/commands/react_native_xcode.rs
+++ b/src/commands/react_native_xcode.rs
@@ -78,6 +78,15 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                 ),
         )
         .arg(
+            Arg::with_name("dist")
+                .long("dist")
+                .value_name("DISTRIBUTION")
+                .required(true)
+                .multiple(true)
+                .number_of_values(1)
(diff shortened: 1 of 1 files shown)


src/commands/react_native_xcode.rs
```diff
@@ -77,6 +77,15 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                      to use.  By default the bundled build script is used.",
                 ),
         )
+        .arg(
+            Arg::with_name("dist")
+                .long("dist")
+                .value_name("DISTRIBUTION")
+                .required(true)
```
jan-auer: This will break the current CLI as it requires the `dist` argument to be passed. Can you remove this line, please?

HazAT:
I think we can merge this @jan-auer , cc @kamilogorek

kamilogorek:
Looks good to me. CI symbolic issues are not related to this PR.

PR #636, opened 2019-12-18
feat: Add support for git:// git+ssh:// and git+https?:// protocols

kamilogorek:
Git integration always provides regular `https://git.whatever.com/repo` url, and some people are using `git+ssh://` protocol locally, which doesn't match and prevents `set-commit --auto` from working.


kamilogorek:
True, but I thought that performing additional look-behind was an overkill

PR #637, opened 2019-12-19
feat: Allow setting of Git remote

nalysius:
Git is decentralised, so it's usual to have several remote
repositories. Origin is usually the default remote repository
but it's not mandatory.

Sentry-cli only allowed to use the remote repository "origin" when
working with commits during a release.

It is now possible to set the remote name in the configuration file
or as environment variable.
In ~/.sentryclirc: defaults.vcs_remote.
In env: SENTRY_VCS_REMOTE.
If not provided, defaults to "origin" so it's backward-compatible.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -39,6 +39,7 @@ pub struct Config {
     cached_auth: Option<Auth>,
     cached_base_url: String,
     cached_log_level: log::LevelFilter,
+    cached_vcs_remote: String,
 }
 
 impl Config {
@@ -51,6 +52,7 @@ impl Config {
             cached_auth: get_default_auth(&ini),
             cached_base_url: get_default_url(&ini),
             cached_log_level: get_default_log_level(&ini)?,
+            cached_vcs_remote: get_default_vcs_remote(&ini),
             ini,
         })
     }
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -546,3 +554,17 @@ fn get_default_log_level(ini: &Ini) -> Result<log::LevelFilter, Error> {
...
     Ok(log::LevelFilter::Warn)
 }
+
+/// Get the default VCS remote.
+///
+/// To be backward compatible the default remote is still
+/// origin.
+fn get_default_vcs_remote(ini: &Ini) -> String {
+    if let Ok(remote) = env::var("SENTRY_VCS_REMOTE") {
+        return remote.to_string();
```
kamilogorek: Remove `return` statements to unify with the rest of the file

src/utils/vcs.rs
```diff
@@ -355,14 +367,19 @@ pub fn find_head() -> Result<String, Error> {
 
 /// Given commit specs and repos this returns a list of head commits
 /// from it.
-pub fn find_heads(specs: Option<Vec<CommitSpec>>, repos: &[Repo]) -> Result<Vec<Ref>, Error> {
```
kamilogorek: The spec comment should be updated to mention remote as well.

kamilogorek:
Also can you update docs and mention that it should be merge once this change is? https://github.com/getsentry/sentry-docs/blob/master/src/collections/_documentation/cli/configuration.md

Thanks!

nalysius:
Thank you for your review @kamilogorek !

I've fixed the two points about the code, removed the `return`s and added the remote_name in the comment of the method. I've force-pushed to keep a clean commit history.
I will update the documentation too.

src/utils/vcs.rs
nalysius: Comment line 368 now speaks about remote.

src/config.rs
nalysius: `return`s have been removed.

PR #642, opened 2020-01-26
support dist option

farfromrefug:
I need this to publish artifacts with a dist so that sourcemap are associated correctly


kamilogorek:
Thanks!

PR #646, opened 2020-02-04
feat: Implement workarounds for dealing with hermes bytecode

Swatinem:
When encountering hermes bytecode, the file is cleared and flagged as minified source, that way it will get a correct sourcemap reference. Other than that, the update to `sourcemap` will make sure that the rewritten sourcemap will keep the `x_facebook_sources` scopes.

### DIFF
--- a/src/utils/sourcemaps.rs
+++ b/src/utils/sourcemaps.rs
@@ -270,9 +270,10 @@ pub struct UploadContext<'a> {
 }
 
 fn is_hermes_bytecode(slice: &[u8]) -> bool {
-    // The hermes bycode format magic is defined here:
+    // The hermes bytecode format magic is defined here:
     // https://github.com/facebook/hermes/blob/5243222ef1d92b7393d00599fc5cff01d189a88a/include/hermes/BCGen/HBC/BytecodeFileFormat.h#L24-L25
-    slice[..8] == 0x1F19_03C1_03BC_1FC6u64.to_le_bytes()
+    const HERMES_MAGIC: [u8; 8] = [0xC6, 0x1F, 0xBC, 0x03, 0xC1, 0x03, 0x19, 0x1F];
+    slice.starts_with(&HERMES_MAGIC)
 }
 
 impl SourceMapProcessor {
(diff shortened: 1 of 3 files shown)


src/utils/sourcemaps.rs
```diff
@@ -320,6 +326,14 @@ impl SourceMapProcessor {
                 || is_likely_minified_js(&contents)
             {
                 SourceFileType::MinifiedSource
+            } else if is_hermes_bytecode(&contents) {
+                // This is actually a big hack:
+                // For the react-native Hermes case, we skip uploading the bytecode bundle,
+                // and rather flag it as an empty "minified source". That way, it
+                // will get a SourceMap reference, and the server side processor
+                // should deal with it accordingly.
+                contents.truncate(0);
```
kamilogorek:
```suggestion
                contents.clear();
```

kamilogorek: Although we don't need to clear this vec at all, it'll be dropped anyway.

src/utils/sourcemaps.rs
```diff
@@ -611,12 +627,19 @@ impl SourceMapProcessor {
...
+                    sm.rewrite(&options)?.to_writer(&mut new_source)?
+                }
+                sourcemap::DecodedMap::Hermes(smh) => {
+                    smh.rewrite(&options)?.to_writer(&mut new_source)?
+                }
+                sourcemap::DecodedMap::Index(smi) => smi
+                    .flatten_and_rewrite(&options)?
+                    .to_writer(&mut new_source)?,
+                _ => panic!("invalid sourcemap type"),
+            };
```
kamilogorek: You are not writing rewritten sourcemaps to the new source in Regular and Hermes.

Swatinem: I am though, or am I missing something?

kamilogorek: Right, my bad, it got squished into a single line

HazAT:
Confirmed, just tested it end to end, it works!

src/utils/sourcemaps.rs
```diff
@@ -269,6 +269,12 @@ pub struct UploadContext<'a> {
     pub wait: bool,
 }
 
+fn is_hermes_bytecode(slice: &[u8]) -> bool {
+    // The hermes bycode format magic is defined here:
+    // https://github.com/facebook/hermes/blob/5243222ef1d92b7393d00599fc5cff01d189a88a/include/hermes/BCGen/HBC/BytecodeFileFormat.h#L24-L25
+    slice[..8] == 0x1F19_03C1_03BC_1FC6u64.to_le_bytes()
```
jan-auer:
```suggestion
    const HERMES_MAGIC: [u8; 8] = [0xC6, 0x1F, 0xBC, 0x03, 0xC1, 0x03, 0x19, 0x1F];
    slice.starts_with(HERMES_MAGIC);
```

jan-auer: Something like this, at least (might not compile off the bat). Nice benefit of this is that it doesn't panic if the slice is smaller than 8 bytes.

PR #653, opened 2020-02-11
feat(appcenter): Add `dist` option to `react-native appcenter`

lobsterkatie:
Allows the user to specify a `dist` value on their release artifacts when using `sentry-cli react-native appcenter` (see docs [here](https://docs.sentry.io/platforms/react-native/codepush/)), so that they'll match with events (which are given a `dist` value by default) - a prerequisite for deminification with source maps.

Fixes https://github.com/getsentry/sentry-react-native/issues/703.


lobsterkatie:
@kamilogorek - I don't have a set-up to test this. Any chance you could put it through its paces?

Issue #623, opened 2019-11-13
Cannot update CLI


ThibaultVlacich:
I have installed on my machine the CLI version 1.48.0 and I'm trying to update to 1.49.0, but I'm getting an error while doing so:

```
# sentry-cli update
Latest release is 1.49.0
Updating executable at /usr/local/bin/sentry-cli
Need to sudo to overwrite /usr/local/bin/sentry-cli
error: No such file or directory (os error 2)

Add --log-level=[info|debug] or export SENTRY_LOG_LEVEL=[info|debug] to see more output.
Please attach the full debug log to all bug reports.
```

Here is the full log:
```
# sentry-cli update --log-level=debug
  INFO    2019-11-13 15:43:04.910514350 +00:00 Loaded config from /root/.sentryclirc
  DEBUG   2019-11-13 15:43:04.910896399 +00:00 sentry-cli version: 1.48.0, platform: "linux", architecture: "x86_64"
  INFO    2019-11-13 15:43:04.911088331 +00:00 sentry-cli was invoked with the following command line: "sentry-cli" "update" "--log-level=debug"
  DEBUG   2019-11-13 15:43:04.919210637 +00:00 request GET https://release-registry.services.sentry.io/apps/sentry-cli/latest
  DEBUG   2019-11-13 15:43:04.919477699 +00:00 retry number 0, max retries: 0
  DEBUG   2019-11-13 15:43:05.261524217 +00:00 > GET /apps/sentry-cli/latest HTTP/1.1
  DEBUG   2019-11-13 15:43:05.261859178 +00:00 > Host: release-registry.services.sentry.io
  DEBUG   2019-11-13 15:43:05.262075095 +00:00 > Accept: */*
  DEBUG   2019-11-13 15:43:05.262287775 +00:00 > Connection: TE
  DEBUG   2019-11-13 15:43:05.262492260 +00:00 > TE: gzip
  DEBUG   2019-11-13 15:43:05.262767358 +00:00 > User-Agent: sentry-cli/1.48.0
  DEBUG   2019-11-13 15:43:05.363295106 +00:00 < HTTP/1.1 200 OK
  DEBUG   2019-11-13 15:43:05.363510562 +00:00 < Server: nginx
  DEBUG   2019-11-13 15:43:05.363755793 +00:00 < Date: Wed, 13 Nov 2019 15:43:05 GMT
  DEBUG   2019-11-13 15:43:05.363970176 +00:00 < Content-Type: application/json
[...]

Issue #620, opened 2019-11-11
Not able to install without bash, and not able to uninstall with bash


AvailCat:
It looks like there is no way to run sentry-cli in a clean way, not install, or installed but not able to uninstall.

**System:** Ubuntu 18.04.3 LTS
**Node**: 12.10.0

**To reproduce this, start a clean environment with docker and run commands below to test.**
```
docker run --rm -it node:lts /bin/bash
```
or 
```
docker run --rm it node:lts-buster /bin/bash
```

**Problem 1**:
Install with npm, `root` don't have permission to /root?
```console
root@8b8fe383db83:~# npm i -g @sentry/cli
/usr/bin/sentry-cli -> /usr/lib/node_modules/@sentry/cli/bin/sentry-cli

> @sentry/cli@1.49.0 install /usr/lib/node_modules/@sentry/cli
> node scripts/install.js

Error: EACCES: permission denied, mkdir '/root/.npm/sentry-cli'
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @sentry/cli@1.49.0 install: `node scripts/install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @sentry/cli@1.49.0 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2019-11-11T06_42_00_118Z-debug.log
root@8b8fe383db83:~# 
```

**Problem 2**:
Run with npx
```console
root@8b8fe383db83:~# npx @sentry/cli sentry-cli
internal/modules/cjs/loader.js:775
    throw err;
    ^

Error: Cannot find module '/root/.npm/_npx/10940/lib/node_modules/@sentry/cli/scripts/install.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:772:15)
    at Function.Module._load (internal/modules/cjs/loader.js:677:27)
    at Function.Module.runMain (internal/modules/cjs/loader.js:999:10)
    at internal/main/run_main_module.js:17:11 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @sentry/cli@1.49.0 install: `node scripts/install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @sentry/cli@1.49.0 install script.
[...]

PR #654, opened 2020-02-11
fix: Check if sudo is available before executing cmd with elevated privileges

kamilogorek:
Fixes https://github.com/getsentry/sentry-cli/issues/623
Fixes https://github.com/getsentry/sentry-cli/issues/620


kamilogorek:
Ported to `runas` instead

PR #657, opened 2020-02-12
fix: Use a bundled, statically linked libcurl and openssl

kamilogorek:


CHANGELOG.md
```diff
@@ -1,19 +1,20 @@
 # Changelog
 
+## sentry-cli 1.51.0
+
+* feat: Add `dist` option to `react-native appcenter` command (#653)
+* ref: Notify user about missing `sudo` command instead of incorrect "No such file or directory" when updating/uninstalling `sentry-cli` (#656)
+* fix: Remove `Closing connection 0` warnings after every HTTP request (#657)
```
rhcarvalho: #657 refers to this very PR, is this fixed by bundling curl?!

kamilogorek: Yup, it has been fixed in `curl 7.65.0`, where OSX is shipped with an older version by default. See: https://github.com/curl/curl/commit/d934059afd8568201d0646d6bfd201f39b6bc412

rhcarvalho: Could we throw a reference to this in the commit message? Would make it easier for anyone later trying to understand where the problem was coming from and how it got solved :)

kamilogorek: Will do once merging 👌

kamilogorek:
@rhcarvalho ref: https://github.com/getsentry/sentry-cli/commit/33f037641209eaa5b3ea75d300d8451f2b2af6ec

PR #659, opened 2020-02-12
fix: Verbose curl only in debug mode

rhcarvalho:
Setting verbose output for curl produces annoying output like
'* Closing connection 0'

This change makes sentry-cli less verbose in its normal usage and keeps
the verbose mode of curl and header dumping when --log-level=debug.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1308,8 +1308,19 @@ fn handle_req<W: Write>(
         handle.progress(true)?;
     }
 
-    // enable verbose mode
-    handle.verbose(true)?;
+    if Config::current().get_log_level() == log::LevelFilter::Debug {
+        handle.verbose(true)?;
+
+        handle.debug_function(move |info, data| match info {
+            curl::easy::InfoType::HeaderIn => {
+                log_headers(false, data);
+            }
+            curl::easy::InfoType::HeaderOut => {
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -1308,8 +1308,19 @@ fn handle_req<W: Write>(
         handle.progress(true)?;
     }
 
-    // enable verbose mode
-    handle.verbose(true)?;
+    if Config::current().get_log_level() == log::LevelFilter::Debug {
```
jan-auer: There is also a `trace` log level.

```suggestion
    if Config::current().get_log_level() >= log::LevelFilter::Debug {
```

kamilogorek:
I pointed out where in curl it has been fixed here https://github.com/getsentry/sentry-cli/pull/657#discussion_r378691354
Your call which one should we merge. I don't have enough insight yet.

kamilogorek:
Fixed in https://github.com/getsentry/sentry-cli/pull/657

rhcarvalho:
I wonder if this is useful anyway to reduce output under normal operation? Haven't noticed yet what the debug output adds, though.

kamilogorek:
I haven't seen any yet tbh. Will keep that in mind if I'll stumble upon one

PR #660, opened 2020-02-13
fix: Use new release identifier for react-native xcode upload

HazAT:


jeremywiebe:
What was the reason for this change?  I spent quite a bit of time getting Sentry working in our codebase using the old release naming. Now with this change (which isn't overridable) we will need to re-work our processes that know how to build the release name.

HazAT:
The motivation was the new release health system we have. 
We needed a consistent way (format) for the release to do things like: Identifying the latest release

The current architecture requires everything to be in the same string (which is not ideal).

jeremywiebe:
Thanks for the explanation! :)

flochtililoch:
With this change, how can we distinguish between different bundle ids of the same version?
While the format of the release passed is `{bundleId}@{version}+{dist}`, Sentry UI omit the "bundleId" value, and shows only  `${version} (${dist})`:

Screen Shot 2020-07-15 at 4 48 47 PM

This is a problem for our setup, as we have multiple bundle id (with different source maps) and we need some ways to differentiate the releases (they are effectively not released at the same time, but do share version numbers / dist).

PR #665, opened 2020-02-21
fix: react-native xcode uses regex to detect Debug builds

bjyoungblood:
The default react-native-xcode.sh script invoked builds determines
whether $CONFIGURATION is a debug build or not by loosely matching on
*Debug*. See: https://github.com/facebook/react-native/blob/f19fb2454d164d88aad619fd992edeae51981096/scripts/react-native-xcode.sh#L33

Using a similar regex prevents build failures from occurring anytime a
build configuration has a name that contains but is not exactly "Debug".
This is useful for projects that utilize build configs for different
product flavors (e.g. Debug-dev, Debug-beta, Debug-prod).


kamilogorek:
Thanks!

PR #667, opened 2020-02-24
fix: Override local env.SENTRY_PROPERTIES rather than global process.env

sman591:
Potential fix for https://github.com/getsentry/sentry-webpack-plugin/issues/175

Right now, providing a `configFile` to `SentryCli()` overrides the `process.env.SENTRY_PROPERTIES` on the entire process. This is problematic when multiple instances of `SentryCli` are used in the same Node process, and each uses a different `configFile` parameter.

One potential fix for this is to override `SENTRY_PROPERTIES` at the local execution level, rather that mutating the global `process.env` environment. I'm open to feedback, though!

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -128,10 +128,14 @@ function getPath() {
  * @param {string[]} args Command line arguments passed to `sentry-cli`.
  * @param {boolean} live We inherit stdio to display `sentry-cli` output directly.
  * @param {boolean} silent Disable stdout for silents build (CI/Webpack Stats, ...)
+ * @param {string} [configFile] Relative or absolute path to the configuration file.
  * @returns {Promise.<string>} A promise that resolves to the standard output.
  */
-function execute(args, live, silent) {
+function execute(args, live, silent, configFile) {
   const env = Object.assign({}, process.env);
+  if (configFile) {
+    env.SENTRY_PROPERTIES = configFile;
+  }
   return new Promise((resolve, reject) => {
     if (live === true) {
       const pid = childProcess.spawn(getPath(), args, {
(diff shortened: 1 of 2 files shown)


kamilogorek:
Thanks @sman591! makes perfect sense.

activeguild:
Hi! @sman591 

I'm getting an error if I don't add an environment variable to the global one, is that the reason?

e.g. 
// webpack.config.js
```
process.env.SENTRY_PROPERTIES = sentry.'properties'
export default () => {
  return [
    { 
      plugins: [new SentryPlugin()],
    }
}
```

When calling from the following code
https://github.com/getsentry/sentry-cli/blob/6d9ba9f3e3f061839735e986839d8935672e6121/js/releases/index.js#L42

The parameter `configFile` will be `undefined.`
https://github.com/getsentry/sentry-cli/blob/6d9ba9f3e3f061839735e986839d8935672e6121/js/helper.js#L134

`process.env.SENTRY_PROPERTIES` will not be set to a file name because `configFile` is `undefined`
https://github.com/getsentry/sentry-cli/blob/6d9ba9f3e3f061839735e986839d8935672e6121/js/helper.js#L137

sman591:
Aha, I was seeing something similar and now it makes sense why. The `env.SENTRY_PROPERTIES = configFile;` line is actually safe -- there's a `if (configFile) {` condition wrapped around it. However, when `releases.new()` gets called, it doesn't use the shared `execute()` function in SentryCli -- it uses a brand new helper.

https://github.com/getsentry/sentry-cli/blob/6d9ba9f3e3f061839735e986839d8935672e6121/js/releases/index.js#L42

I didn't realize they don't use the same initialized SentryCli() class, so my testing didn't catch this. I can put up a pull request to fix later today

sman591:
Fix: #700

PR #670, opened 2020-03-03
misc: Show project and organization when using info log level

kamilogorek:


kamilogorek:
Every command has it's own `execute` method, and not all of them require `project` and `organization`. I didn't want to put it inside `code.get_org_and_project` getter, as logging shouldn't be part of getting the value itself. Maybe @jan-auer can chime in?

Issue #663, opened 2020-02-17
Sentry fails to upload debug files due to a duplicated Authorization header


Tymolc:
We are trying to set up a self-hosted Sentry instance. We use a prebuilt Sentry Docker image and have an nginx load balancer connected as well.

When trying to use `sentry-cli` to upload debug files (using `upload-dif` or `upload-dsym`), the upload always fails with a 400 error response. 

The most prominent difference between the successful and the failing requests I can see is that the failing requests have a duplicated Authorization header. According to [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2.2) this is not correct behaviour, so it's expected that some servers will reject the request. 
I tried to send those requests per postman manually and I also got a 400 Bad Request when using two Authorization headers and a 200 Success when using only one.
Is this some kind of faulty setup on our end?

This is the log:

<details>

```
sentry-cli upload-dif ~/Downloads/dysms --log-level debug                                                                                                             
  INFO    2020-02-17 11:33:51.264996 +01:00 Loaded config from /Users/me/.sentryclirc
  DEBUG   2020-02-17 11:33:51.265314 +01:00 sentry-cli version: 1.51.0, platform: "darwin", architecture: "x86_64"
  INFO    2020-02-17 11:33:51.265328 +01:00 sentry-cli was invoked with the following command line: "sentry-cli" "upload-dif" "/Users/me/Downloads/dsyms" "--log-level" "debug"
  DEBUG   2020-02-17 11:33:51.265875 +01:00 request GET https://sentry.backend.dev.myProject.otc/api/0/organizations/sentry/chunk-upload/
  DEBUG   2020-02-17 11:33:51.265901 +01:00 using token authentication
  DEBUG   2020-02-17 11:33:51.265914 +01:00 retry number 0, max retries: 0
  DEBUG   2020-02-17 11:33:51.480542 +01:00 > GET /api/0/organizations/sentry/chunk-upload/ HTTP/1.1
  DEBUG   2020-02-17 11:33:51.480591 +01:00 > Host: sentry.backend.dev.myProject.otc
  DEBUG   2020-02-17 11:33:51.480604 +01:00 > Accept: */*
  DEBUG   2020-02-17 11:33:51.480614 +01:00 > Connection: TE
  DEBUG   2020-02-17 11:33:51.480623 +01:00 > TE: gzip
  DEBUG   2020-02-17 11:33:51.480632 +01:00 > User-Agent: sentry-cli/1.51.0
[...]

PR #651, opened 2020-02-06
Use config generated chunk-upload URL and handle upload_chunks 301|302 statuses

amoskyler:
Proposal to fix #649
I am not a Rust developer, so I apologize if this code is extremely naive.

This successfully attempts to:
1. Rewrite the ChunkUploadOptions.url with a locally generated API endpoint instead of the endpoint returned by `GET /organizations/{}/chunk-upload/`
2. Handles the case where `POST /organizations/{}/chunk-upload/` returns 301 or 302 redirects (the case with proxies/wrong address)

Issue #649, opened 2020-02-05
Inconsistent base url handling while processing source map uploads


amoskyler:
### What's happening
1. sentry-cli overrides the base URL with the value from the `url` field returned by `api/0/organizations/<org>/chunk-upload/`
2. sentry-cli is returned a 302 when posting to the returned URL, however this is not treated as an error - sentry-cli continues on and completes the process as if it was successful.

### What should happen
1. sentry-cli should treat a 302 status code as an error when issuing a POST to `/api/0/organizations/<org>/chunk-upload/`
2. sentry-cli should use the configured base url instead of the fully qualified URL returned by the sentry API. I can see this being done in three ways:
- sentry-cli strips the base url and replaces it with the locally configured value
- sentry-cli totally ignores the returned URL and generates the POST path itself (as it did with the get). I may be missing context on why this URL is returned in the first place though
- Sentry API returns the path sans base url.

### Explanation
I know the above example doesn't make a ton of sense on it's face, but here is the explanation.
I am running an onpremise sentry with a couple of unique properties:
1. The sentry API is exposed internally to my infrastructure at a URL that _differs_ from the configured Root URL.
2. The Root URL points to an authenticated reverse tunnel which CANNOT be authenticated against with username and password headers.

### sentry-cli config
```
SENTRY_URL=http://proxy_host:9000
SENTRY_ORG=<org>
SENTRY_PROJECT=<project>
SENTRY_AUTH_TOKEN=<auth>
SENTRY_LOG_LEVEL=debug
```

Below is stripped debug output from the cli:
```
  DEBUG   2020-02-05 19:51:35.219983500 +00:00 request POST http://proxy_host:9000/api/0/projects/<org>/<project>/releases/
  DEBUG   2020-02-05 19:51:35.220008700 +00:00 using token authentication
  DEBUG   2020-02-05 19:51:35.220022 +00:00 json body: {"version":"test_release_01","projects":["<project>"]}
  DEBUG   2020-02-05 19:51:35.220073900 +00:00 retry number 0, max retries: 0
  DEBUG   2020-02-05 19:51:35.223218 +00:00 > POST /api/0/projects/<org>/<project>/releases/ HTTP/1.1
  DEBUG   2020-02-05 19:51:35.223258900 +00:00 > Host: proxy_host:9000
  DEBUG   2020-02-05 19:51:35.223272100 +00:00 > Accept: */*
[...]

PR #675, opened 2020-03-06
fix: Make sure that requests are not authenticated twice and warn for relative urls

kamilogorek:
Closes: https://github.com/getsentry/sentry-cli/issues/649
Closes: https://github.com/getsentry/sentry-cli/issues/651
Closes: https://github.com/getsentry/sentry-cli/issues/663

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -36,7 +36,7 @@ use uuid::Uuid;
 use crate::config::{Auth, Config};
 use crate::constants::{ARCH, EXT, PLATFORM, RELEASE_REGISTRY_LATEST_URL, VERSION};
 use crate::utils::android::AndroidManifest;
-use crate::utils::http::{self, parse_link_header};
+use crate::utils::http::{self, is_absolute_url, parse_link_header};
 use crate::utils::progress::ProgressBar;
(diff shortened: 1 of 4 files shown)


src/api.rs
```diff
@@ -1054,7 +1055,15 @@ impl Api {
         // to be a Sentry-compatible endpoint, we force the Authorization header at
         // this point.
         let request = match Config::current().get_auth() {
-            Some(auth) => request.with_auth(auth)?,
+            Some(auth) => {
```
jan-auer: You can simplify this:

```suggestion
            Some(auth) if !request.is_authenticated => {
```

src/utils/http.rs
```diff
@@ -63,6 +63,14 @@ pub fn parse_link_header<'a>(s: &'a str) -> Vec<HashMap<&'a str, &'a str>> {
     rv
 }
 
+/// Checkes whether an url starts with http:// or https:// prefix
+pub fn is_absolute_url(url: &str) -> bool {
+    if url.starts_with("http://") || url.starts_with("https://") {
```
jan-auer: Oh, and also this:

```
pub fn is_absolute_url(url: &str) -> bool {
    url.starts_with("http://") || url.starts_with("https://")
}
```

PR #683, opened 2020-03-11
feat: Detect GitHub Actions slug for propose-version

hinaloe:
Suggests support for GitHub Actions hash as well as CodeBuild (#681) and Heroku.


hinaloe:
un... Thinking calmly, this may not be very useful

Issue #120, opened 2017-09-19
support `--save-global` flag to have login auth details saved in global .sentryclirc


G-Rath:
It would be great if we had something like a `--save-global` option we could use to have the sentry-cli write the config defaults for [auth] to a more global per-computer/per-use `.sentryclirc`.

We're using Sentry with a lot of projects, so it makes sense for us to commit a `.sentryclirc` file with the project name, to make it easier for both developers, and automation tools such as BitBucket Pipelines.

Currently when we do `sentry-cli login`, it will save the key under the [auth] field of the local per-project `.sentryclirc` file, and in turn get picked up by our VCS as a change to the file and want to commit to the repo, which is something we don't want. 

It would be great if there was a flag that could be provided to instead have the sentry-cli use the default global `.sentryclirc` file (located in the user folder on Windows).

PR #690, opened 2020-03-19
feat: Introduce g/global flag for login command

kamilogorek:
Closes: https://github.com/getsentry/sentry-cli/issues/120

### DIFF
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -24,8 +24,8 @@ fn update_config(config: &Config, token: &str) -> Result<(), Error> {
 }
 
 pub fn execute<'a>(matches: &ArgMatches<'a>) -> Result<(), Error> {
-    let local_config = Config::current();
-    let token_url = format!("{}/api/", local_config.get_base_url()?);
+    let config = Config::current();
+    let token_url = format!("{}/api/", config.get_base_url()?);
 
     println!("This helps you signing in your sentry-cli with an authentication token.");
     println!("If you do not yet have a token ready we can bring up a browser for you");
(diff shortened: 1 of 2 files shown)


src/commands/login.rs
```diff
@@ -58,9 +63,17 @@ pub fn execute<'a>(_matches: &ArgMatches<'a>) -> Result<(), Error> {
         }
     }
 
-    update_config(&config, &token)?;
-    println!();
-    println!("Stored token in {}", config.get_filename().display());
+    let mut config_to_update = local_config;
+    if matches.is_present("global") {
+        let (global_filename, global_config) = load_global_config_file()?;
+        config_to_update = Config::from_file(global_filename, global_config)?.bind_to_process();
```
kamilogorek: Is there a better way to get an `Arc` reference here?

src/commands/login.rs
```diff
@@ -58,9 +63,17 @@ pub fn execute<'a>(_matches: &ArgMatches<'a>) -> Result<(), Error> {
...
+    }
+
+    update_config(&config_to_update, &token)?;
+    println!(
+        "\nStored token in {}",
```
jan-auer: _nit_: Rather than leading with a newline, could you add a `println!();` above?

src/commands/login.rs
jan-auer: Which part exactly? You could probably:

- Add a function `Config::global()?` which runs the above two lines 
- Remove bind_to_process, which you probably don't need?

That way you leave functions like `load_global_config_file` internal, and you just expose the `Config` object.

src/commands/login.rs
kamilogorek: It was like this before and I changed it because it looked unnecessary. Will change it back.

Issue #372, opened 2018-09-06
sentry-cli react-native appcenter breaks for promotions to production; fix: packageHash


enthal:
OS:
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

_Platform:_
- [x]  iOS
- [ ]  Android

---
# tl;dr
 `sentry-cli react-native appcenter` marks release with codepush *`label`*, but that changes in codepush when `staging` is promoted to `production`.  Solution: use `packageHash`, which remains constant and always reflects the actual code.

# details
Per your instructions here: 
https://docs.sentry.io/clients/react-native/codepush/

When we push code+maps to sentry, the `sentry-cli react-native appcenter` command tags it with a release/version that ends in the codepush version label.  Since the app can get these values from codepush at runtime via
`update.appVersion + '-codepush:' + update.label`
we have a match and we see code for app errors in the sentry web console.

## HERE'S THE PROBLEM:
That only works for staging!  When we `code-push promote` to production, the `label` changes to a new number... not determined from the staging `label`.  The match is broken.

## HERE'S THE SOLUTION:
If you used the `packageHash` instead of the `label`, both of which are available in the same places, the problem goes away!  Unlike the label, the packageHash stays the same when the release is promoted to production!  This follows the design of code-push: a code bundle from the local filesystem is released to staging, but only selected in the codepush server-side for production.  Sentry should follow this: the uploaded bundle files match the codepush packageHash, and may be used for error code display regardless of staging or production.

I can't fix this on my own (without forking sentry-cli), since `sentry-cli react-native appcenter` appends the label.  There's no way for the app to get the staging label when running a production codepush update.

Have other's found some other solution here?  I can't be the only one.

For completeness, here's a way to form the sentry release/version we need, from the command line.  It doesn't solve the problem since sentry-cli appends label on its own.
```
code-push deployment history Org/App Staging --format json | node -e "
  fs=require('fs');
  data = JSON.parse(fs.readFileSync('/dev/stdin'))
  last = data[data.length-1]
  console.log(last.appVersion+'-'+last.packageHash.slice(0,16))
"
````

kamilogorek:
Is this still a thing that `label` changes between staging/production? I cannot find any reasonable information about that in codepush documentation.
I'll close the issue due to it being outdated, but please do not hesitate to ping me if it is still relevant, and I will happily reopen and work on it.
Cheers!

j-fro:
[...]

PR #692, opened 2020-03-20
Add an optional argument to override the entire release name for a CodePush release

j-fro:
Add `--release-name` as an optional argument to `react-native appcenter` to override the entire release name. This can be used to get around issues with the CodePush package label (used by the CLI by default) not being consistent between CodePush environments (#372 ) among other uses.


kamilogorek:
Can you add `.conflicts_with_all(&["bundle_id", "version_name"])`?

kamilogorek:
Thanks!

PR #693, opened 2020-03-23
misc: Spring dependencies update

kamilogorek:
Encoding update is based on https://docs.rs/crate/percent-encoding/1.0.1/source/lib.rs
`uuid` cannot be updated due to `proguard` `symbolic-common` and `debugid` depending on version `0.7`.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -44,12 +44,7 @@ use crate::utils::ui::{capitalize_string, make_byte_progress_bar};
 use crate::utils::xcode::InfoPlist;
 
 const QUERY_ENCODE_SET: &AsciiSet = &CONTROLS.add(b' ').add(b'"').add(b'#').add(b'<').add(b'>');
-const DEFAULT_ENCODE_SET: &AsciiSet = &CONTROLS
-    .add(b' ')
-    .add(b'"')
-    .add(b'#')
-    .add(b'<')
-    .add(b'>')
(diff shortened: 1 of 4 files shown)


src/api.rs
```diff
@@ -43,6 +43,18 @@ use crate::utils::sourcemaps::get_sourcemap_reference_from_headers;
 use crate::utils::ui::{capitalize_string, make_byte_progress_bar};
 use crate::utils::xcode::InfoPlist;
 
+const QUERY_ENCODE_SET: &AsciiSet = &CONTROLS.add(b' ').add(b'"').add(b'#').add(b'<').add(b'>');
```
jan-auer: Isn't this missing the condition from `SIMPLE_ENCODE_SET`?

```
byte < 0x20 || byte > 0x7E
```

jan-auer: Nvm, this is in `CONTROLS`.

src/api.rs
```diff
@@ -43,6 +43,18 @@ use crate::utils::sourcemaps::get_sourcemap_reference_from_headers;
...
+const DEFAULT_ENCODE_SET: &AsciiSet = &CONTROLS
```
jan-auer: Could you add to `QUERY_ENCODE_SET` instead of repeating?

kamilogorek: Wasn't sure if it's the same reference or does builder always return new instance.

Issue #672, opened 2020-03-04
Provide binary for FreeBSD


versedi:
Hello,

Would it be possible to provide the executable binary for FreeBSD x86/x64?

kamilogorek:
Hey, there is no plan to support that on our roadmap, but we are open for the PR introducing it :)

mazafak:
the binary builds properly without changes. all that needs to be done is:
- binary uploaded to https://downloads.sentry-cdn.com/sentry-cli/
- referenced in https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js#L45

kamilogorek:
Thanks for the valuable feedback @mazafak! I'll try to add it once I find some spare time :)

oswaldofreitas:
@kamilogorek I'm also looking forward to this change, thank you in advance for applying @mazafak suggestion, it will be a great value for us since we can't upload sourcemaps with the webpack plugin from FreeBSD currently

djhype:
@kamilogorek Can you share whether this is on your roadmap now? If so, where would an announcement be made when support has been added for a FreeBSD binary? In this issue, or somewhere else?

zilti:
Yes, either upload it, or fix sentry-cli so it will work properly even if the binary is not available at that URL. I installed sentry-cli from the FreeBSD ports, and it works fine in that it causes the effects on sentry.io it is supposed to cause, but at the same time, it makes it seem like all hell broke loose on the command line:

```
  INFO    2020-12-09 13:52:05.206194336 +00:00 Running update nagger update check
  DEBUG   2020-12-09 13:52:05.206220974 +00:00 request GET https://release-registry.services.sentry.io/apps/sentry-cli/latest
  DEBUG   2020-12-09 13:52:05.206229559 +00:00 retry number 0, max retries: 0
  DEBUG   2020-12-09 13:52:05.221478812 +00:00 > GET /apps/sentry-cli/latest HTTP/2
  DEBUG   2020-12-09 13:52:05.221568066 +00:00 > Host: release-registry.services.sentry.io
  DEBUG   2020-12-09 13:52:05.221604298 +00:00 > accept: */*
  DEBUG   2020-12-09 13:52:05.221636703 +00:00 > connection: TE
  DEBUG   2020-12-09 13:52:05.221669418 +00:00 > te: gzip
  DEBUG   2020-12-09 13:52:05.221747605 +00:00 > user-agent: sentry-cli/1.54.0
[...]

PR #710, opened 2020-04-08
Add FreeBSD binary download URL

oswaldofreitas:
The current url for the FreeBSD binary is using the Linux binary and it's not working. I've updated it to get directly from the URL pointed in https://github.com/getsentry/sentry-cli/issues/672 and it should be updated when the binary is uploaded to the default releases url.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -42,8 +42,10 @@ function getDownloadUrl(platform, arch) {
     case 'win32':
       return `${releasesUrl}-Windows-${archString}.exe`;
     case 'linux':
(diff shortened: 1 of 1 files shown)


kamilogorek:
Hey, thanks for the contribution. Unfortunately, we cannot point to a 3rd party server for such an important process as installation of the binary. We'll have to take care of the CI build ourselves first and host it on our CDN.
I'm not able to tell when it'll be done, but I'll make sure to let everyone know once it is.
Thanks!

Issue #723, opened 2020-04-29
`releases` requiring `--org`, but fails when argument is passed


someguynamedmatt:
`sentry-cli`: `1.52.3`
`MacOS`: `10.14.6`

I'm following the [documentation](https://docs.sentry.io/workflow/releases/?platform=javascript#create-release) to create releases, but it won't accept the args I'm passing. Specifically, it's complaining that I'm not passing an `org`, so I try that with (as suggested) `--org`. It will then throw an error saying it doesn't recognize the option:

> without org being passed
```bash
$ sentry-cli releases new $VERSION --project my-project

error: An organization slug is required (provide with --org)
```

> with org being passed
```bash
$ sentry-cli releases new $VERSION --project my-project --org my-organization

error: Found argument '--org' which wasn't expected, or isn't valid in this context
```

jan-auer:
This is because the argument is given in the wrong place. It should be:

```
sentry-cli releases --org my-org new $VERSION --project my-project
```

someguynamedmatt:
Ah, okay. That does seem to work for me. However, I'm confused because [in config.rs](https://github.com/getsentry/sentry-cli/blob/121f1b75e468f21378b2993e369b762f2c05ddf1/src/config.rs#L274) it appears that it should be reading it from the `env` var `SENTRY_ORG`, which is set in my case.

Can the documentation to be made more clear to help future users out? 

Thanks for taking the time clarify!

jan-auer:
> it appears that it should be reading it from the `env` var `SENTRY_ORG`

That's right. For me it also does work, at least when I invoke like this:

```
SENTRY_ORG=foobar sentry-cli releases list
```

To test this, I've removed the `org=` key from my `.sentryclirc` to ensure there's nothing configured. Even without removing, the environment takes precedence. Does this not work in your case?

> Can the documentation to be made more clear to help future users out?

Definitely. What part of the documentation can we extend, or where would you expect this? We have the [Working with Projects](https://docs.sentry.io/cli/configuration/#sentry-cli-working-with-projects) section, but that's indeed a bit hard to find.

someguynamedmatt:
@jan-auer Okay. I figured out why my env var wasn't being picked (TL;DR: I didn't `export` the var in my shell). 

[...]

PR #724, opened 2020-04-29
fix: fixes org not being parsed for subcommands

someguynamedmatt:
Relates to Issue #723 

Following [the docs](https://docs.sentry.io/workflow/releases/?platform=javascript#create-release) for creating a new release was giving errors despite passing in the requested `org` command-line arg. I (think I) tracked it down to `new` not expecting the `organization` in its argument parsing despite searching for it on invocation.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,16 +268,33 @@ impl Config {
 
     /// Given a match object from clap, this returns the org from it.
     pub fn get_org(&self, matches: &ArgMatches<'_>) -> Result<String, Error> {
-        Ok(matches
-            .value_of("org")
-            .map(str::to_owned)
-            .or_else(|| env::var("SENTRY_ORG").ok())
-            .or_else(|| {
-                self.ini
-                    .get_from(Some("defaults"), "org")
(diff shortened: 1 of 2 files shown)


someguynamedmatt:
@jan-auer Updated. Thanks for looking at this. I've changed the code slightly to check the top-level command first then recursively checking the subcommands for the `org` value.

jan-auer:
> I've changed the code slightly to check the top-level command first then recursively checking the subcommands for the `org` value.

Thanks! This surprises me, but it seems that we're hitting https://github.com/clap-rs/clap/issues/1570. With `global(true)` the value should just magically appear without changes in `get_org`. However, due to the bug, you'd have to make those changes.

Would you mind waiting until the upstream issue gets resolved? Otherwise, we'd have to check at every nesting level of subcommands, which seems rather dirty.

someguynamedmatt:
@jan-auer Oh, yeah, that would be a much better solution. I was wondering why `global` didn't seem to solve the problem. At least it's a known issue. If that upstream fix fixes this issue, all the better.

PR #753, opened 2020-06-01
feat(set-manual-commits): Create a new release version

NisanthanNanthakumar:
When you run the command `sentry-cli releases set-manual-releases` it should run the POST request to create a new version if there are default projects set in `.sentryclirc`.

### DIFF
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -516,6 +516,24 @@ fn execute_set_commits<'a>(
     Ok(())
 }
 
+fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches<'a>) -> Result<(), Error> {
+    let version = matches.value_of("version").unwrap();
+    let org = ctx.get_org()?;
+
+    if let Ok(projects) = ctx.get_projects(matches) {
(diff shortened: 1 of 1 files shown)


src/commands/releases.rs
```diff
@@ -516,6 +516,24 @@ fn execute_set_commits<'a>(
     Ok(())
 }
 
+fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches<'a>) -> Result<(), Error> {
+    let version = matches.value_of("version").unwrap();
+    let org = ctx.get_org()?;
+
+    if let Ok(projects) = ctx.get_projects(matches) {
```
untitaker, line 523: Not sure if this error should be handled differently. If we don't know if this matters I would even opt for `unwrap()` or `?`

untitaker:
```
let projects = ctx.get_projects(matches)?;
```

untitaker: This was apparently copied from another fn which may have to be fixed too

kamilogorek:
Moved to https://github.com/getsentry/sentry-cli/pull/763

PR #756, opened 2020-06-03
ref: Switch from symbolic::proguard to proguard crate

Swatinem:


src/utils/dif.rs
```diff
@@ -131,17 +130,34 @@ impl fmt::Display for DifFeatures {
...
 }
 
 impl DifFile<'static> {
     fn open_proguard<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
-        let data = ByteView::open(&path).map_err(SyncFailure::new)?;
-        let pg = ProguardMappingView::parse(data).map_err(SyncFailure::new)?;
+        let data = ByteView::open(path).map_err(SyncFailure::new)?;
+        let pg = SelfCell::new(data, |d| SelfProguard(ProguardMapping::new(unsafe { &*d })));
 
-        if path.as_ref().extension() == Some(OsStr::new("txt")) || pg.has_line_info() {
```
Swatinem, line 144: this is technically a change in behavior. Before this change, the cli would erroneously flag *any* `.txt` file as proguard. Not you will get a `Unsupported file` error.

kamilogorek: I think that's fine though

PR #759, opened 2020-06-08
fix: Handle 301|302 for dsym uploads

kamilogorek:


kamilogorek:
cc @jan-auer as I'm not sure if I understood this flow correctly

kamilogorek:
Let's close this for now, and reiterate when it pops up again, as I don't know enough details of this issue just yet.

PR #763, opened 2020-06-16
feat(set-manual-commits): Update release with commits.

NisanthanNanthakumar:
Sorry in advance, this kinda turned into a big PR, but it's mostly the snapshot tests.

## Objective
In this PR, we have 4 parts.

**1. Get the commits for the current release.**
We make an API call to `/organizations/{}/releases/{}/previous-with-commits/` to get the previous release with commits. From this we will either have a valid SHA or null hereinafter referred to as `prev_commit`.

Then we traverse down the git tree up to `prev_commit` inclusive or to the default of 20. If there is a previous commit but cannot find it in git history, throw an error.

**2. Get the patch set for each commit in this release**
From the array of commits, calculate the diff between the commits. This is the patch set. 

**3. Parse the git url to returns a string with the format `owner-name/repo-name`.** 
There already exists a well-tested private function that generates this string. I opened up the API with a public function `parse_git_url`.

**4. Update the release with the commits.**
Break up the final array of commits into chunks of 50 and make consecutive call to update the release.

## Test Plan
I added snapshot testing with Insta for the `get_commits_from_git` function and `generate_patch_set` function.


src/api.rs
```diff
@@ -2239,3 +2275,21 @@ pub struct AssembleArtifactsResponse {
...
+    pub path: String,
+    #[serde(rename = "type")]
+    pub ty: String,
+}
+
+#[derive(Debug, Serialize, Clone)]
+pub struct GitCommit {
+    pub patch_set: Vec<PatchSet>,
+    pub repository: String,
+    pub author_name: Option<String>,
```
untitaker, line 2290: Would add `#[serde(skip_serializing_if = "Option::is_none")]` if possible

src/commands/releases.rs
```diff
@@ -531,9 +532,48 @@ fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches
         )?;
     }
 
+    // Get the commit of the most recent release.
+    let prev_commit = match ctx.api.get_previous_release_with_commits(org, version)? {
+        OptionalReleaseInfo::Some(prev) => prev.last_commit.map(|c| c.id).unwrap_or_default(),
+        OptionalReleaseInfo::None{} => "".to_string(),
```
untitaker, line 538:
```suggestion
        OptionalReleaseInfo::None{} => String::new(),
```

src/commands/releases.rs
```diff
@@ -531,9 +532,48 @@ fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches
...
+    let remote = repo.find_remote("origin")?;
+    let url = remote.url();   
+    let parsed = match url {
+        Some(url) => parse_git_url(url),
+        None => return Err(format_err!("Failed to parse remote url!"))
+    };
+
+    // Fetch all the commits upto the `prev_commit` or return the default (20).
+    // Will return a tuple of Vec<GitCommits> and the `prev_commit` if it exists in the git tree.
+    let commit_log = get_commits_from_git(&repo, &prev_commit)?;
```
untitaker, line 554:
```suggestion
    let (commit_log, prev_commit) = get_commits_from_git(&repo, &prev_commit)?;
```

src/commands/releases.rs
```diff
@@ -531,9 +532,48 @@ fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches
...
+    
+    // Calculate the diff for each commit in the Vec<GitCommit>.
+    let commits = generate_patch_set(&repo, commit_log.0, commit_log.1, &parsed)?;
```
untitaker, line 557:
```suggestion
    let commits = generate_patch_set(&repo, commit_log, prev_commit, &parsed)?;
```

src/utils/vcs.rs
```diff
@@ -427,8 +434,136 @@ pub fn find_heads(
...
+                ));
+            }
+            let prev = result.pop();
+            Ok((result, prev))
+        }
+        Err(_) => {
+            // If there is no previous commit, return the default number of commits
+            let mut result: Vec<Commit> = revwalk
+                .take(default_count + 1)
+                .filter_map(move |id| {
```
untitaker, line 485: I would dedupe `filter_map` argument into one variable defined beforehand

src/api.rs
```diff
@@ -2239,3 +2275,22 @@ pub struct AssembleArtifactsResponse {
     pub missing_chunks: Vec<Digest>,
     pub detail: Option<String>,
 }
+
+#[derive(Debug, Serialize, Clone)]
+pub struct PatchSet {
+    pub path: String,
+    #[serde(rename = "type")]
+    pub ty: String,
```
kamilogorek, line 2283: Why not just use `type` instead and skip the renaming?

src/commands/releases.rs
```diff
@@ -118,6 +119,15 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                         the current commit of the repository at the given PATH is \
                         assumed.  To override the revision `@REV` can be appended \
                         which will force the revision to a certain value.")))
+        .subcommand(App::new("set-manual-commits")
+            .about("Set commits of a release from local git.")
+            .version_arg(1)
+            .arg(Arg::with_name("set-initial")
```
kamilogorek, line 125: I'd rename it to something like `count` or `commits-count`

src/commands/releases.rs
```diff
@@ -507,6 +517,63 @@ fn execute_set_commits<'a>(
     Ok(())
 }
 
+fn execute_set_manual_commits<'a>(ctx: &ReleaseContext<'_>, matches: &ArgMatches<'a>) -> Result<(), Error> {
+    let version = matches.value_of("version").unwrap();
+    let org = ctx.get_org()?;
+
+    if let Ok(projects) = ctx.get_projects(matches) {
```
kamilogorek, line 524: I'd copy the comment from `execute_set_commit` to make sure it's clear:

> // make sure the release exists if projects are given

src/commands/releases.rs
```diff
@@ -507,6 +517,63 @@ fn execute_set_commits<'a>(
...
+    let prev_commit = match ctx.api.get_previous_release_with_commits(org, version)? {
+        OptionalReleaseInfo::Some(prev) => prev.last_commit.map(|c| c.id).unwrap_or_default(),
+        OptionalReleaseInfo::None{} => String::new(),
+    };
+
+    // Find and connect to local git.
+    let repo = git2::Repository::open_from_env()?;
+
+    // Parse the git url.
+    let remote = repo.find_remote("origin")?;
```
kamilogorek, line 545: Use `config.get_cached_vcs_remote()` instead. This way you'll get free remote configuration feature.

src/commands/releases.rs
```diff
@@ -507,6 +517,63 @@ fn execute_set_commits<'a>(
...
+    };
+
+    // Fetch all the commits upto the `prev_commit` or return the default (20).
+    // Will return a tuple of Vec<GitCommits> and the `prev_commit` if it exists in the git tree.
+    let (commit_log, prev_commit) = get_commits_from_git(&repo, &prev_commit)?;
+    
+    // Calculate the diff for each commit in the Vec<GitCommit>.
+    let commits = generate_patch_set(&repo, commit_log, prev_commit, &parsed)?;
+
+    for chunk in commits.chunks(50) {
```
kamilogorek, line 559: Magic number could be extracted and named to be more descriptive (and most likely configurable in the future).

src/utils/vcs.rs
```diff
@@ -198,6 +200,11 @@ fn is_matching_url(a: &str, b: &str) -> bool {
     VcsUrl::parse(a) == VcsUrl::parse(b)
 }
 
+pub fn parse_git_url(repo: &str) -> String {
```
kamilogorek, line 203: This function name is not too precise. It does parsing, but it returns only `id`, nothing else.
So maybe a better name would be something like `get_id_from_remote` or something of this sort.

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+// Get commits from git history upto previous commit.
+// Returns a tuple of Vec<GitCommits> and the `prev_commit` if it exists in the git tree.
+pub fn get_commits_from_git<'a>(
+    repo: &'a Repository,
+    prev_commit: &str,
+) -> Result<(Vec<Commit<'a>>, Option<Commit<'a>>), Error> {
+    let mut revwalk = repo.revwalk()?;
+    revwalk.push_head()?;
+
+    let default_count = 20;
```
kamilogorek, line 446: This should be passed as the argument from the caller.
And the caller should use `matches.value_of("arg-name").unwrap_or(20)`

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+    let commit_filter = move |id:Result<git2::Oid, git2::Error>| {
```
kamilogorek, line 447: I'd inline it, as it's very short filter

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+                    Ok(id) => {
+                        if found {
+                            return false;
+                        }
+                        if prev == *id {
+                            found = true;
+                        }
+                        true
+                    }
+                    _ => true,
```
kamilogorek, line 469: How can `id` be absent? If so, shouldn't we stop the iteration?

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+            let mut result: Vec<Commit> = revwalk
+                .take_while(|id| match id {
+                    Ok(id) => {
+                        if found {
+                            return false;
+                        }
+                        if prev == *id {
+                            found = true;
+                        }
+                        true
```
kamilogorek, line 467:
```suggestion
                        if prev == *id {
                            found = true;
                            return false;
                        }
                        true
```

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+                    }
+                    _ => true,
+                })
+                .filter_map(commit_filter)
+                .collect();
```
kamilogorek, line 472: You don't need to `collect` if the only thing you are doing is popping at the end.
Skip `collect()` call and change `pop` to `last()`

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+pub fn get_commits_from_git<'a>(
+    repo: &'a Repository,
+    prev_commit: &str,
+) -> Result<(Vec<Commit<'a>>, Option<Commit<'a>>), Error> {
+    let mut revwalk = repo.revwalk()?;
+    revwalk.push_head()?;
+
+    let default_count = 20;
+    let commit_filter = move |id:Result<git2::Oid, git2::Error>| {
+        let id = id.ok()?;
```
kamilogorek, line 448: You already make sure in `take_while` that `id` is present.

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+            }
+            let prev = result.pop();
+            Ok((result, prev))
+        }
+        Err(_) => {
+            // If there is no previous commit, return the default number of commits
+            let mut result: Vec<Commit> = revwalk
+                .take(default_count + 1)
+                .filter_map(commit_filter)
+                .collect();
```
kamilogorek, line 487: Same here. No need to collect it.

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+            author_email: commit.author().email().map(|s| s.to_owned()),
+            message: commit.message().map(|s| s.to_owned()),
+            repository: repository.to_string(),
+            timestamp: get_commit_time(commit.author().when()),
+            patch_set: vec![],
+        };
+
+        let tree1 = commit.tree()?;
+
+        let tree2 = if index + 1 < commits.len() {
```
kamilogorek, line 519:
```suggestion
        let tree2 = if commits.len() > index + 1 {
```

I always prefer `>` instead of `<`, easier to read imo

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+            Some(commits[index + 1].tree()?)
+        } else {
+            match &previous {
```
kamilogorek, line 522: `&previous.map(|c| c.tree()?)`

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+        diff.print(git2::DiffFormat::NameStatus, |_, _, l| {
+            let line = std::str::from_utf8(l.content()).unwrap();
+            let mut parsed = line.trim_end().splitn(2, '\t');
+            let patch_set = PatchSet {
+                ty: parsed.next().unwrap().to_owned(),
+                path: parsed.next().unwrap().to_owned(),
+            };
+            git_commit.patch_set.push(patch_set);
+
+            true
```
kamilogorek, line 539: Please leave a comment what this boolean does

src/utils/vcs.rs
```diff
@@ -427,8 +434,131 @@ pub fn find_heads(
...
+        })?;
+
+        result.push(git_commit)
+    }
+
+    Ok(result)
+}
+
+pub fn get_commit_time(time: Time) -> DateTime<FixedOffset> {
+    FixedOffset::east(time.offset_minutes() * 60).timestamp(time.seconds(), 0)
```
kamilogorek, line 549: I don't know the logic behind commits, so I'll trust you on this one :3

src/utils/vcs.rs
```diff
@@ -668,3 +798,443 @@ fn test_url_normalization() {
...
+        "\"third commit\"",
+    );
+
+    let repo = git2::Repository::open(dir.path()).expect("Failed");
+    let commits = get_commits_from_git(&repo, "").expect("Failed");
+    println!("Commits: {}", commits.0.len());
+    let patch_set =
+        generate_patch_set(&repo, commits.0, commits.1, "example/test-repo").expect("Failed");
+
+    assert_yaml_snapshot!(patch_set, {
```
kamilogorek, line 916: I think we can store these in separate files. You can ask Armin about how to do that.

src/api.rs
NisanthanNanthakumar: @kamilogorek I believe `type` is a reserved keyword. We do something similar here: `sentry-cli/src/commands/difutil_find.rs` line 29

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek I agree, but I like `get_repo_from_remote` because the `id` in VcsUrl struct returns in the format `"owner-name/repo-name"` which is the repo path.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek Markus suggested to extract the filter. Do you think inlining is better for maintainability/readability? What about the code duplication?

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek not sure if there's a scenario where the git id is absent. But the `id` in the tree has type `Result<Oid, Error>`. It's just handling the possibility of an error and then we filter it out in the `filter_map`.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek this change fails a test because it doesn't include the commit where `prev == *id` in the result.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek I tried doing the suggestion of removing `collect` but got a compiler error of mismatched types.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek actually in `take_while`, if `id` does not exist, we still continue in the loop. So we need to check in the `filter_map`.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek Same here. I got error of `mismatched types`.

src/utils/vcs.rs
NisanthanNanthakumar: @kamilogorek not sure why but got an error of `incompatible types`.

src/api.rs
kamilogorek: 👌

src/utils/vcs.rs
kamilogorek: Let's leave it. I only realised now that it's used in more than 1 place.

src/utils/vcs.rs
kamilogorek: Oh right, off by one. Nvm then.

src/utils/vcs.rs
kamilogorek: You are right, I didn't realize that you pass whole result later on in the ok tuple

Issue #769, opened 2020-06-22
Release starting with - is treated as an argument


Multiply:
When using `@sentry/cli` in a javascript project, you can't have a release starting with `-`, as it's forwarded incorrectly as an argument.

My proposal is to escape/quote the release, when calling `this.execute` in `js/releases/index.js`.

This was found when using [next.js + sentry-webpack-plugin](https://github.com/getsentry/sentry-webpack-plugin/issues/197).

PR #770, opened 2020-06-22
fix: Allow for release names with leading hyphen

kamilogorek:
We should use https://docs.rs/clap/2.33.1/clap/struct.Arg.html#method.allow_hyphen_values but it's currently broken for positional arguments https://github.com/clap-rs/clap/issues/1437 https://github.com/clap-rs/clap/issues/1287
so we have to use https://docs.rs/clap/2.33.1/clap/enum.AppSettings.html#variant.AllowLeadingHyphen instead.

closes: https://github.com/getsentry/sentry-cli/issues/769
ref: https://github.com/getsentry/sentry-webpack-plugin/issues/197

### DIFF
--- a/src/utils/args.rs
+++ b/src/utils/args.rs
@@ -3,6 +3,7 @@
 use std::str::FromStr;
 
 use chrono::{DateTime, TimeZone, Utc};
+use clap::AppSettings;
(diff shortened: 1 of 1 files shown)


kamilogorek:
image

PR #772, opened 2020-06-22
feat: mv set-manual-commits into set-commits --auto

NisanthanNanthakumar:
## Objective
In this PR, we are moving the code in `set-manual-commits` into `set-commits`. When you run `set-commits --auto` if the user does not have a repo based integration, run the `set-manual-commits` code.
We have agreed to use `--local` optional flag to represent always using the local git tree to get the commits and to use `--initial-depth` as the optional flag for the number of commits for the initial release.

### DIFF
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -100,18 +100,20 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                  .long("clear")
                  .help("Clear all current commits from the release."))
             .arg(Arg::with_name("auto")
-                 .long("auto")
-                 .help("Enable completely automated commit management.{n}\
+                .long("auto")
+                .help("Enable completely automated commit management.{n}\
                         This requires that the command is run from within a git repository.  \
                         sentry-cli will then automatically find remotely configured \
                         repositories and discover commits."))
             .arg(Arg::with_name("local")
+                .conflicts_with_all(&["auto", "clear", "commits", ])
                 .long("local")
                 .help("Set commits of a release from local git.{n}\
                         This requires that the command is run from within a git repository.  \
                         sentry-cli will then automatically find remotely configured \
                         repositories and discover commits."))
             .arg(Arg::with_name("initial-depth")
+                .conflicts_with("auto")
                 .long("initial-depth")
                 .value_name("INITIAL DEPTH")
                 .validator(validate_int)
@@ -430,10 +432,7 @@ fn execute_set_commits<'a>(
     matches: &ArgMatches<'a>,
 ) -> Result<(), Error> {
     let version = matches.value_of("version").unwrap();
-    let default_count = matches
-        .value_of("initial-depth")
-        .unwrap_or("20")
-        .parse::<usize>()?;
+
     let org = ctx.get_org()?;
     let repos = ctx.api.list_organization_repos(org)?;
     let mut commit_specs = vec![];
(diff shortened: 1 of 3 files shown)


src/commands/releases.rs
```diff
@@ -104,6 +105,17 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                         This requires that the command is run from within a git repository.  \
                         sentry-cli will then automatically find remotely configured \
                         repositories and discover commits."))
+            .arg(Arg::with_name("local")
```
kamilogorek, line 108: Should include `.conflicts_with("auto")` and/or `.conflicts_with_all()` for other supporting arguments.

kamilogorek: Same, but inverse for `auto`

src/commands/releases.rs
```diff
@@ -428,6 +430,10 @@ fn execute_set_commits<'a>(
     matches: &ArgMatches<'a>,
 ) -> Result<(), Error> {
     let version = matches.value_of("version").unwrap();
+    let default_count = matches
```
kamilogorek, line 433: It's unnecessary to parse it here, as it's relatable only for `--local` code path. Move it there instead.

src/utils/vcs.rs
```diff
@@ -481,6 +485,10 @@ pub fn get_commits_from_git<'a>(
         }
         Err(_) => {
             // If there is no previous commit, return the default number of commits
+            println!(
+                "Could not find previous commit. We will create a release with {} commits",
```
kamilogorek, line 489:
```suggestion
                "Could not find the previous commit. Creating a release with {} commits.",
```

src/commands/releases.rs
```diff
@@ -443,21 +449,14 @@ fn execute_set_commits<'a>(
...
-            bail!(
-                "Could not determine any commits to be associated automatically.\n\
-                 Please provide commits explicitly using --commit | -c.\n\
-                 \n\
-                 HINT: Did you add the repo to your organization?\n\
-                 Configure it at {}/settings/{}/repos/",
-                config.get_base_url()?,
-                org
-            );
+            None
```
kamilogorek, line 452: Why did the logic change here? I thought `--auto` should keep working as before?

src/commands/releases.rs
```diff
@@ -516,67 +515,44 @@ fn execute_set_commits<'a>(
         }
         ctx.api.set_release_refs(&org, version, heads)?;
     } else {
-        println!("No commits found. Leaving release alone.");
-    }
+        println!("Could not determine any commits to be associated with a repo-based integration. Proceeding to find commits from local git tree.");
```
kamilogorek, line 518: This is also a feature that should be enabled for `--local` only. Right now, when someone uses `--auto` and head will be missing, it'll execute this code path as well.

src/utils/vcs.rs
```diff
@@ -237,7 +237,7 @@ fn find_reference_url(repo: &str, repos: &[Repo]) -> Result<String, Error> {
     }
 
     if found_non_git {
-        bail!("For non git repositories explicit revisions are required");
+        Ok(None)
```
kamilogorek, line 240: `found_`, yet return value is `Ok(None)`, it's confusing.

src/commands/releases.rs
NisanthanNanthakumar: I checked clap docs and its says that we do not need to define the conflict for both arguments. Apparently once is sufficient.

src/commands/releases.rs
NisanthanNanthakumar: If we cannot find commits associated with a repo-based integration, the new expected behaviour is to automatically push commits from local git tree. That's why I removed the bail.

src/utils/vcs.rs
NisanthanNanthakumar: True, changing it to just `non_git`.

src/commands/releases.rs
```diff
@@ -515,7 +514,14 @@ fn execute_set_commits<'a>(
         }
         ctx.api.set_release_refs(&org, version, heads)?;
     } else {
-        println!("Could not determine any commits to be associated with a repo-based integration. Proceeding to find commits from local git tree.");
+        let default_count = matches
+            .value_of("initial-depth")
+            .unwrap_or("20")
+            .parse::<usize>()?;
+
+        if matches.is_present("local") {
```
kamilogorek, line 522: You log this only when `local` is present, yet the code runs every time

NisanthanNanthakumar: I thought about it and realized that this message should only run during `--auto` bc it's saying that we couldn't find any **repo-based integrations**, which we don't check for during `--local`. 

Also, I ran it 'as is' with `--auto` and `--local` and it only showed up for `--local`.

Issue #768, opened 2020-06-19
sentry-cli projects list: error: could not parse JSON response


md5:
We're running into an error when `"team": null` is returned in the list of projects. I'm not sure if projects should be allowed to exist with no team, but if so it looks like `struct Project` should have `Option<Team>` instead of `Team`. Otherwise, I guess this is something we need to fix on our server.

```console
$ sentry-cli projects list
error: could not parse JSON response
  caused by: invalid type: null, expected struct Team at line 1 column 37043
```

PR #773, opened 2020-06-23
fix: Handle teamless projects correctly

kamilogorek:
fixes https://github.com/getsentry/sentry-cli/issues/768

### DIFF
--- a/src/commands/projects.rs
+++ b/src/commands/projects.rs
@@ -19,7 +19,12 @@ pub fn execute<'a>(matches: &ArgMatches<'a>) -> Result<(), Error> {
     let api = Api::current();
     let org = config.get_org(matches)?;
     let mut projects = api.list_organization_projects(&org)?;
-    projects.sort_by_key(|p| (p.team.name.clone(), p.name.clone()));
+    projects.sort_by_key(|p| {
+        (
+            p.team.as_ref().map_or(String::new(), |t| t.name.clone()),
+            p.name.clone(),
+        )
+    });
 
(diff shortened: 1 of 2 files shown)


src/commands/projects.rs
```diff
@@ -19,7 +19,12 @@ pub fn execute<'a>(matches: &ArgMatches<'a>) -> Result<(), Error> {
     let api = Api::current();
     let org = config.get_org(matches)?;
     let mut projects = api.list_organization_projects(&org)?;
-    projects.sort_by_key(|p| (p.team.name.clone(), p.name.clone()));
+    projects.sort_by_key(|p| {
+        (
+            p.team.as_ref().map_or(String::new(), |t| t.name.clone()),
```
kamilogorek, line 24: It also works with `p.team.as_ref().map_or(None, |t| Some(t.name.clone()))`, but I couldn't find a way to force it being last in the sorted list (other than using something ridiculous like very high unicode character

Swatinem, line 24: I wonder why the old code uses `clone`. I think you should avoid allocating here, as this is potentially called multiple times. Maybe `p.team.as_ref().unwrap_or("")` should do the trick? Similarly below.

kamilogorek: We want to access `p.team.name`, so we cannot do `unwrap_or` in this case.

Swatinem: Oh, right! Is there a way to avoid the `clone` though? I don’t see any reason why `sort_by_key` would require owned data.

PR #774, opened 2020-06-24
feat: Add pipeline env variable option and include it in UA string

NisanthanNanthakumar:
## Objective
We want to see how successful our pipelines are.

Our pipelines use `sentry-cli` to send release and commit information. We want to see how widely these pipelines are being used. So, we need to have some way of being able to distinguish when a release is created through a pipeline vs just via `sentry-cli`.

Currently, all `sentry-cli` requests are sent with a `sentry-cli` user agent. 

In this PR, we are allowing users to configure an environment variable for their pipeline that is then sent in through the user agent.

Screen Shot 2020-06-24 at 1 27 37 AM

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -426,7 +426,7 @@ impl Api {
 
         let env = match self.config.get_pipeline_env() {
             Ok(env) => env,
-            Err(_) => String::new()
+            Err(_) => String::new(),
         };
 
         ApiRequest::create(handle, &method, &url, auth, env)
@@ -1413,7 +1413,7 @@ impl ApiRequest {
         method: &Method,
         url: &str,
         auth: Option<&Auth>,
-        pipeline_env: String
+        pipeline_env: String,
     ) -> ApiResult<Self> {
         debug!("request {} {}", method, url);
         debug!("pipeline: {}", pipeline_env);
@@ -1421,7 +1421,10 @@ impl ApiRequest {
         let mut headers = curl::easy::List::new();
         headers.append("Expect:").ok();
         headers
-            .append(&format!("User-Agent: sentry-cli/{} {}", VERSION, pipeline_env))
+            .append(&format!(
+                "User-Agent: sentry-cli/{} {}",
+                VERSION, pipeline_env
+            ))
             .ok();
 
         match method {
--- a/src/api.rs
+++ b/src/api.rs
@@ -424,10 +424,7 @@ impl Api {
         // This toggles gzipping, useful for uploading large files
         handle.transfer_encoding(self.config.allow_transfer_encoding())?;
 
-        let env = match self.config.get_pipeline_env() {
-            Ok(env) => env,
-            Err(_) => String::new(),
-        };
+        let env = self.config.get_pipeline_env();
 
         ApiRequest::create(handle, &method, &url, auth, env)
     }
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -305,6 +305,18 @@ impl Config {
...
+        Ok(env::var("SENTRY_PIPELINE")
+            .ok()
+            .or_else(|| {
+                self.ini
+                    .get_from(Some("defaults"), "pipeline")
+                    .map(str::to_owned)
+            })
+            .unwrap_or_default())
+    }
+
```
kamilogorek, line 319:
```suggestion
    /// Return the default pipeline env.
    pub fn get_pipeline_env(&self) -> Option<String> {
        env::var("SENTRY_PIPELINE").ok().or_else(|| {
            self.ini
                .get_from(Some("defaults"), "pipeline")
                .map(str::to_owned)
        })
    }

```

kamilogorek: We don't care about anything else than the value itself, as it's optional.

src/api.rs
```diff
@@ -1408,13 +1413,18 @@ impl ApiRequest {
...
         url: &str,
         auth: Option<&Auth>,
+        pipeline_env: String,
     ) -> ApiResult<Self> {
         debug!("request {} {}", method, url);
+        debug!("pipeline: {}", pipeline_env);
 
         let mut headers = curl::easy::List::new();
         headers.append("Expect:").ok();
         headers
```
kamilogorek, line 1423:
```rs
        match pipeline_env {
            Some(env) => {
                debug!("pipeline: {}", env);
                headers
                    .append(&format!("User-Agent: sentry-cli/{} {}", VERSION, env))
                    .ok();
            }
            None => {
                headers
                    .append(&format!("User-Agent: sentry-cli/{}", VERSION))
                    .ok();
            }
        }
```

kamilogorek: Otherwise you'll get empty `pipeline: ` logs and unnecessary whitespace in the header

src/api.rs
```diff
@@ -1408,13 +1413,18 @@ impl ApiRequest {
         method: &Method,
         url: &str,
         auth: Option<&Auth>,
+        pipeline_env: String,
     ) -> ApiResult<Self> {
         debug!("request {} {}", method, url);
+        debug!("pipeline: {}", pipeline_env);
```
kamilogorek, line 1419:
```suggestion
```

src/api.rs
```diff
@@ -1408,13 +1413,18 @@ impl ApiRequest {
         method: &Method,
         url: &str,
         auth: Option<&Auth>,
+        pipeline_env: String,
```
kamilogorek, line 1416:
```suggestion
        pipeline_env: Option<String>,
```

src/api.rs
kamilogorek: GitHub doesnt support suggestions on empty opines, so had to paste it manually. Should replace lines 1423:1428

src/api.rs
```diff
@@ -424,7 +424,12 @@ impl Api {
         // This toggles gzipping, useful for uploading large files
         handle.transfer_encoding(self.config.allow_transfer_encoding())?;
 
-        ApiRequest::create(handle, &method, &url, auth)
+        let env = match self.config.get_pipeline_env() {
+            Ok(env) => env,
+            Err(_) => String::new(),
+        };
+
+        ApiRequest::create(handle, &method, &url, auth, env)
```
kamilogorek, line 432:
```suggestion
        let env = self.config.get_pipeline_env();
```

kamilogorek:
Perfect.

PR #775, opened 2020-06-24
fix: Parse BitBucket Server integration repo url correctly

kamilogorek:
Based on https://github.com/getsentry/sentry/blob/87a67caeebbadc1b2f40ab11fd99fe113feb7042/src/sentry/integrations/bitbucket_server/repository.py#L43 and https://github.com/getsentry/sentry/blob/87a67caeebbadc1b2f40ab11fd99fe113feb7042/tests/sentry/integrations/bitbucket_server/test_repository.py#L96

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -139,6 +139,8 @@ impl VcsUrl {
             static ref HOST_WITH_PORT: Regex = Regex::new(r"(.*):\d+$").unwrap();
             static ref GCB_GIT_PATH_RE: Regex =
                 Regex::new(r"^p/.+/r/github_(.+?)_(.+?)(?:\.git)?$").unwrap();
+            static ref BITBUCKET_SERVER_PATH_RE: Regex =
+                Regex::new(r"projects/(.+)/repos/(.+)/browse").unwrap();
         }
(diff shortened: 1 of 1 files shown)


src/utils/vcs.rs
```diff
@@ -139,6 +139,8 @@ impl VcsUrl {
             static ref HOST_WITH_PORT: Regex = Regex::new(r"(.*):\d+$").unwrap();
             static ref GCB_GIT_PATH_RE: Regex =
                 Regex::new(r"^p/.+/r/github_(.+?)_(.+?)(?:\.git)?$").unwrap();
+            static ref BITBUCKET_SERVER_PATH_RE: Regex =
+                Regex::new(r"projects/(.+)/repos/(.+)/browse").unwrap();
```
jan-auer, line 143: Is `/browse` also part of the git URL / ssh URL?

kamilogorek: Yes, it's part of the URL we provide - https://github.com/getsentry/sentry/blob/87a67caeebbadc1b2f40ab11fd99fe113feb7042/src/sentry/integrations/bitbucket_server/repository.py#L72-L75

PR #778, opened 2020-06-29
test: releases command integration tests

kamilogorek:
Covered:
- info
- releases
- releases new
- releases finalize
- releases delete
- releases list
- releases info


Cargo.toml
```diff
@@ -40,22 +40,29 @@ plist = "0.5.3"
...
 username = "0.2.0"
 uuid = { version = "0.8.1", features = ["v4", "serde"] }
 walkdir = "2.3.1"
 which = "3.1.1"
-zip = "0.5.5"
+zip = "0.5.6"
+
+[dev-dependencies]
+assert_cmd = "1.0.1"
+insta = {version= "0.16.0", features= ["redactions"]}
```
Swatinem, line 62:
```suggestion
insta = { version = "0.16.0", features = ["redactions"] }
```

I wish rustfmt would also fmt the `toml` files

tests/commands/info.rs
```diff
@@ -0,0 +1,41 @@
+use assert_cmd::Command;
+use mockito::mock;
+use predicates::prelude::*;
+use predicates::str::contains;
+
+use crate::common;
+
+#[test]
+fn info_works_when_all_required_env_are_present() {
```
Swatinem, line 9: I think the convention is to prefix these with `test_`?

tests/commands/releases_finalize.rs
```diff
@@ -0,0 +1,80 @@
...
+        "PUT",
+        "/api/0/projects/wat-org/wat-project/releases/wat-release/",
+    )
+    .match_body(Matcher::AllOf(vec![
+        Matcher::PartialJsonString(r#"{"projects":["wat-project"]}"#.to_string()),
+        Matcher::Regex("dateReleased".to_string()),
+    ]))
+    .with_status(200)
+    .with_header("content-type", "application/json")
+    .with_body(r#"{"dateReleased":"2020-06-29T12:16:49.368667Z","newGroups":0,"commitCount":0,"url":null,"data":{},"lastDeploy":null,"deployCount":0,"dateCreated":"2020-06-29T11:36:59.612687Z","lastEvent":null,"version":"wat-release","firstEvent":null,"lastCommit":null,"shortVersion":"wat-release","authors":[],"owner":null,"versionInfo":{"buildHash":null,"version":{"raw":"wat-release"},"description":"wat-release","package":null},"ref":null,"projects":[{"name":"test","platform":"javascript","slug":"test","platforms":["javascript"],"newGroups":0,"id":1861017}]}"#)
```
Swatinem, line 19: you can put line breaks in raw strings. this is super hard to read.

tests/commands/releases_finalize.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    Command::cargo_bin("sentry-cli")
+        .unwrap()
+        .envs(common::get_base_env())
+        .args(vec![
+            "releases",
+            "finalize",
+            "wat-release",
+            "--started",
+            "1431648100",
```
Swatinem, line 73: from a UI perspective, why not accept ISO strings as well?

tests/commands/releases_list.rs
```diff
@@ -0,0 +1,76 @@
...
+        .with_body(r#"[{"dateReleased":"2020-03-19T10:11:35.128919Z","newGroups":1,"commitCount":0,"url":null,"data":{},"lastDeploy":{"name":null,"url":null,"environment":"x","dateStarted":null,"dateFinished":"2020-05-18T13:39:06.033442Z","id":"6447717"},"deployCount":1,"dateCreated":"2020-03-19T10:11:31.983994Z","lastEvent":null,"version":"vue-1","firstEvent":null,"lastCommit":null,"shortVersion":"vue-1","authors":[],"owner":null,"versionInfo":{"buildHash":null,"version":{"raw":"vue-1"},"description":"vue-1","package":null},"ref":null,"projects":[{"name":"test","platform":"javascript","slug":"test","platforms":["javascript"],"newGroups":1,"id":1861017}]},{"dateReleased":null,"newGroups":0,"commitCount":0,"url":null,"data":{},"lastDeploy":null,"deployCount":0,"dateCreated":"2020-03-16T16:16:12.655209Z","lastEvent":null,"version":"ok","firstEvent":null,"lastCommit":null,"shortVersion":"ok","authors":[],"owner":null,"versionInfo":{"buildHash":null,"version":{"raw":"ok"},"description":"ok","package":null},"ref":null,"projects":[{"name":"test","platform":"javascript","slug":"test","platforms":["javascript"],"newGroups":0,"id":1861017}]}]"#)
```
Swatinem, line 39: is this the same as above? maybe dedupe this into a `static` or `const`?

tests/commands/info.rs
kamilogorek: I didn't find anything about this in Rust Book, and I like how the output looks now:

image

tests/commands/releases_finalize.rs
kamilogorek: I'm aware of this, but this is a raw dump from the real request debug log. Some of them are quite large, and I didn't want it to obscure the rest of an actual test.

tests/commands/releases_finalize.rs
kamilogorek: We do accept RFC3339 and RFC2822 as well, but I wanted to also verify here that timestamps are translated currently into `2020-06-29T12:16:49.368667Z` format.

tests/commands/releases_list.rs
kamilogorek: Most of the responses are the same (not all), and I wrote them as consts at first (see initial commits), but I unified it to use separate strings for now to make it self-contained and easier to move to `insta` crate later on.
It's also easier to just use the cli to get the output and copy it here directly.

PR #783, opened 2020-07-06
fix: binaryPath resolution on macos

aleclarson:
The entry module for `@sentry/cli` is located at `./bin/sentry-cli`, 
yet the `binaryPath` variable resolves to `./sentry-cli` accidentally.


aleclarson:
Oh nvm, I see what went wrong. The install script was never executed. Sorry for the noise!

kamilogorek:
No problem, glad you found it.

Issue #789, opened 2020-07-20
Releases set-commits auto, with no linked repos.


samuelhwilliams:
We're running a self-hosted Sentry and I've been following the docs to set up release tracking. I've got to the part where we're  creating the release and associating commits. We aren't using the sentry integrations, and it seems like this causes `set-commits --auto` to short-circuit prematurely. [The docs suggest that it's fine to not set up integrations](https://docs.sentry.io/workflow/releases/#alternatively-without-a-repository-integration), but the code snippet below stops the command from proceeding. Am I missing something?

I checked out the repo locally, dropped the code block below, built it and ran it to see if it would work. I ended up hitting a validation error on the author email for one of the commits, but that seems to be a separate issue. After dropping out enough commits to drop the "invalid" email address, the release was logged correctly.

https://github.com/getsentry/sentry-cli/blob/aef73a9aef9a181ad6e130454f251a669dc10496/src/commands/releases.rs#L473-L478

NisanthanNanthakumar:
@samuelhwilliams Hi! So I think you are right! I just tested it out on a fresh install and if you don't have any repos connected/created the command will bail. Obviously this doesn't make sense if you want to only push local commit metadata.

Thank you for bringing this up!! I'll start working on a patch for this.

Also, I'm curious about the "invalid" author emails. Can you elaborate on this issue?

samuelhwilliams:
Hey @NisanthanNanthakumar - thanks for the reply! Sounds good - will keep an eye out for the patch and release. :)

I think the email issue I mentioned is not actually a problem. It seems to be a  configurable option [here](https://github.com/getsentry/sentry/blob/95767d455b8004ec4b4c5026d84b64b6348e6d37/src/sentry/security/utils.py#L38-L39), so might be that we've just configured it to be too restrictive and I'm checking with our team internally. FYI only, the email it's rejecting are dependabot's automated ones (e.g. 123456789+dependabot[bot]@users.noreply.github.com). But as I say - on further inspection I don't think this is a problem on Sentry's end. :)

kamilogorek:
@samuelhwilliams PR is pending and should be merged and released soon. In the meantime, however, would you be able to give it a try locally?

```terminal
# This step can be skipped if you already have rust locally
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

git clone https://github.com/getsentry/sentry-cli.git && cd $_
git checkout API-1165
cargo build
[...]

PR #791, opened 2020-07-22
fix: Default to sending local commits if no repos linked

NisanthanNanthakumar:
Refer to https://github.com/getsentry/sentry-cli/issues/789

## Objective
When an org has no integrations setup and runs `sentry-cli releases set-commits --auto`, it currently bails. The expected behaviour is to create a new repo (with Unknown Provider) in Sentry and upload the local git commit metadata.


kamilogorek:
@NisanthanNanthakumar update lint error, and feel free to merge it afterwards without approval

PR #793, opened 2020-07-24
ref: Make Update Nagger less aggressive

kamilogorek:
Remove unnecessary log and move nagger after we printed the original error message.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -261,17 +261,7 @@ pub fn execute(args: &[String]) -> Result<(), Error> {
 
 fn run() -> Result<(), Error> {
     prepare_environment();
-    match execute(&env::args().collect::<Vec<String>>()) {
-        Ok(()) => Ok(()),
-        Err(err) => {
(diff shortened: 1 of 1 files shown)


src/commands/mod.rs
```diff
@@ -314,7 +304,14 @@ pub fn main() {
                     crate::utils::crashreporting::try_report_to_sentry(&err);
                 }
                 1
-            }
+            };
+
+            // if the user hit an error, it might be time to run the update
+            // nagger because maybe they tried to do something only newer
+            // versions support.
+            run_sentrycli_update_nagger();
```
jan-auer, line 312: This now runs unconditionally, which I think makes sense in general. The comment no longer matches, though.

kamilogorek: It's not true. It'll only run if `run` returns the error, which is the exact same behavior as before, just moved to after original error printing.

jan-auer: Oh right, sorry, I read this wrong. All good!

PR #809, opened 2020-08-20
Bump macOS VM version on Azure Pipeline

kamilogorek:
`macos-10.13` is already deprecated and removed from Azure, ~and I don't see a reason why not to use latest instead.~

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ jobs:
   - job: sentry_cli_macos
     displayName: "Build: MacOS x86_64"
     pool:
-      vmImage: "macos-10.13"
+      vmImage: "macOS-10.14"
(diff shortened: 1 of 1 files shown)


tonyo (commented):
Two things here:
1. We haven't fully migrated from Travis to Azure for this repo (e.g., we still upload to Zeus from Travis), so Azure results are basically unused now.
2. There's a certain value in building the final binaries we distribute on older systems, because (I think?) there's a higher chance that they will work on the newer system versions, while the opposite is usually not true (or at least not with the same confidence level). I don't have specific numbers/version to back these claims, but at least it's worth checking that e.g. `sentry-cli` built on the latest windows version still works on an older one.

tonyo (approved):

PR #811, opened 2020-08-27
Added possibility to support other arch than x86/x64

fastlorenzo:
This PR adds the possibility to download other architecture builds than x86_64 or i686.

os.arch returns the following possible values: 'arm', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 's390', 's390x', 'x32', and 'x64'. [https://nodejs.org/api/os.html#os_os_arch](Source)

With the current implementation, it is not possible to have it work with any other architecture, even when changing the CDN_URL value via env. variables as it always defaults to i686.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -37,7 +37,7 @@ function shouldRenderProgressBar() {
 
 function getDownloadUrl(platform, arch) {
   const releasesUrl = `${CDN_URL}/${pkgInfo.version}/sentry-cli`;
-  const archString = arch == "x64" ? 'x86_64' : arch == "x86" ? 'i686' : arch;
+  const archString = arch === "x64" ? 'x86_64' : arch === "x86" ? 'i686' : arch;
   switch (platform) {
     case 'darwin':
       return `${releasesUrl}-Darwin-x86_64`;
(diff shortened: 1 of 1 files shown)


kamilogorek:
Thanks for the PR. Please fix the linter issues and it should be good to go

kamilogorek:
Thanks!

zyf0330:
os.arch won't return x86 but ia32. So ridiculous.

scripts/install.js
```diff
@@ -37,7 +37,17 @@ function shouldRenderProgressBar() {
 
 function getDownloadUrl(platform, arch) {
   const releasesUrl = `${CDN_URL}/${pkgInfo.version}/sentry-cli`;
-  const archString = arch.indexOf('64') > -1 ? 'x86_64' : 'i686';
+  let archString = '';
+  switch (arch) {
+    case 'x64':
+      archString = 'x86_64';
+      break;
+    case 'x86':
```
sqxieshuai, line 45: i install @sentry/cli failed with node@12.18.3 32bit

because sentry only provide sentry-cli-Windows-i686.exe 

but by your rule, install script will try to download file sentry-cli-Windows-ia32.exe ...

kamilogorek: @zyf0330 @sqxieshuai it's a one-line fix which I'd greatly appreciate :)

fastlorenzo: I guess adding `case 'ia32':` after should do the trick?

zyf0330: You are right, use 'ia32' as replacement of 'x86'.Could you issue a PR?

fastlorenzo: Done in #824

PR #812, opened 2020-08-28
feat: add skip confirmation arg to uninstall cmd

erictaylor:
This adds a `--skip-confirmation` argument to the `uninstall` command.

I use this tool on CI/Docker and uninstall after use so it doesn't add to an image. Uninstalling with the confirmation prompt is an issue, so having an argument like this ideal to skip the confirmation prompt all together and simply uninstall.

---

This is my first Rust code ever written, so please be gentle if I made a mistake. 😆

### DIFF
--- a/src/commands/uninstall.rs
+++ b/src/commands/uninstall.rs
@@ -2,7 +2,7 @@
 use std::env;
 use std::fs;
 
-use clap::{App, AppSettings, ArgMatches};
+use clap::{App, AppSettings, Arg, ArgMatches};
 use console::style;
 use failure::Error;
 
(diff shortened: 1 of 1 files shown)


kamilogorek:
The code looks good, thanks! :) Can we rename the flag to `--confirm` instead? This is what some of the CLI tools that I know already use.

kamilogorek:
You can use `make lint` and `make format` to fix linting issues. And once you rebase on top of master, build failure should be gone as well.

erictaylor:
@kamilogorek I've updated the option to `confirm` and rebased from `master` as requested. When I run `make lint` I get two errors that don't appear to be related to my changes. Let me know if there is anything else I need to correct/change. 😄

kamilogorek:
Looks good and passes all checks. Thanks! :)

erictaylor:
@kamilogorek just curious when I might see a release including this?

kamilogorek:
@erictaylor I'll cut the new release Today in few hours

kamilogorek:
Released as `1.56.0`. Cheers!

PR #830, opened 2020-09-18
feat: Expose environment configuration in javascript

dcramer:
Expose common options available via environment variables as configurable parameters to the CLI constructor.

```javascript

new SentryCli(null, {
  authToken: ...
});
```


dcramer:
The intent here is to make this less error-prone for instances like the sentry webpack plugin. We cannot (nor should we) ask users to setup some arbitrary configuration file or do extra work if we can avoid it. This means we can just have them define org, authToken, etc via their config file (+ env vars when appropriate).

dcramer:
https://github.com/getsentry/sentry-docs/pull/2305

rhcarvalho (approved):

Issue #734, opened 2020-05-14
Releasing to mulitple projects at a time


ndench:
The [releases documentation](https://docs.sentry.io/workflow/releases/#using-the-cli) shows that we can create a new release for multiple projects:

```sh
sentry-cli releases new -p project1 -p project2 $VERSION
```

However, specifying the `-p` option multiple times give the following error:

```
error: The argument '--project <PROJECT>' was provided more than once, but cannot be used multiple times

USAGE:
    sentry-cli releases <SUBCOMMAND>

For more information try --help
```

A workaround for now is to run the command multiple times for each project:

```sh
sentry-cli releases new -p project1 $VERSION
sentry-cli releases new -p project2 $VERSION
```

It would be great if the cli supported multiple projects for this operation. It looks like #52 had a crack at this but then closed the PR.

If this support is not going to be added, the docs should probably be updated to reflect that.

kamilogorek:
Fixed in https://github.com/getsentry/sentry-cli/pull/775

kamilogorek:
Right it's done that way because every other command is project-scoped. We also heavily rely on ini/properties files for this configuration, and as you can tell from `--help` flag, `-p/--project` is even hidden from there to not encourage it's usage.
We'll definitely rethink this behavior, but I'm afraid it won't be anytime soon, as it can be achieved with loops in any scripting language/shell.

PR #842, opened 2020-10-07
feat(js): allows the user to specify multiple projects for a release

scefali:
This PR enables setting multiple projects when creating a release with the JS wrapper for the CLI. It adds a new optional parameter to the `new` function called `options` which has a `projects` field which is an array of strings. Example:

```
cli.releases.new('my-version', { projects: ['proj-a', 'proj-b'] });
```

Becomes:
```
sentry-cli releases new -p proj-a -p proj-b my-version
```
CLI docs: https://docs.sentry.io/product/releases/#using-the-cli

I've also added the multiple project options for the `uploadSourceMaps` command as well. However, the underlying CLI cannot handle multiple options and will throw this error:
```
error: The argument '--project <PROJECT>' was provided more than once, but cannot be used multiple times
```
However, you can still provide a single option which is useful so you can upload source maps in a loop. Furthermore, if we ever do update the CLI to handle multiple options we won't need to update the JS wrapper since it can pass the arguments correctly.

Discussion on this limitation here: https://github.com/getsentry/sentry-cli/issues/734#issuecomment-652318315

Note I did not update `setCommits` because that doesn't need a project at all. After creating the release with `new`, you just need to pass in the release name and it will work correctly.

Here's the script I was using to test these changes:

```
async function main() {
  const cli = new SentryCli().releases;
  const version = "10.9.2";
  const prefix = "static/js";

  const projects = ["test-sentry-cli-js", "test-steve"];
  await cli.new(version, { projects });
  await cli.setCommits(version, { auto: true });

  await Promise.all(
    projects.map((project) => {
      const sourceMapOptions = {
        projects: [project],
        urlPrefix: `~/${prefix}`,
        validate: true,
        include: [`build/${prefix}`],
      };
      return cli.uploadSourceMaps(version, sourceMapOptions);
    })
  );
}
```

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -181,7 +181,7 @@ function execute(args, live, silent, configFile, config = {}) {
 }
 
 function getProjectFlagsFromOptions({ projects = [] } = {}) {
-  return projects.reduce((flags, project) => flags.concat('-p', project), [])
+  return projects.reduce((flags, project) => flags.concat('-p', project), []);
 }
 
 module.exports = {
--- a/js/helper.js
+++ b/js/helper.js
@@ -180,10 +180,22 @@ function execute(args, live, silent, configFile, config = {}) {
   });
 }
 
+function getProjectFlagsFromOptions(options) {
+  if (options && options.projects) {
+    return options.projects.reduce((projectFlags, project) => {
+      projectFlags.push('-p');
+      projectFlags.push(project);
+      return projectFlags;
+    }, []);
+  }
+  return [];
+}
+
 module.exports = {
   mockBinaryPath,
   serializeOptions,
   prepareCommand,
   getPath,
   execute,
+  getProjectFlagsFromOptions,
 };
--- a/js/helper.js
+++ b/js/helper.js
@@ -180,15 +180,8 @@ function execute(args, live, silent, configFile, config = {}) {
   });
 }
 
-function getProjectFlagsFromOptions(options) {
-  if (options && options.projects) {
-    return options.projects.reduce((projectFlags, project) => {
-      projectFlags.push('-p');
-      projectFlags.push(project);
-      return projectFlags;
-    }, []);
-  }
-  return [];
+function getProjectFlagsFromOptions({ projects = [] } = {}) {
+  return projects.reduce((flags, project) => flags.concat('-p', project), [])
 }
 
 module.exports = {
--- a/js/releases/__tests__/index.test.js
+++ b/js/releases/__tests__/index.test.js
@@ -3,9 +3,44 @@
 const SentryCli = require('../../');
 
 describe('SentryCli releases', () => {
+  afterEach(() => {
+    jest.resetModules();
+  });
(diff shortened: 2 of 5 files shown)


kamilogorek:
If we want to make it future-proof, can we rework it in a way, where constructor will hold `baseArgs`, which include top-level flags and then it's reused everywhere? Something like

```js
constructor(options) {
  this.baseArgs = ['release'].concat(getCommandFlagsFromOptions(options));
}

// create map of name/flag so we can extend it later
const commandFlags = {
  'projects': '-p'
}

function getCommandFlagsFromOptions(options) {
  const flags = [];
  Object.keys(commandFlags).forEach(key => {
    const option = options[key];
    if (!option) return;
    (Array.isArray(option) ? option : [option]).forEach(option => {
      flags.push(commandFlags[key]);
      flags.push(options[key]);
    });
  });
  return flags;
}

// Later in the command

new(release, options) {
  const args = this.baseArgs.concat(['new', release, helper.getProjectFlagsFromOptions(options)]);
  return this.execute(args, null);
}
```

(this `getCommandFlagsFromOptions` part is actually working code (i guess) so you can copy-paste it :P)

scefali:
@kamilogorek I think this approach would be good if we didn't have the problem with the source maps command not being able to take multiple projects. But since it can't, you'd need to make a new instance of releases when you upload source maps which makes it pretty clunky. IMO, better to supply the arguments for each function individually until the CLI can accommodate multiple projects with upload commits.

kamilogorek (changes requested):

kamilogorek (approved):

MrGWest:
How is `finalize` supposed to work? Our builds used to work, but now `--project` is required it seems, but `-p proj-a -p proj-b` is failing with `The argument '--project <PROJECT>' was provided more than once, but cannot be used multiple times`.

kamilogorek:
@MrGWest did it happened using `1.60.0` by any chance? If so, please see https://github.com/getsentry/sentry-cli/issues/868

scefali:
@MrGWest Some of the release actions don't actually support using multiple projects, unfortunately (source maps being one of them). This is a limitation in the CLI core rather than the JS wrapper.

guillaume86:
This is really confusing in the documentation (and convoluted to use), from https://docs.sentry.io/product/cli/releases/

> Release Management
> The sentry-cli tool can be used for release management on Sentry. It allows you to create, edit and delete releases as well as upload release artifacts for them. Note that **releases are global per organization**. If you want the releases in different projects to be treated as separate entities, make the version name unique across the organization. For example, if you have projectA and projectB that share version numbers, you can name the releases projectA-1.0 and projectB-1.0 respectively.
> 
> **Because releases work on projects you will need to specify the organization and project you are working with**. For more information about this refer to Working with Projects.

The two statements seem to contradict each other. My preferred solution would be to never specify a project when using the release commands (append project name to release name if needed) and let sentry auto link releases to projects with events data.

PR #852, opened 2020-10-30
feat(bash-hook): Add cli arg to override sentry-cli command

BYK:
When using `sentry-cli` through Docker, the bash hook does not work as:
1. The inferred path of `sentry-cli` becomes the path inside the container, and does not apply to the host
2. The files used for reporting live outside the container and they need to be mapped to the container when running it.

This patch chooses the more flexible `--cli` argument to solve both problems where one can do something like

```shell
docker run --rm getsentry/sentry-cli bash-hook --cli='docker run --rm -v "$_SENTRY_TRACEBACK_FILE:$SENTRY_TRACEBACK_FILE" -v "$_SENTRY_LOG_FILE:$_SENTRY_LOG_FILE" getsentry/sentry-cli
```

Now that's ugly as hell, yes and we may opt to have auto detection for running inside docker to handle this. The downside is then we'd not support any other container engine.

### DIFF
--- a/src/commands/bash_hook.rs
+++ b/src/commands/bash_hook.rs
@@ -36,6 +36,12 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                 .help("Do not send environment variables along"),
         )
         .arg(
+            Arg::with_name("cli")
+                .long("cli")
+                .value_name("CMD")
+                .help("Explicitly set/override the sentry-cli command"),
+        )
+        .arg(
             Arg::with_name("send_event")
                 .long("send-event")
                 .requires_all(&["traceback", "log"])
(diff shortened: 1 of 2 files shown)


src/commands/bash_hook.rs
```diff
@@ -186,11 +192,17 @@ pub fn execute<'a>(matches: &ArgMatches<'a>) -> Result<(), Error> {
...
+        if matches.is_present("cli") {
+            script = script.replace(
             "___SENTRY_CLI___",
-            &env::current_exe().unwrap().display().to_string(),
-        );
+            matches.value_of("cli").unwrap()
+            );
+        } else {
+            script = script.replace("__SENTRY_CLI__", &env::current_exe().unwrap().display().to_string());
+        }
```
BYK, line 204: Instead of this, I tried every possible variation of `matcher` and `unwrap_or_else` etc. but they gave me hell with the temporary return value of the `env::current_exe().unwrap().display().to_string()` call or some weird mismatch between `String` and `str` types.

Open to suggestions to make this more "rustic".

kamilogorek:
```rust
.replace(
  "__SENTRY_CLI__",
  matches
    .value_of("cli")
    .unwrap_or(&env::current_exe().unwrap().display().to_string()),
);
```

kamilogorek (changes requested):

src/commands/bash_hook.rs
BYK: @kamilogorek that runs the expression `&env::current_exe().unwrap().display().to_string()` for the fallback unconditionally, which is why I went with this path (after exhausting the path of closures)

kamilogorek: True, that'll be eagerly evaluated, and returning a ref from `unwrap_or_else` is tricky.

kamilogorek (approved):

PR #853, opened 2020-11-03
Remove --rewrite flag and make it a default

kamilogorek:


mitsuhiko (changes requested):
I would leave the flag as noop hidden option for backwards compat.

mitsuhiko (approved):

PR #862, opened 2020-11-23
(chunk-upload): add option to set URL when uploading chunks

### LINKED ISSUE #839: [feature-request] Add a configuration for upload-url
## Why

Our Sentry instance is accessible under two domains: one is private and only reachable from inside the company network, the other is accessible from internet for error reporting and some api calls.

When trying to upload sourcemaps using `sentry-cli`, there is a call to `/api/0/organizations/<org>/chunk-upload/` returning a config containing a field `url`.

`sentry-cli` then uses this url instead of the one specified in `SENTRY_URL` to send file chunks, leading to an error because this url is not reachable in our case because it's the internal url.

Setting `system.upload-url-prefix` server side could work but it'll break the upload from the other domain.

## Feature request

Add a configuration in `sentry-cli` to override the upload url, something like `defaults.upload_url`?

Another option is to force the use of `defaults.url` but I'm not sure it's a good idea.

NeilHanlon:
Fixes #839 - to bypass options received from GET request to /chunk-upload/

Add --upload-url parameter
Support for SENTRY_UPLOAD_URL environment variable
"hopefully" safe (this is my first time 'writing' rust...
Added test to check for output of upload url in sentry-cli info
command
I've tested on linux and macos (had someone cross compile for me) and it seems to be working fine for all tasks (uploading dsyms mostly), but I'm more trying to solve an issue here than actually understanding exactly what is going on with the tool.

That said, I am open to suggestions/code review for what I've added and changed.. As I said I am brand new to Rust (though not strictly new to programming..)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1100,6 +1100,16 @@ impl Api {
         I: IntoIterator<Item = &'data T>,
         T: AsRef<(Digest, &'data [u8])> + 'data,
     {
+        
+        // If config.upload_url is set, override whatever was set by asking for the options.
+        let cached_upload_url = self.config.get_upload_url();
+
+        let url = if cached_upload_url.is_some() {
+            Cow::Owned(cached_upload_url.unwrap())
+        } else {
+            Cow::Borrowed(url)
+        };
+
         // Curl stores a raw pointer to the stringified checksum internally. We first
         // transform all checksums to string and keep them in scope until the request
         // has completed. The original iterator is not needed anymore after this.
@@ -1118,7 +1128,7 @@ impl Api {
         }
 
         let request = self
-            .request(Method::Post, url)?
+            .request(Method::Post, &url)?
             .with_form_data(form)?
             .with_retry(
                 self.config.get_max_retry_count().unwrap(),
--- a/src/api.rs
+++ b/src/api.rs
@@ -1102,13 +1102,11 @@ impl Api {
     {
         
         // If config.upload_url is set, override whatever was set by asking for the options.
-        let cached_upload_url = self.config.get_upload_url();
 
-        let url = if cached_upload_url.is_some() {
-            Cow::Owned(cached_upload_url.unwrap())
-        } else {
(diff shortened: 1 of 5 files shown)


NeilHanlon:
@mitsuhiko - any possibility you or someone could review this PR for me?

NeilHanlon:
Hi folks,

I've reformatted this with rustfmt so it passes linting. Apologies for not doing that earlier. It seems now this is passing all integration tests.

Neil

Johennes:
@jan-auer any news on the review? We'd like to start using this feature without having to rely on our fork of sentry-cli.

chenyanfei-m:
expect

mbertheau:
Instead of adding this bunch of code and increasing complexity, I have the following two ideas:

- instead of adding a new parameter to set a certain upload url, add an argument that lets sentry-cli upload to the URL given with --url or any of the other ways: "ignore server provided upload url". Because: what's the use case to have different `--url` and `--upload-url`?
- revisit why the upload URL is configurable at all, and remove that completely from the sentry server code

NeilHanlon:
> Instead of adding this bunch of code and increasing complexity, I have the following two ideas:
> 
> * instead of adding a new parameter to set a certain upload url, add an argument that lets sentry-cli upload to the URL given with --url or any of the other ways: "ignore server provided upload url". Because: what's the use case to have different `--url` and `--upload-url`?
> * revisit why the upload URL is configurable at all, and remove that completely from the sentry server code

#839 Describes why this is necessary, IMO. I'm not sure if things have changed in the past year or so since I've opened this PR. Frankly I haven't had any time to look at it, but if my dayjob team needs me to look at it again, I can.

I've heard no feedback from the getsentry team on this PR w.r.t. feedback on if they care about the implementation, and this is working for at least my dayjob. Agreed there may be a better less-complex solution to be implemented.

kamilogorek:
Will be fixed sentry-side here: https://github.com/getsentry/sentry/pull/29347

PR #872, opened 2020-12-09
fix(bash-hook): Add missing underscores for template

benregn:
Fixes regression introduced in https://github.com/getsentry/sentry-cli/pull/852, where the path of `sentry-cli` would be interpolated as `_/usr/local/bin/sentry-cli_`.


kamilogorek:
🤦‍♂️ how did I miss that. Thanks!

benregn:
You're welcome! 🙂

PR #876, opened 2020-12-16
fix(craft): Bump version for the release action

jan-auer:
Drops support for releasing sentry-cli locally and replaces sed for
compatibility with Linux.

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -eux
 
+if [ "$(uname -s)" != "Linux" ]; then
+    echo "sentry-cli can only be released on Linux!"
+    echo "Please use the GitHub Action instead."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/bump-version.sh
```diff
@@ -10,7 +16,7 @@ TARGET="${2}"
 echo "Current version: $VERSION"
 echo "Bumping version: $TARGET"
 
-sed -i '' -e "1,/^version/ s/^version.*/version = \"$TARGET\"/" Cargo.toml
+perl -pi -e "s/^version = .*\$/version = \"$TARGET\"/" Cargo.toml
```
jan-auer, line 19: Note that this will now replace all occurrences and not just the first. Since this matches the start of the line, however, this should be fine.

BYK: PERL!

invincible

BYK (approved):

Issue #676, opened 2020-03-06
Provide binary for ARM


kamilogorek:
ref: https://github.com/getsentry/sentry-cli/issues/592

Alexandero89:
Would really appreciate a binary for ARM.

Raspberry 4 with 4 Gig RAM and SSD should work fine with Sentry.

fastlorenzo:
Same here, preferably an ARM binary for musl as well (alpine docker)

omaraltayyan:
please note this has increased in importance with the release of AWS's new Gravaton2 processors into EC2 mainstream instances (T4G), so it's becoming a necessity to support ARM in popular packages.

albertshaw:
Any luck? sentry-webpack-plugin does not work with Macbook M1

omaraltayyan:
> Any luck? sentry-webpack-plugin does not work with Macbook M1

You can build sentry cli on arm from source, there's still no provided binary for it.

gersakbogdan:
Until this will be available you can use my repo: https://github.com/gersakbogdan/sentry-cli-Linux-arm64
Those are built from source.

zackify:
would be great to have this so that I can use sentry inside docker on M1

mercs600:
the same issue here - with installation inside of docker

Mati365:
bump

kamilogorek:
Released as `1.63.0`. Cheers!

tmakin:
@kamilogorek I've just tried the new 1.63 release on an EC2 ARM instance but the binary still appears to be missing. Here is my output from `yarn --verbose`
```
Error: Unable to download sentry-cli binary from https://downloads.sentry-cdn.com/sentry-cli/1.63.0/sentry-cli-Linux-arm64.
Server returned 403: Forbidden.
```

It's looking for `arm64` which seems to conflict with the `aarch64` naming convention [here](https://github.com/getsentry/sentry-cli/blob/master/.craft.yml#L41).

kamilogorek:
Released as `1.63.1`

PR #879, opened 2020-12-17
fix(brew): Prefer universal binaries in homebrew

jan-auer:
#878 missed the homebrew config, which was still installing single-arch binaries.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -32,13 +32,8 @@ targets:
         license "BSD-3-Clause"
 
         if OS.mac?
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):

kamilogorek:
I'd be nice to get this out so we can resolve https://github.com/getsentry/sentry-cli/issues/676 as wel

jan-auer:
This doesn't resolve #676, since we only build macOS ARM but this issue is looking for Linux. For that, we'd have to add another arch build.

PR #890, opened 2021-01-12
build: ARM for Linux

jan-auer:
TODO:

- [x] Set up ARM builds for https://github.com/getsentry/rust-musl-cross
- [x] Build aarch64
- [x] Build armv7
- [x] Update homebrew
- [x] Check updater target
- [x] Update install script in getsentry

We need to decide whether this should be called `aarch64`, which seems to be the output of `uname -m` by 64-bit applications, or `arm64`, which would also match macOS.

get-cli: https://github.com/getsentry/sentry-cli/pull/890

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -34,6 +34,12 @@ targets:
         if OS.mac?
           url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Darwin-universal"
           sha256 "{{checksums.sentry-cli-Darwin-universal}}"
+        elsif Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
+          url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Linux-aarch64"
+          sha256 "{{checksums.sentry-cli-Linux-aarch64}}"
+        elsif Hardware::CPU.arm?
+          url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Linux-armv7"
+          sha256 "{{checksums.sentry-cli-Linux-armv7}}"
         elsif Hardware::CPU.is_64_bit?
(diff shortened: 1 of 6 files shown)


.craft.yml
```diff
@@ -34,6 +34,12 @@ targets:
         if OS.mac?
           url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Darwin-universal"
           sha256 "{{checksums.sentry-cli-Darwin-universal}}"
+        elsif Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
+          url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Linux-aarch64"
+          sha256 "{{checksums.sentry-cli-Linux-aarch64}}"
+        elsif Hardware::CPU.arm?
+          url "https://downloads.sentry-cdn.com/sentry-cli/{{version}}/sentry-cli-Linux-armv7"
+          sha256 "{{checksums.sentry-cli-Linux-armv7}}"
```
jan-auer, line 42: How would you feel if we start nesting this to show the structure more clearly? 

- if mac? -> universal
- elsif linux?
  - if arm?
    - if 64bit -> aarch64
    - else -> armv7
  - elsif intel?
    - if 64bit? -> linux x64
    - else -> linux x86
  - else -> unsupported arch
- else -> unsupported OS

scripts/build-in-docker.sh
```diff
@@ -4,11 +4,14 @@ set -eux
 DOCKER_IMAGE="getsentry/rust-musl-cross:${DOCKER_TAG}"
 BUILD_DIR="/work"
 
+# TODO: Remove OPENSSL_NO_VENDOR once openssl-src >111.11.0 is released.
```
jan-auer, line 7: openssl-src 111.14.0 has been released. If it still doesn't work, let's update or remove the comment.

We control the openssl version in our fork of the rust-musl-cross container. Once we're able to vendor it on all architectures, we can get rid of that change in the fork.

PR #896, opened 2021-01-29
fix(chunk_upload): Limit chunk upload waiting to 5 minutes

jan-auer:
Under some conditions, chunk upload may not respond within a given time frame or
even time out completely. This leads sentry-cli to wait forever on a response
from Sentry.

This PR introduces an upper bound of 5 minutes. In chunk upload options, the
Sentry server can override this limit. It is not configurable on the CLI.

Fixes https://github.com/getsentry/sentry-cli/issues/891 in combination with https://github.com/getsentry/sentry/pull/23455

### DIFF
--- a/src/utils/dif_upload.rs
+++ b/src/utils/dif_upload.rs
@@ -14,6 +14,7 @@ use std::process::Command;
 use std::slice::{Chunks, Iter};
 use std::str;
 use std::thread;
+use std::time::{Duration, Instant};
 
 use console::style;
 use failure::{bail, err_msg, Error, SyncFailure};
@@ -32,7 +33,7 @@ use crate::api::{
     Api, ChunkUploadCapability, ChunkUploadOptions, ChunkedDifRequest, ChunkedFileState,
 };
 use crate::config::Config;
-use crate::constants::DEFAULT_MAX_DIF_SIZE;
+use crate::constants::{DEFAULT_MAX_DIF_SIZE, DEFAULT_MAX_WAIT};
 use crate::utils::chunks::{
     upload_chunks, BatchedSliceExt, Chunk, ItemSize, ASSEMBLE_POLL_INTERVAL,
 };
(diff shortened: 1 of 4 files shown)


src/utils/file_upload.rs
```diff
@@ -252,6 +260,10 @@ fn upload_files_chunked(
             break response;
         }
 
+        if context.wait && assemble_start.elapsed() > max_wait {
```
kamilogorek, line 263:
```suggestion
        if assemble_start.elapsed() > max_wait {
```

`context.wait` will always evaluate to `true` here, as L259 would short-circuit it already if it'd be `false`

src/utils/file_upload.rs
```diff
@@ -263,7 +275,11 @@ fn upload_files_chunked(
     progress.finish_and_clear();
 
     if response.state.is_pending() {
-        println!("{} File upload complete", style(">").dim());
+        if context.wait {
+            bail!("Failed to process files in {}s", max_wait.as_secs());
+        } else {
+            println!("{} File upload complete", style(">").dim());
```
kamilogorek, line 281: Maybe we could say something like `upload complete (processing pending)` or something of that sort? As this is what is actually happening?

kamilogorek (changes requested):

src/utils/dif_upload.rs
```diff
@@ -1130,9 +1137,9 @@ fn poll_dif_assemble(
         println!("{} File processing complete:\n", style(">").dim());
     }
 
-    let (mut successes, errors): (Vec<_>, _) = response
+    let (errors, mut successes): (Vec<_>, _) = response
         .into_iter()
-        .partition(|&(_, ref r)| !r.state.is_err());
+        .partition(|&(_, ref r)| r.state.is_err() || options.wait && r.state.is_pending());
```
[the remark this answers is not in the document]
jan-auer: Yeah. The intent is to put pending ones under errors, then print the "The file is still processing and not ready yet" message and fail the command.

src/utils/file_upload.rs
jan-auer: How about:

```suggestion
            println!("{} File upload complete (processing pending on server)", style(">").dim());
```

kamilogorek: Sounds good to me!

kamilogorek (approved):

grzegorzkrukowski:
Hey - do you think this PR can also fix this problem ?

https://github.com/getsentry/sentry-cli/issues/1004

Issue #898, opened 2021-02-01
Specious warning to upgrade to 1.61.0 when already using 1.62.0


jeffrey-aguilera:
### Environment

How do you use Sentry?
SaaS

Which SDK and version?
sentry-cli 1.62.0

### Steps to Reproduce

1. Run any sentry-cli command other than sentry-cli --version

### Expected Result

1. No version warning, as 1.62.0 is the latest version

### Actual Result

```
❯ sentry-cli releases -o my-organization list
...
sentry-cli update to 1.61.0 is available!
run brew upgrade sentry-cli to update

❯ sentry-cli --version
sentry-cli 1.62.0
```

kamilogorek:
Can you provide logs using `SENTRY_LOG_LEVEL=debug`? The update logic is based on the binary stored in the cache directory, which seems to be out-of-sync.

kamilogorek:
@jeffrey-aguilera your error should be gone now :)

brfrth:
Still getting this one 1.63.

![image](https://user-images.githubusercontent.com/6641565/110481249-a6d96d80-80e7-11eb-844b-a296d5ec81f6.png)

ThibaultVlacich:
Same, getting the issue with 1.63.1

```
$ sentry-cli info
Sentry Server: https://sentry.io
Default Organization: -
Default Project: -

sentry-cli update to 1.63.1 is available!
run sentry-cli update to update
$ sentry-cli --version
sentry-cli 1.63.1
```

kamilogorek:
Whoops, forgot to cut a release after the patch 😅 Released as `1.63.2`.

PR #899, opened 2021-02-03
ref: is_outdated should not report when release contain older version

kamilogorek:
This can happen only in the case of a broken registry, but we shouldn't report false-positives to the end-user.
https://sentry.io/get-cli/ is also using registry, so someone would let us know that it's not working correctly anyway.

Fixes https://github.com/getsentry/sentry-cli/issues/898


jan-auer (approved):

kamilogorek:
https://github.com/getsentry/sentry-cli/issues/898#issuecomment-812413050

PR #916, opened 2021-02-26
Verify CLI installation before launching

fmauquie:
It looks like `spawn` does not fail if the command does not exist, causing the script to return a 0 code although it didn't actually run.

We have had problems with the CLI not downloading on `npm install`, that caused the built React Native app to not contain the JS bundle.

Code uses synchronous/blocking flavor of `fs.exists()`: because we run in a separate process we can allow to block. I chose the exit code at random (as long as it is non-zero).


kamilogorek:
Thanks!

Issue #913, opened 2021-02-25
npm install Fails in AWS Graviton2


omaraltayyan:
there's an issue when installing the latest sentry cli on ARM with a linux OS, the npm installation generates a malformed URL for downloading the sentry cli binary causing a 403 error

### Environment

Sentry Cli: 1.63.0

### Steps to Reproduce

1. install sentry_cli on aws graviton 2 using npm

### Expected Result

installs successfully using this URL for downloading the CLI

```
https://downloads.sentry-cdn.com/sentry-cli/1.63.0/sentry-cli-Linux-aarch64
```

### Actual Result

the npm installation uses a malformed URL, due to the architecture build not having the same name as the CLI's binary.

```
https://downloads.sentry-cdn.com/sentry-cli/1.63.0/sentry-cli-Linux-arm64
```

kamilogorek:
Should be fixed with https://github.com/getsentry/sentry-cli/commit/11c0a6559bfa41bfa95d838d15b7701d2f3d6409

PR #917, opened 2021-03-02
rename arm64 as aarch64

jkatagi:
- When install v1.64.0 inside arm64 Linux Docker, we got 403 error.
```
Error: Unable to download sentry-cli binary from https://downloads.sentry-cdn.com/sentry-cli/1.63.0/sentry-cli-Linux-arm64.
Server returned 403: Forbidden.
```
- In #890 , arm64 renamed as aarch64, so we should update install/scripts as same.

## Ref
- https://github.com/getsentry/sentry-cli/issues/913

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -46,6 +46,9 @@ function getDownloadUrl(platform, arch) {
     case 'ia32':
       archString = 'i686';
       break;
+    case 'arm64':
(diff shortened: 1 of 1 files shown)


kamilogorek:
Ugh, sorry about that and thanks for the fix! Will release patch Today

PR #927, opened 2021-03-23
fix: Update proguard version

JaCzekanski:
Fixes has_line_info for certain mappings (warning: proguard mapping was ignored because it does not contain any line information.)

[More info](https://github.com/getsentry/rust-proguard/pull/21)


kamilogorek:
Thanks!

PR #928, opened 2021-03-23
feat: install git in Dockerfile

remcohaszing:
The following command depends on `git` being available:

```sh
sentry-cli releases set-commits --auto 1.2.3
```

This increases the Docker image size from 23.6MB to 39.6MB. IMO this is a relatively big, but acceptable increase.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN cp target/$BUILD_TARGET/release/sentry-cli /usr/local/bin/sentry-cli
 
 FROM alpine:3.12
 WORKDIR /work
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache ca-certificates git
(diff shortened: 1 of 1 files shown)


kamilogorek:
I'm fine adding this, but I'd love to have a confirmation from @jan-auer or @mitsuhiko

jan-auer:
What precisely do we need `git` as a dependency for? sentry-cli uses [git2](https://docs.rs/git2/0.13.17/git2/) which is based on the statically linked [libgit2](https://libgit2.org/).

@remcohaszing can you share the error message you are getting from sentry-cli without this dependency installed with `--log-level=debug` enabled?

remcohaszing:
Never mind. Apparently this is no longer necessary.

https://gitlab.com/appsemble/appsemble/-/jobs/1226653216

Issue #560, opened 2019-06-27
TypeScript Type Definitions


sameoldmadness:
Hi!

It would be handy to have type annotations for this module.

Do you have any plans to rewrite it in TypeScript or provide `d.ts` file?

I have written some types for myself, and could either prepare PR for this repo or upload it to DefinitelyTyped.

Which approach would be better?

In case someone needs the types:

```ts
declare module '@sentry/cli' {
  interface UploadSourceMapsOptions {
    include: string[]
    ignore?: string[]
    ignoreFile?: string
    rewrite?: boolean
    sourceMapReference?: boolean
    stripPrefix?: string[]
    stripCommonPrefix?: boolean
    validate?: boolean
    urlPrefix?: string
    urlSuffix?: string
    ext?: string[]
  }

  export default class SentryCli {
    constructor(configFile?: string, options?: Record<string, any>)
    releases: {
      new (release: string): Promise<string | undefined>
      finalize(release: string): Promise<string | undefined>
      proposeVersion(): Promise<string>
      uploadSourceMaps(
        release: string,
        options: UploadSourceMapsOptions
      ): Promise<Array<string | undefined>>
    }
  }
}
```

kamilogorek:
Sorry for such a long response, it'd be great if someone could provide those types. We can host them here, direclty in the repo :)

vdh:
@sameoldmadness I extended this out with the comments and extra comments I saw in `index.js` and `releases/index.js` in https://github.com/getsentry/sentry-cli/tree/master/js

https://gist.github.com/vdh/905e063b03c58166c21c25d1dbb9d663

I could keep going and type it out further, but it feels like this would be a lot more streamlined if the existing files were converted to Typescript because I was doing a lot of copy/pasting of the existing doc blocks.

redareda9:
Here we are, in 2021, and no type definitions for sentry-cli.

kamilogorek:
Right? So many people using it, yet so few wanting to help.

redareda9:
Yeah it's not like some people were getting paid to improve it

ssthouse:
I found there are some 3rd party sentry plugins(which uses `sentry-cli` underneath) that already write TS types for this repo.
It shouldn't take too much effort to add TS types, I'll try to move the type definitions to this repo and submit a PR~

> Right? So many people using it, yet so few wanting to help.

kamilogorek:
[...]

PR #934, opened 2021-04-02
feat: add basic SentryCli type support

ssthouse:
hi，I add the basic SentryCli type definitions and test locally with `npm link`.

the result looks fine.

image

image


ssthouse:
#560

ssthouse:
@kamilogorek hi, please check when you have time : D

js/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+declare module '@sentry/cli' {
+  export interface SentryCliOptions {
+    url?: string;
+    authToken?: string;
+    apiKey?: string;
+    dsn?: string;
+    org?: string;
+    project?: string;
+    vscRemote?: string;
+    silent: boolean;
```
kamilogorek, line 13: `silent` can be marked optional, as we only change the behavior if it's `true`. Any falsy value will work for disabling it.
```suggestion
    silent?: boolean;
```

js/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+  export default class SentryCli {
+    constructor(configFile?: string | null, options?: SentryCliOptions)
+
+    public configFile?: string;
+    public options?: SentryCliOptions;
+    public releases: SentryCliReleases
+
+    public static getVersion(): string
+    public static getPath(): string
+    public static execute(args: string[], live: boolean): Promise<string>
```
kamilogorek, line 85: `execute` is not static in this case:

```suggestion
    public execute(args: string[], live: boolean): Promise<string>
```

js/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  export interface SentryCliCommitsOptions {
+    auto?: boolean;
+    repo?: string;
+    commit?: string;
+    previousCommit?: string;
+  }
+
+  export interface SentryCliReleases {
+    ['new'](
```
kamilogorek, line 47: is `new` reserved keyword here and we cannot use `new(release: string...`? 😶

kamilogorek (changes requested):
Just a few details. Thanks, really appreciate it :)

js/index.d.ts
ssthouse: yes~

kamilogorek (approved):

kamilogorek:
Thanks! Will try to release it this week.

ckknight:
@kamilogorek I'm noticing that the `.d.ts` file isn't being included in the package published on npm (v1.63.2)

kamilogorek:
@ckknight it's because this commit has not been release yet :) I'll cut off a new release someday this week.

kamilogorek:
@ckknight release has been cut off. Should be deployed any hour tomorrow as `1.64.0`. Cheers!

PR #952, opened 2021-04-28
feat(upload_dif): Support BCSymbolMap uploading

flub:
Adds support for uploading BCSymbolMaps and their associated PList
files.

----

Related PRs:
https://github.com/getsentry/sentry/pull/25712
https://github.com/getsentry/symbolic/pull/373
https://github.com/getsentry/symbolicator/pull/414

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,8 @@ serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
 sha1 = { version = "0.6.0", features = ["serde"] }
 sourcemap = { version = "5.0.0", features = ["ram_bundle"] }
-symbolic = { version = "8.0.5", features = ["debuginfo-serde"] }
+# symbolic = { version = "8.0.5", features = ["debuginfo-serde"] }
+symbolic = { git = "https://github.com/getsentry/symbolic", branch="master", features = ["debuginfo-serde"]}
 url = "2.1.1"
 username = "0.2.0"
 uuid = { version = "0.8.1", features = ["v4", "serde"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,8 +50,7 @@ serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
 sha1 = { version = "0.6.0", features = ["serde"] }
 sourcemap = { version = "5.0.0", features = ["ram_bundle"] }
-# symbolic = { version = "8.0.5", features = ["debuginfo-serde"] }
-symbolic = { git = "https://github.com/getsentry/symbolic", branch="master", features = ["debuginfo-serde"]}
+symbolic = { version = "8.1.0", features = ["debuginfo-serde"] }
 url = "2.1.1"
 username = "0.2.0"
 uuid = { version = "0.8.1", features = ["v4", "serde"] }
--- a/src/utils/dif.rs
+++ b/src/utils/dif.rs
@@ -62,10 +62,8 @@ impl str::FromStr for DifType {
 }
 
 /// Declares which features an object may have to be uploaded.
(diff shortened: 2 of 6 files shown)


src/utils/dif_upload.rs
```diff
@@ -1680,21 +1881,82 @@ impl DifUpload {
...
+            return false;
+        }
+
+        // Skip if this DIF does not have features we want.
+        if !self.valid_features(&dif) {
+            debug!("skipping {} because of features", dif.name);
+            return false;
+        }
+
+        // Skip if this DIF if it has no DebugId or we are only looking for certain IDs.
```
Swatinem, line 1947:
```suggestion
        // Skip if this DIF has no DebugId or we are only looking for certain IDs.
```

Swatinem (commented):
lgtm

flub:
This PR currently treats `.bcsymbolmap` and `.plist` files as unrelated DIFs.  An alternative is to use the "attachments" system of uploaded difs to treat the `.plist` as a related file to the `.bcsymbolmap`, which would appear to have some logic to it.  The downside of doing that is that is becomes harder to upload just a `.plist` on its own.  It is not unreasonable to normally only ever upload `.bcsymbolmap` files as the `.plist` could be fetched directly by sentry using appstore connect.

But is something fails in a weird way it is possible for a user to try and fetch the `dSYM`s from appstore connect themselves and then upload the result with sentry-cli.  If they did the download with their browser rather than XCode they would end up with the `.plist` file but not with the `.bcsymbolmap` file, and thus would not be able to upload the `.plist` files.

This is probably a strange edge case, but it seems more failure-resilient by not requiring to have the `.bcsymbolmap` to be able to upload the `.plist` file.

But please disagree, that's why I wrote this 😄

Cargo.toml
```diff
@@ -50,7 +50,8 @@ serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
 sha1 = { version = "0.6.0", features = ["serde"] }
 sourcemap = { version = "5.0.0", features = ["ram_bundle"] }
-symbolic = { version = "8.0.5", features = ["debuginfo-serde"] }
+# symbolic = { version = "8.0.5", features = ["debuginfo-serde"] }
+symbolic = { git = "https://github.com/getsentry/symbolic", branch="master", features = ["debuginfo-serde"]}
```
flub, line 54: TODO: use a released symbolic

src/utils/dif.rs
```diff
@@ -62,6 +62,8 @@ impl str::FromStr for DifType {
 }
 
 /// Declares which features an object may have to be uploaded.
+// TODO(flub): Consider renaming this to ObjectDifFeatures, since non-object DIFs do not
+// have features.  The doc comment is correct here.
```
flub, line 66: Please provide opinions in favour or against this renaming.

Swatinem (approved):

Swatinem:
Related to your question about related DIFs: I think its better to keep them unrelated, right? As that would make it easier to separately upload them?

kamilogorek:
Feel free to merge it any time you think it's ready.

src/utils/dif_upload.rs
```diff
@@ -1345,7 +1345,7 @@ fn poll_dif_assemble(
                     k => format!(" {:#}", k),
                 },
                 ParsedDif::BcSymbolMap(_) => String::from("bcsymbolmap"),
-                ParsedDif::PList(_) => String::from("plist"),
+                ParsedDif::UuidMap(_) => String::from("plist"),
```
Swatinem, line 1348: I think this only prints on the console, right? not sure if we should also change that string

flub: 😱  yes, you're entirely right, updated.

PR #961, opened 2021-05-16
Feat/dotnet tool

bruno-garcia:


dotnet/Sentry.Cli.MSBuild/SentryCliDownloader.cs
```diff
@@ -0,0 +1,94 @@
...
+        public SentryCliDownloader(HttpMessageHandler? handler)
+            => _handler = handler;
+
+        public async Task DownloadForCurrentDevice(CancellationToken token, string? version = null)
+        {
+            version ??= await FindLatestRelease(token).ConfigureAwait(false);
+            var executableName = GetSentryCliName();
+            var url = GetUrl(version, executableName);
+            var client = GetHttpClient();
+            var cli = client.GetStreamAsync(url).GetAwaiter().GetResult();
```
Tyrrrz, line 30:
```suggestion
            var cli = await client.GetStreamAsync(url);
```

kamilogorek:
What is this black magic?

bruno-garcia:
Part of my Saturday. I was experimenting with having `dotnet tool install sentry-cli` but I'm not sure I'll go through with it.

The MSBuild stuff I think I'll keep but move to `sentry-dotnet`. It can be used by .NET on mobile platforms to upload symbols for Android and iOS. It can be used then by Xamarin, Unity and the new .NET 6 coming later this year.

bruno-garcia:
There's an MSBuild task `DownloadFile` that can be used to download a compiled `sentry-cli` to embed in the `Sentry` nuget package and ship it all there so this PR will be used on that instead of become a wrapper dotnet tool

Issue #956, opened 2021-05-03
Command should succeed in case previous release commit was not found


rreckonerr:
### Environment

I'm using an ember deployment addon that uses JS @senty/cli v1.41.0 under the hood and I've noticed that [if the previous release commit wasn't found](https://github.com/Exelord/ember-cli-deploy-sentry-cli/issues/10) the build pipeline fails with
```zsh
Error: Command failed: node_modules/.bin/sentry-cli  --auth-token XXX releases --org XXX --project XXX set-commits --auto XXX@XXX+XXX
error: Could not find the SHA of the previous release in the git history. Increase your git clone depth.

Add --log-level=[info|debug] or export SENTRY_LOG_LEVEL=[info|debug] to see more output.
Please attach the full debug log to all bug reports.
```

Looking through your repo I've noticed that other users are facing the same issues but there's no viable solution to the problem.

- Archiving releases that contain non-existing commits is not an option, since in some cases releases are gradually delivered to the users.
- `"my-repo@from..to"` makes no sense if the release commit is gone for good.

### Steps to Reproduce

1. Add a fixup commit
2. Deploy
3. Squash
4. Deploy

### Expected Result

I expect the command to succeed even in the case of missing the previous commit.

### Actual Result

An error is thrown

### Linked issues
- https://github.com/getsentry/sentry-cli/issues/792
- https://github.com/getsentry/sentry-cli/issues/945
- https://github.com/Exelord/ember-cli-deploy-sentry-cli/issues/10

drjayvee:
Our CD is also broken by this issue I believe. It would be nice if there were some way to "restart" the change list, or of course `--ignore-previous` or similar.

drjayvee:
To add insult to injury, I can't seem to manually set of clear commits for past releases.

For example:
```
$ sentry-cli releases set-commits --clear 493899324
Clearing commits for release.
```
And yet when I Ctrl-F5 that release's web page, it's still showing releases.

Maybe I should create a separate issue for this, but then maybe it's not working because of this issue.

kamilogorek:
@drjayvee thanks for letting us know regarding `--clear` issue. Someone will take care of it, and I'll let you know once it's fixed.

rreckonerr:
Hi @kamilogorek!
And what about the issue itself?

rreckonerr:
Thanks a lot!

drummerjolev:
[...]

Issue #945, opened 2021-04-14
sentry release failed on GitHub workflow


MahaZaid:
### Environment

Which SDK and version?
.dotnet core 3

### Steps to Reproduce
I am trying to publish a release through GitHub workflow

```
- uses: actions/checkout@v2
      with:
        fetch-depth: 0

    - name: Create Sentry release
      uses: getsentry/action-release@v1
      env:
        SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
        SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
        SENTRY_PROJECT: ${{ env.SENTRY_PROJECT }}
      with:
        environment: ${{ env.EKS_CLUSTER_NAME }}
        version: ${{ env.VERSION }}
```
### Expected Result

publish the release

### Actual Result

<img width="1075" alt="Screen Shot 2021-04-14 at 2 27 37 PM" src="https://user-images.githubusercontent.com/20710308/114710038-91321600-9d2d-11eb-86b9-4e20089a03e9.png">

I am already adding fetch before running the plugin

Is there a way to publish the release without the GitHub plugin ?

kamilogorek:
If `fetch-depth: 0` is set, as you already mentioned, it means that the last release containing associated commits (you can find it in the Releases section of your Sentry account) is using commit hash which has been somehow removed from your git repository. Most likely by either amending or rebasing. And because of that, we cannot find it in the repo and create a valid diff.

The easiest way to fix this is to archive releases that contain non-existing commits.

iwakoscott:
I am not sure if this is similar but, I am seeing `error: release not found` when I run - `sentry-cli releases set-commits --auto $VERSION`. Hitting `https://sentry.io/api/0/organizations/<org_name>/releases/<commit_sha>/` gives me a 404.

kamilogorek:
@iwakoscott it should work not, we had a small hiccup on our end.

marandaneto:
@kamilogorek it should work `now` you meant I believe

kamilogorek:
Details, details. Thanks for the correction :D

iwakoscott:
Thank you! 🙏

kamilogorek:
Released `--ignore-missing` flag in `1.65.0`

Issue #792, opened 2020-07-23
error: Couldn’t not find the SHA of the previous release in the git history.


sgametrio:
Hey everybody,
I'm using Sentry on a repository where we have several branches that upload source maps to Sentry, each time with a new release. 

When calling these CLI commands:
```
sentry-cli releases new ${CI_COMMIT_SHA}
sentry-cli releases set-commits --auto ${CI_COMMIT_SHA}  # <--- here I get the error
```

It happens that sometimes we get the error
```
error: Couldn’t not find the SHA of the previous release in the git history. Increase your git clone depth.
```
And if I increase the git depth, it shows`error: No commits found. Leaving release alone.`

It has to be said that when merging into master we are squashing commits and that could be the cause of our problems, but we can't figure out exactly because sometimes the set-commits command just work as expected.

kamilogorek:
`--auto` is a flag, and it doesn't accept any additional value, so this last `${CI_COMMIT_SHA}` is ignored.

If you squash or rebase git history, then as you already mentioned there's a chance that hashes will be overridden, and we won't be able to find old release reference. If you cannot change that, then you'll most likely have to explicitly tell the CLI which hash is the one that you want to use as the base, eg.

`sentry-cli releases set-commits "$VERSION" --commit "my-repo@from..to"`

vinnymac:
@kamilogorek your advice is certainly helpful, but the first thing you mention is that `--auto` takes no additional value and that `${CI_COMMIT_SHA}` is ignored. The [documentation](https://docs.sentry.io/product/releases/#associate-commits-with-a-release) for the CLI has examples that show this exact usage however.

<img width="771" alt="Screen Shot 2020-09-12 at 12 00 23 PM" src="https://user-images.githubusercontent.com/1832781/92999535-8b522900-f4ef-11ea-9cb4-eb80e5a37dfa.png">


Is the example wrong, or does the auto flag actually accept a value?

I see competing [examples here](https://docs.sentry.io/cli/releases/#sentry-cli-commit-integration).

<img width="769" alt="Screen Shot 2020-09-12 at 12 01 41 PM" src="https://user-images.githubusercontent.com/1832781/92999551-ba689a80-f4ef-11ea-8795-dc54f178586c.png">

kamilogorek:
@vinnymac good catch, updated my comment. I misread the command 😅

x-yuri:
[...]

PR #963, opened 2021-05-24
feat: Allow for ignoring missing commits in set-commit

kamilogorek:
We _could_ go with the more complex route of fetching all N releases and use their previous commits one by one, until we find a working one, however:
- it wouldn't cut it for fresh projects (a single release which has squashed commit)
- it would require us to fetch all that data from the API, as `get_previous_release_with_commits` that we use currently, is a very specific `/previous-with-commits` endpoint that returns only a single release

This change will allow us to be more "generous" for malformed repositories and it will still work fine with `initial-depth` option. Tested it on production already and works as expected.

Fixes: https://github.com/getsentry/sentry-cli/issues/792
Fixes https://github.com/getsentry/sentry-cli/issues/945
Fixes: https://github.com/getsentry/sentry-cli/issues/956
Fixes: https://github.com/Exelord/ember-cli-deploy-sentry-cli/issues/10

### DIFF
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -104,6 +104,10 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                         This requires that the command is run from within a git repository.  \
                         sentry-cli will then automatically find remotely configured \
                         repositories and discover commits."))
+            .arg(Arg::with_name("ignore-missing")
+                        .help("When the flag is set and the previous release commit was not found in the repository, \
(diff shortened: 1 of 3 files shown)


NisanthanNanthakumar (approved):
@kamilogorek Looks good to me! I think we should add docs for this flag. Since this is primarily an issue that affects the `--local` flow, I think https://docs.sentry.io/product/cli/releases/#alternatively-without-a-repository-integration is a good place.

kamilogorek:
Added https://github.com/getsentry/sentry-docs/pull/3625

PR #967, opened 2021-05-29
[Fix] Allow to define a custom release name for source-maps upload

retyui:
Fix an issue: https://github.com/getsentry/sentry-react-native/issues/1229

Android cli PR: https://github.com/getsentry/sentry-react-native/pull/1548

### DIFF
--- a/src/commands/react_native_xcode.rs
+++ b/src/commands/react_native_xcode.rs
@@ -274,12 +274,10 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
         processor.rewrite(&[base.parent().unwrap().to_str().unwrap()])?;
         processor.add_sourcemap_references()?;
 
-        let default_release_name = format!(
-            "{}@{}+{}",
-            plist.bundle_id(),
-            plist.version(),
-            plist.build()
-        );
+        let default_dist = env::var("SENTRY_DIST").unwrap_or(plist.build().to_string());
+
+        let default_release_name =
+            format!("{}@{}+{}", plist.bundle_id(), plist.version(), default_dist);
 
         let release_name = env::var("SENTRY_RELEASE").unwrap_or(default_release_name);
 
@@ -298,7 +296,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
                     org: &org,
                     project: Some(&project),
                     release: &release.version,
-                    dist: Some(&env::var("SENTRY_DIST").unwrap_or(plist.build().to_string())),
+                    dist: Some(&default_dist),
                     wait: matches.is_present("wait"),
                 })?;
             }
--- a/src/commands/react_native_xcode.rs
+++ b/src/commands/react_native_xcode.rs
@@ -283,7 +283,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
 
         let release_name = env::var("SENTRY_RELEASE").unwrap_or(default_release_name);
 
-        let default_release = api.new_release(
+        let release = api.new_release(
             &org,
             &NewRelease {
                 version: release_name,
@@ -298,7 +298,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
                     org: &org,
                     project: Some(&project),
                     release: &release.version,
-                    dist: env::var("SENTRY_DIST").unwrap_or(Some(&plist.build())),
+                    dist: Some(&env::var("SENTRY_DIST").unwrap_or(plist.build().to_string())),
                     wait: matches.is_present("wait"),
                 })?;
             }
@@ -308,7 +308,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
                         org: &org,
                         project: Some(&project),
                         release: &release.version,
-                        dist: env::var("SENTRY_DIST").unwrap_or(Some(dist)),
+                        dist: Some(&env::var("SENTRY_DIST").unwrap_or(dist.to_string())),
                         wait: matches.is_present("wait"),
                     })?;
                 }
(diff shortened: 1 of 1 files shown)


src/commands/react_native_xcode.rs
```diff
@@ -274,15 +274,19 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
...
 
-        let release = api.new_release(
+        let default_release_name = format!(
+            "{}@{}+{}",
+            plist.bundle_id(),
+            plist.version(),
+            plist.build()
+        );
+
+        let release_name = env::var("SENTRY_RELEASE_NAME").unwrap_or(default_release_name);
```
jennmueng, line 284:
```suggestion
        let release_name = env::var("SENTRY_RELEASE").unwrap_or(default_release_name);
```
Let's call it `SENTRY_RELEASE`.

jennmueng (commented):
Again, thanks for this! Let's call it `SENTRY_RELEASE` and `SENTRY_DIST`. We would also need the dist to be configurable via setting the `SENTRY_DIST` environment variable too and this should be good to go!

cc @bruno-garcia

src/commands/react_native_xcode.rs
retyui: done

src/commands/react_native_xcode.rs
```diff
@@ -274,15 +274,20 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
...
+        let default_release_name = format!(
+            "{}@{}+{}",
+            plist.bundle_id(),
+            plist.version(),
+            plist.build()
+        );
+
+        let release_name = env::var("SENTRY_RELEASE").unwrap_or(default_release_name);
+        
+
```
jennmueng, line 286:
```suggestion
```
Test failed due to extra line break

retyui:
Test for macOS failed (but Linux and Windows test passed successfully) is it ok?

bruno-garcia (commented):
Do we have any tests that validate this behavior?
@kamilogorek might have thoughts here though

jennmueng:
@retyui The Xcode script would only be run on macOS, this means it does not work. Can you verify that it works locally?

retyui:
>  Can you verify that it works locally?


I will try)

1) Build `cargo build --release --locked --target=x86_64-apple-darwin`
2) Copy `target/x86_64-apple-darwin/release/sentry-cli` to `node_modules/@sentry/cli/sentry-cli`
3) Store in repo `git add -f node_modules/@sentry/cli/sentry-cli ; git commit`
4) Update package.json (to use binary from repo) `"postinstall": "git checkout -f -- :/node_modules"`
5) Push to trigger build on CI


^^^^ Could you please verify that I build binary it correctly? @jennmueng

retyui:
Seems it works as expected

```sh
export SENTRY_RELEASE=1.22.333
```

Снимок экрана 2021-06-02 в 23 11 17

retyui:
I added a small fix to handle `SENTRY_DIST` env


Now it works fine

```
export SENTRY_DIST=777
```

image

retyui:
@bruno-garcia @jennmueng cc

kamilogorek:
Instead of using env vars, this PR should add a flag called `release` which can be used instead. This way, we'll make the experience inline with the one we already have for Gradle. Also overriding `dist` is not necessary, as we already do have `--dist` flag that you can use.

retyui:
> Also overriding dist is not necessary, as we already do have --dist flag that you can use.

You have to understand that CLI script will be run by the build phase "Bundle React Native code and images". 
 `../node_modules/@sentry/cli/bin/sentry-cli react-native xcode` in the xCode

And to define `--dist` need to path that script or use env that I propose to do

src/commands/react_native_xcode.rs
```diff
@@ -274,15 +274,17 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
         processor.rewrite(&[base.parent().unwrap().to_str().unwrap()])?;
         processor.add_sourcemap_references()?;
 
+        let default_dist = env::var("SENTRY_DIST").unwrap_or(plist.build().to_string());
+
+        let default_release_name =
+            format!("{}@{}+{}", plist.bundle_id(), plist.version(), default_dist);
+
+        let release_name = env::var("SENTRY_RELEASE").unwrap_or(default_release_name);
```
kamilogorek, line 282:
```suggestion
        let dist = env::var("SENTRY_DIST").unwrap_or(plist.build().to_string());
        let release_name = env::var("SENTRY_RELEASE").unwrap_or(format!("{}@{}+{}", plist.bundle_id(), plist.version(), dist));
```

src/commands/react_native_xcode.rs
```diff
@@ -294,7 +296,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
                     org: &org,
                     project: Some(&project),
                     release: &release.version,
-                    dist: Some(&plist.build()),
+                    dist: Some(&default_dist),
```
kamilogorek, line 299:
```suggestion
                    dist: Some(&dist),
```

src/commands/react_native_xcode.rs
```diff
@@ -304,7 +306,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
                         org: &org,
                         project: Some(&project),
                         release: &release.version,
-                        dist: Some(dist),
+                        dist: Some(&env::var("SENTRY_DIST").unwrap_or(dist.to_string())),
```
kamilogorek, line 309:
```suggestion
                        dist: Some(dist),
```

kamilogorek (changes requested):
In this case, flags should always have precedence over env variables. Because of that, `SENTRY_DIST` should be applied only if `--dist` is absent.

src/commands/react_native_xcode.rs
kamilogorek: This will probably fail formatting :3

kamilogorek (approved):

kamilogorek:
That's what I thought. Thanks `rustfmt` :S Looks good, will merge once formatted is happy.

retyui:
@kamilogorek seems it needs to start CI manually

PR #974, opened 2021-06-05
Adding ignore-missing flag to setCommits types

dscdac:
Hey!

I just noticed that the last release added support for the `--ignore-missing` flag for set-commit.
Unfortunately that field is not in the `SentryCliCommitsOptions` types.
Hope this solves it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## sentry-cli 1.65.1
+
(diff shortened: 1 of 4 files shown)


kamilogorek:
Thanks! Merged manually, as we are not bumping version numbers by hand.

ref: https://github.com/getsentry/sentry-cli/commit/b9dc16bed5110913058aeb3038806b579e51b49a

Issue #980, opened 2021-06-10
Additional properties are not allowed ('debug_id' was unexpected)


Kotevode:
### Environment

How do you use Sentry?
Self-hosted@9.1.2

Which SDK and version?
sentry-cli@1.66.0 through fastlane-plugin-sentry@1.8.1

### Steps to Reproduce

1. Build and archive iOS app with bitcode disabled and DWARF with dSYM file debug information format
2. Run sentry_upload_dsym

### Expected Result

Processed and uploaded symbols

### Actual Result

```
[20:42:10]: --------------------------------
[20:42:10]: --- Step: sentry_upload_dsym ---
[20:42:10]: --------------------------------
[20:42:11]: sentry-cli 1.66.0 installed!
[20:42:11]: Starting sentry-cli...
[20:42:15]: > Found 1 debug information file
[20:42:15]: > Prepared debug information file for upload
[20:42:16]: error: API request failed
[20:42:16]: caused by: sentry reported an error: Additional properties are not allowed ('debug_id' was unexpected) (http status: 400)
```

I assume the latest sentry-cli is incompatible with our self-hosted instance therefore the question is what version of sentry-cli should I use?

flub:
It seems https://github.com/getsentry/sentry/commit/296a423b3653df7b81c9e500fbfd8bd941b7f1d5 introduced this requirement accidentally in sentry gated on the `ChunkedUploadCapability::Pdbs` capability, but accidentally made this schema change not gated on the capability.

The corresponding commit in sentry-cli is https://github.com/getsentry/sentry-cli/commit/9e25d168f7d068f1721377507173d220cf88e882.  From those two commits (in)compatible releases could be figured out I guess.

I'm not sure if fixing the sentry side is worth it, given that this is an ancient release.  sentry-cli could potentially be fixed to not emit the `ChunkedDifRequest::debug_id` field if the `Pdbs` capability is missing, but again this is an old release and the fix is probably harder than it sounds as getting access to the capabilities at the right point in sentry-cli might not be easy.

flub:
That's not the entire compatibility story I think, while that does explain the protocol bug it was not triggered because sentry-cli used to be less eager to supply a debug_id.  https://github.com/getsentry/sentry-cli/blob/e224e0ebf5120b21273b156c105e78e7ac26fc51/src/utils/dif_upload.rs#L122-L125 is probably responsible for being more eager in supplying a `debug_id`, previously I think it just supplied `None` all the time.

[...]

PR #981, opened 2021-06-14
fix(dif-upload): Do not supply debug_id for object files

flub:
The debug_id can be supplied by the client side according to the
protocol, but there is a protocol bug in that this should have been
gated on the ChunkedUploadCapability::Pdbs feature but is not.

Additionally the server-side does this itself if no debug_id is
supplied and we have more control over fixing the server-side code
than people upgrading sentry-cli, so this is preferable.

Lastly this lucky side-effect is the cause of an incompatibility of
newer sentry-cli versions working with very old sentry servers, kind
of accidentally.

Fixes https://github.com/getsentry/sentry-cli/issues/980 (ish)

Please beware, I haven't fully tested upload locally yet with this PR and don't know what integration testing, if any, CI does.


flub:
Hum, this might be too naive and break bcsymbolmaps.  need to check

flub:
So finally tested this and works fine with bcsymbolmaps.  I think this can be merged, @kamilogorek if you'd like to review/approve?

kamilogorek (approved):

PR #984, opened 2021-06-20
Fix type for started and finished deploy options

janicduplessis:
When passing a time number to `started` or `finished` option of `newDeploy` it errors with

```
Object({"dateFinished": Array([String("Datetime has wrong format. Use one of these formats instead: YYYY-MM-DDThh:mm[:ss[.uuuuuu]][+HH:MM|-HH:MM|Z].")]), "dateStarted": Array([String("Datetime has wrong format. Use one of these formats instead: YYYY-MM-DDThh:mm[:ss[.uuuuuu]][+HH:MM|-HH:MM|Z].")])})
```

This updates the type to be a string instead since this is what the cli expects.

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -29,8 +29,8 @@ declare module '@sentry/cli' {
 
   export interface SentryCliNewDeployOptions {
     env: string;
-    started?: number;
-    finished?: number;
+    started?: string;
+    finished?: string;
     time?: number;
     name?: string;
     url?: string;


kamilogorek:
Can you provide a repro for this issue? I tried both, directly calling CLI and through JS API and it works as expected. `started` and `finished` are unix timestamps, so they should be numbers.

```terminal
sentry-cli releases deploys foo new --env bar --started 0 --finished 1337
```

```js
const SentryCli = require("@sentry/cli");

const cli = new SentryCli();

(async () => {
  console.log(await cli.releases.newDeploy("foo", {
    env: "bar",
    started: 0,
    finished: 1337,
  }));
})();
```

In both cases it correctly logs `Created new deploy unnamed for 'bar'`.

kamilogorek:
Closing until repro is provided

janicduplessis:
Sorry I forgot to get back to you on this.

Looks like the issue is that rust timestamps are in seconds while JS timestamps are in milliseconds.

This should repro:

```
const SentryCli = require("@sentry/cli");

const cli = new SentryCli();

(async () => {
  console.log(await cli.releases.newDeploy("foo", {
    env: "bar",
    started: Date.now(),
    finished: Date.now() + 1337,
  }));
})();
```

This works:

```
const SentryCli = require("@sentry/cli");

const cli = new SentryCli();

(async () => {
  console.log(await cli.releases.newDeploy("foo", {
    env: "bar",
    started: Math.floor(Date.now() / 1000),
    finished: Math.floor((Date.now() + 1337) / 1000),
  }));
})();
```

Maybe the JS wrapper should convert from MS to S, or at least be documented. Also TS type could be updated to `number | string` as string dates work.

kamilogorek:
Thanks for the followup @janicduplessis. Fixed it here https://github.com/getsentry/sentry-cli/pull/999

PR #993, opened 2021-07-01
feat: Add --ignore-empty flag that will not bail command when no patchset is created

kamilogorek:
Otherwise, there is no easy way to ignore this error and allow the CI to progress when triggering the same `set-commit` command with no new commits. Simple `|| true` is not sufficient, as it will silent all possible valid errors.

Q: Not sure if we should call it `--allow-empty` or `--ignore-empty` to be consistent with newly added `ignore-missing` flag.

### DIFF
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -108,8 +108,8 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                 .help("When the flag is set and the previous release commit was not found in the repository, \
                         will create a release with the default commits count (or the one specified with `--initial-depth`) \
                         instead of failing the command."))
-            .arg(Arg::with_name("allow-empty")
-                .long("allow-empty")
+            .arg(Arg::with_name("ignore-empty")
+                .long("ignore-empty")
                 .help("When the flag is set, command will not fail and just exit silently \
                         if no new commits for a given release have been found."))
             .arg(Arg::with_name("local")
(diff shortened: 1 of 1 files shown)


src/commands/releases.rs
```diff
@@ -588,7 +592,14 @@ fn execute_set_commits<'a>(
         let commits = generate_patch_set(&repo, commit_log, prev_commit, &parsed)?;
 
         if commits.is_empty() {
-            bail!("No commits found. Leaving release alone.");
+            // NOTE: Make it a default behavior on next major release instead?
```
AbhiPrasad, line 595: Maybe we explicitly say `TODO(v2):` so it's easier to grep? Then we can make a decision on what we wanna do when we think about the next major bump.

src/commands/releases.rs
```diff
@@ -108,6 +108,10 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                 .help("When the flag is set and the previous release commit was not found in the repository, \
                         will create a release with the default commits count (or the one specified with `--initial-depth`) \
                         instead of failing the command."))
+            .arg(Arg::with_name("allow-empty")
+                .long("allow-empty")
```
AbhiPrasad, line 112: Not sure if `allow-empty` is the correct phrase to use here since `--allow-empty` in git adds commits with no content in them. In this case, this option refers to the fact that there is no commits to add to a release - I'm sure someone will get this confused when they try and use the `allow-empty` option in it's current state.

Maybe we should use `ignore-empty`. It won't conflict with git's `--allow-empty` (which may prevent future user misunderstandings), and `ignore` matches what is happening better than `allow` (updating the release is ignored).

AbhiPrasad (changes requested):
Think we should use `ignore-empty`

src/commands/releases.rs
kamilogorek: Valid reasoning. Changed

src/commands/releases.rs
kamilogorek: Updated

src/commands/releases.rs
```diff
@@ -588,7 +592,14 @@ fn execute_set_commits<'a>(
         let commits = generate_patch_set(&repo, commit_log, prev_commit, &parsed)?;
 
         if commits.is_empty() {
-            bail!("No commits found. Leaving release alone.");
+            // TODO(v2): Make it a default behavior on next major release instead?
+            let allow_empty = matches.is_present("ignore-empty");
+            if allow_empty {
```
AbhiPrasad, line 597:
```suggestion
            let ignore_empty = matches.is_present("ignore-empty");
            if ignore_empty {
```

AbhiPrasad (approved):

src/commands/releases.rs
codingjoe: I don't know if that is a good idea. A release could consist of configuration changes through your environment. In those cases, I still want to have a release on Sentry. This could be a very common use case for every one using the 12 factor app approach.

kamilogorek: Noted. Thanks for the feedback!

PR #994, opened 2021-07-02
feat: Add --raw flag to releases list command

kamilogorek:
This makes it way more convenient to run commands like `sentry-cli releases list --raw | xargs sentry-cli releases delete` or `sentry-cli releases list --raw --delimiter " " | xargs -n1 sentry-cli releases delete`when messing around locally with our releases feature.


src/commands/releases.rs
```diff
@@ -170,7 +170,17 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
...
+            .arg(Arg::with_name("raw")
+                .short("R")
+                .long("raw")
+                .help("Print raw, delimiter separated list of releases. [defaults to new line]"))
+            .arg(Arg::with_name("delimiter")
+                .short("D")
+                .long("delimiter")
+                .takes_value(true)
+                .requires("raw")
+                .help("Delimiter for the --raw flag")))
```
AbhiPrasad, line 183: Do you think we should set a `default_value` to a `\n`? here instead of unwrapping the option below?

AbhiPrasad (approved):
Can we add a quick test for this? Prob just need one with only `--raw` and then `--delimeter + --raw`. It'll also help with seeing why these options were added. 

Approving to unblock though.

src/commands/releases.rs
kamilogorek: No, because then it'd make `--raw` always required per `requires('raw')` is triggered when default value is provided.
Hence this commit: `Use onwrap_or instead of default_value to make it non-required`

kamilogorek: Added tests

AbhiPrasad: Makes sense 👍

PR #995, opened 2021-07-06
feat: Print sourcemaps/files operations timings information

kamilogorek:
### Before:
image

### After
image

### DIFF
--- a/src/utils/dif_upload.rs
+++ b/src/utils/dif_upload.rs
@@ -655,7 +655,7 @@ fn search_difs(options: &DifUpload) -> Result<Vec<DifMatch<'static>>, Error> {
         }
         walk_difs_directory(base_path, options, |source, name, buffer| {
             debug!("trying to process {}", name);
-            progress.set_message(&name);
+            progress.set_message(name.to_owned());
 
             if Archive::peek(&buffer) != FileFormat::Unknown {
                 let mut difs =
@@ -671,7 +671,7 @@ fn search_difs(options: &DifUpload) -> Result<Vec<DifMatch<'static>>, Error> {
                 }
             };
 
-            progress.set_prefix(&collected.len().to_string());
+            progress.set_prefix(collected.len().to_string());
             Ok(())
         })?;
     }
@@ -941,7 +941,7 @@ where
     let mut calculated = Vec::new();
     for item in items {
         progress.inc(1);
-        progress.set_message(item.path());
+        progress.set_message(item.path().to_owned());
         calculated.push(func(item)?);
     }
 
@@ -1001,7 +1001,7 @@ fn process_symbol_maps<'a>(
 
     for dif in with_hidden {
         progress.inc(1);
-        progress.set_message(dif.path());
+        progress.set_message(dif.path().to_owned());
         without_hidden.push(resolve_hidden_symbols(dif, symbol_map)?);
     }
 
@@ -1052,7 +1052,7 @@ fn create_source_bundles<'a>(difs: &[DifMatch<'a>]) -> Result<Vec<DifMatch<'a>>,
 
     for dif in difs {
         progress.inc(1);
-        progress.set_message(dif.path());
+        progress.set_message(dif.path().to_owned());
 
         let object = match dif.object() {
             Some(object) => object,
--- a/src/utils/progress.rs
+++ b/src/utils/progress.rs
@@ -47,8 +47,9 @@ impl ProgressBar {
     }
 
     pub fn finish_with_duration(&self, op: &str) {
-        let e = self.inner.elapsed();
-        let msg = format!("{} completed in {}.{}s", op, e.as_secs(), e.as_millis());
+        let dur = self.inner.elapsed();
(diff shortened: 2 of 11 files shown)


src/utils/dif_upload.rs
```diff
@@ -655,7 +655,7 @@ fn search_difs(options: &DifUpload) -> Result<Vec<DifMatch<'static>>, Error> {
         }
         walk_difs_directory(base_path, options, |source, name, buffer| {
             debug!("trying to process {}", name);
-            progress.set_message(&name);
+            progress.set_message(name.to_owned());
```
AbhiPrasad, line 658: Rust Q, do we always prefer `to_owned` over `clone` because it's more generic? (`to_owned` might not clone right? - depends on the trait impl)

src/utils/progress.rs
```diff
@@ -42,6 +42,16 @@ impl ProgressBar {
...
+    }
+
+    pub fn finish_with_duration(&self, op: &str) {
+        let e = self.inner.elapsed();
+        let msg = format!("{} completed in {}.{}s", op, e.as_secs(), e.as_millis());
+        let progress_style = ProgressStyle::default_bar().template("{prefix:.dim} {msg}");
+        self.inner.set_style(progress_style);
+        self.inner.set_prefix(">");
         self.inner.finish_with_message(msg);
         logging::set_progress_bar(None);
```
AbhiPrasad, line 56: Shouldn't we use `self.finish_with_message(&msg)` to avoid duplicating the logic here?

lobsterkatie (commented):
I can't speak to the rust-specific changes, but overall this looks good, and I definitely think it will be helpful for debugging!

Nit: There's inconsistency in terms of what we log and when. Sometimes it's

```
Doing X
Did X in 123 ms
```

sometimes it's

```
Did X in 123 ms
Did X to 5 files
```

sometimes it's

```
Doing X
```

and sometimes it's

```
Did X in 123 ms
```

In a perfect world, IMHO it'd be

```
Doing X
    Did X in 123 ms
```

or

```
Doing X
    Did X to 5 files in 123 ms
```

as makes sense in context.

(I recognize this isn't a perfect world, though, so feel free to take this with however many grains of salt you'd like.)

src/utils/progress.rs
```diff
@@ -42,6 +42,16 @@ impl ProgressBar {
     }
 
     pub fn finish_with_message(&self, msg: &str) {
+        self.inner.finish_with_message(msg.to_owned());
+        logging::set_progress_bar(None);
+    }
+
+    pub fn finish_with_duration(&self, op: &str) {
+        let e = self.inner.elapsed();
+        let msg = format!("{} completed in {}.{}s", op, e.as_secs(), e.as_millis());
```
lobsterkatie, line 51: Do `as_secs()` and `as_millis()` give the second and millisecond parts of the time, or do they do what their names imply and give the same time expressed in different units? 

If the latter, it seems like this would be problematic in the case of something which took, say, 873 ms - in seconds that's 1 second, so you'd end up with `completed in 1.873s`, which obviously isn't correct.

src/utils/progress.rs
```diff
@@ -42,6 +42,16 @@ impl ProgressBar {
     }
 
     pub fn finish_with_message(&self, msg: &str) {
+        self.inner.finish_with_message(msg.to_owned());
+        logging::set_progress_bar(None);
+    }
+
+    pub fn finish_with_duration(&self, op: &str) {
```
AbhiPrasad, line 49: you think we can include extra data other than the operation here so the message is more detailed?

src/utils/progress.rs
kamilogorek: You are totally correct, great catch, thanks!

src/utils/dif_upload.rs
kamilogorek: `clone()` on `&str` returns a `&str`, where in this case `to_owned` will return `String` which we need for `set_message`. I prefer `to_owned` as it's more explicit and obvious what are you want to get out of that call.

src/utils/progress.rs
kamilogorek: I was thinking about it but decided no to do that to keep everything in line. This way, every method is a standalone thing that just proxies data to `indicatif` methods. It's very rarely touched part of code and adds 1 more line of code anyway.

src/utils/progress.rs
kamilogorek: Sure we could. Anything you have in mind? I kept it concise so it reads nicely 😅

kamilogorek:
@lobsterkatie totally agrees. The problem here is that some messages are displayed using raw `println!()` and some through the progress bar. Those that use the progress bar are disregarded after the operation is complete and overwritten by estimates now, that's why sometimes it may look out of order or illogical. We also don't have all the data stored necessary to show things like `Added source map references to X files`. But it's definitely something we can iterate over.

src/utils/progress.rs
AbhiPrasad: Yeah I got nothing for now, but I guess we can always come back and iterate.

AbhiPrasad (approved):

PR #999, opened 2021-07-15
ref: Provide JSDocs for TS typings and change started/finished to number|string

kamilogorek:
Fixes https://github.com/getsentry/sentry-cli/pull/984

`started` and `finished` can use either timestamp in seconds or ISO 8601 format, so it _can_ accept a string.

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -4,7 +4,7 @@
 declare module '@sentry/cli' {
   export interface SentryCliOptions {
     /**
-     * Sentry instance
+     * The URL of the Sentry instance you are connecting to. Defaults to https://sentry.io/.
      */
     url?: string;
     /**
@@ -28,7 +28,7 @@ declare module '@sentry/cli' {
      */
     project?: string;
     /**
-     * VCS remote name
+     * Version control system remote name
      */
     vscRemote?: string;
     /**
@@ -44,22 +44,22 @@ declare module '@sentry/cli' {
 
   export interface SentryCliUploadSourceMapsOptions {
     /**
-   * One or more paths that Sentry CLI should scan recursively for sources.
-   * It will upload all .map files and match associated .js files.
-   */
+     * One or more paths that Sentry CLI should scan recursively for sources.
+     * It will upload all .map files and match associated .js files.
+     */
     include: string | string[];
     /**
-   * One or more paths to ignore during upload. Overrides entries in ignoreFile file.
-   */
+     * One or more paths to ignore during upload. Overrides entries in ignoreFile file.
+     */
     ignore?: string[];
     /**
-   * Path to a file containing list of files/directories to ignore.
-   * Can point to .gitignore or anything with same format.
-   */
+     * Path to a file containing list of files/directories to ignore.
+     * Can point to .gitignore or anything with same format.
+     */
     ignoreFile?: string | null;
     /**
      * Enables rewriting of matching sourcemaps so that indexed maps are flattened
-     * and missing sources are inlined if possible., defaults to `true`.
+     * and missing sources are inlined if possible. Defaults to `true`.
      */
     rewrite?: boolean;
     /**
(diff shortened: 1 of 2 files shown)


js/index.d.ts
```diff
@@ -1,47 +1,159 @@
-/*
-  Typings for @sentry/cli
-*/
+/**
+ * Typings for @sentry/cli
+ */
 declare module '@sentry/cli' {
   export interface SentryCliOptions {
+    /**
+     * Sentry instance
```
AbhiPrasad, line 7: This may be a bit brittle, but can we reference the env variables https://github.com/getsentry/sentry-cli/blob/49eeb9aab399cb92b097c3a6c213c9c8dcf0c0df/js/helper.js#L142 that the config options update? That way people can also search up the env variables and get see associated docs/GH issues.

AbhiPrasad, line 7:
```suggestion
     * The URL of the Sentry instance you are connecting to. Defaults to https://sentry.io/.
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+     */
     url?: string;
+    /**
+     * Authentication token for API, interchangeable with `apiKey`
+     */
     authToken?: string;
+    /**
+     * Authentication token for API, interchangeable with `authToken`
```
AbhiPrasad, line 15: Is `apiKey` deprecated?

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+    /**
+     * Organization slug
+     */
     org?: string;
+    /**
+     * Project Project slug
+     */
     project?: string;
+    /**
+     * VCS remote name
```
AbhiPrasad, line 31:
```suggestion
     * Version control system remote name
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+     * If true, all logs are suppressed
+     */
     silent?: boolean;
   }
 
   export interface SentryCliUploadSourceMapsOptions {
+    /**
+   * One or more paths that Sentry CLI should scan recursively for sources.
+   * It will upload all .map files and match associated .js files.
+   */
```
AbhiPrasad, line 49:
```suggestion
    /**
     * One or more paths that Sentry CLI should scan recursively for sources.
     * It will upload all .map files and match associated .js files.
     */
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
     include: string | string[];
+    /**
+   * One or more paths to ignore during upload. Overrides entries in ignoreFile file.
+   */
```
AbhiPrasad, line 53:
```suggestion
    /**
     * One or more paths to ignore during upload. Overrides entries in ignoreFile file.
     */
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
     ignore?: string[];
+    /**
+   * Path to a file containing list of files/directories to ignore.
+   * Can point to .gitignore or anything with same format.
+   */
```
AbhiPrasad, line 58:
```suggestion
    /**
     * Path to a file containing list of files/directories to ignore.
     * Can point to .gitignore or anything with same format.
     */
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
     ignoreFile?: string | null;
+    /**
+     * Enables rewriting of matching sourcemaps so that indexed maps are flattened
+     * and missing sources are inlined if possible., defaults to `true`.
```
AbhiPrasad, line 62:
```suggestion
     * and missing sources are inlined if possible. Defaults to `true`.
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+     */
     rewrite?: boolean;
+    /**
+     * This prevents the automatic detection of sourcemap references.
+     */
     sourceMapReference?: boolean;
+    /**
+     * When paired with rewrite this will chop-off a prefix from uploaded files.
```
AbhiPrasad, line 70:
```suggestion
     * When paired with the rewrite option this will remove a prefix from uploaded files.
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+     * For instance you can use this to remove a path that is build machine specific.
+     */
     stripPrefix?: string[];
+    /**
+     * When paired with rewrite this will add ~ to the stripPrefix array.
```
AbhiPrasad, line 75:
```suggestion
     * When paired with the rewrite option this will add ~ to the stripPrefix array.
```

js/index.d.ts
```diff
@@ -1,47 +1,159 @@
...
+     */
     stripCommonPrefix?: boolean;
+    /**
+     * This attempts sourcemap validation before upload when rewriting is not enabled.
+     * It will spot a variety of issues with source maps and cancel the upload if any are found.
+     * This is not the default as this can cause false positives.
```
AbhiPrasad, line 81:
```suggestion
     * This is not enabled by default as this can cause false positives.
```

AbhiPrasad (approved):

js/index.d.ts
kamilogorek: I think so, although it's still fully functional. It's a cascade of `SENTRY_AUTH_TOKEN > SENTRY_API_KEY > config.token > config.api_key`

https://github.com/getsentry/sentry-cli/blob/758d6ffb324691f54223098403f1d35684584dc0/src/config.rs#L542-L554

Issue #1007, opened 2021-08-11
Progress Bar breaking npm install


mdgozza:
### Environment

How do you use Sentry?
We used Sentry SaaS (sentry.io)

Which SDK and version?
Version 1.68.0

### Steps to Reproduce

1. Have machine ssl traffic intercepted by Enterprise Security Platform Netskope.
2. Attempt to npm install @senty/cli

### Expected Result

npm package should install successfully and download the proper binary.

### Actual Result

It would appear that netskope intercepting https traffic is causing a loss of the http header `content-length`. This setting is not configurable in netskope.

When `content-length` is missing, it causes the `total` variable to become `NaN`
https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js#L176

Later on in the code, when the progress bar tick method is called, a breaking error is thrown from `node-progress`
https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js#L184

```
/Users/marcellog/projects/sentry-cli/node_modules/progress/lib/node-progress.js:160
  complete = Array(Math.max(0, completeLength + 1)).join(this.chars.complete);
             ^

RangeError: Invalid array length
    at ProgressBar.render (/Users/marcellog/projects/sentry-cli/node_modules/progress/lib/node-progress.js:160:14)
    at ProgressBar.tick (/Users/marcellog/projects/sentry-cli/node_modules/progress/lib/node-progress.js:97:8)
    at PassThrough.<anonymous> (/Users/marcellog/projects/sentry-cli/scripts/install.js:184:54)
    at PassThrough.emit (events.js:388:22)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at PassThrough.Readable.push (internal/streams/readable.js:223:10)
    at PassThrough.Transform.push (internal/streams/transform.js:166:32)
    at PassThrough.afterTransform (internal/streams/transform.js:101:10)
    at PassThrough._transform (internal/streams/passthrough.js:46:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @sentry/cli@1.68.0 install: `node scripts/install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @sentry/cli@1.68.0 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

PR #1008, opened 2021-08-11
(issue #1007)  prevent progress bar from breaking install

mdgozza:
See Issue #1007 

Prevent the progress bar renderer from breaking when content-length http header does not exist ( or any other kind of weird breaking error )


kamilogorek:
Closed in favor of https://github.com/getsentry/sentry-cli/pull/1011

Issue #1006, opened 2021-08-11
Configure npmlog in install.js to write to stdout instead of stderr


CeciliaSilvera:
In `install.js` you log using `npmlog` and it per default logs to `stderr`. Checking **"Fail on standard error"** in our Azure CI build means that all logs to stderr gets interpreted as errors and builds will break. Even if you log with `npmLog.info` as you are in this file since it gets printed to that  output stream.
It would be much appreciated if you could set the output stream to stdout for at least anything other than error logging.

from [npmlog](https://github.com/npm/npmlog) :
_By default, logs are written to stderr. If you want to send log messages to outputs other than streams, then you can change the log.stream member, or you can just listen to the events that it emits, and do whatever you want with them_

PR #1010, opened 2021-08-17
feat: Allow for changing log stream through SENTRYCLI_LOG_STREAM variable

kamilogorek:
Not that we cannot do this change globally in one place, as this would be a breaking change due to the fact that install logs went to `stderr` and progress bar to `stdout` in the current implementation.

Closes https://github.com/getsentry/sentry-cli/issues/1006

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -28,6 +28,22 @@ const CDN_URL =
   process.env.SENTRYCLI_CDNURL ||
   'https://downloads.sentry-cdn.com/sentry-cli';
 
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):
Do we need a docs PR for this?

kamilogorek:
I don't think it's necessary for now. It's kind of an edge-case, which can be easily verified in the codebase. We don't document any `SENTRYCLI_` variables from what I checked.

PR #1024, opened 2021-09-19
fix: Set archString to "armv7" when arch="arm"

### LINKED ISSUE #1023: Cannot download binary in armv7 environment
### Environment

Docker container(node:14-buster) running on Raspberry pi 4. 
sentry-cli version: 1.68.0

### Steps to Reproduce

1. install Docker on Raspberry pi 4(Raspberry pi OS).
2. install gitlab-runner and configure it
3. make a project with @sentry/webpack-plugin in its package.json
4. add gitlab-ci.yml, which has `npm ci --cache .npm --prefer-offline` in its before_script
5. push the project, and run ci.

I don't think gitlab is related to this issue, as I will explain.

### Expected Result

`npm ci` finishes without any error.

### Actual Result

In some cases `npm ci` stops with error, 
```
> @sentry/cli@1.68.0 install /builds/**/**/node_modules/@sentry/cli
> node scripts/install.js
info sentry-cli Downloading from https://downloads.sentry-cdn.com/sentry-cli/1.68.0/sentry-cli-Linux-arm
Error: Unable to download sentry-cli binary from https://downloads.sentry-cdn.com/sentry-cli/1.68.0/sentry-cli-Linux-arm.
Server returned 403: Forbidden.
```
but in other cases `npm ci` hangs, with the following log.
```
> @sentry/cli@1.68.0 install /builds/**/**/node_modules/@sentry/cli
> node scripts/install.js
info sentry-cli Downloading from https://downloads.sentry-cdn.com/sentry-cli/1.68.0/sentry-cli-Linux-arm
ERROR: Job failed: execution took longer than 1h0m0s seconds
```

I think the problem is that the install script is trying to fetch `https://downloads.sentry-cdn.com/sentry-cli/1.68.0/sentry-cli-Linux-arm`, while the correct url is `https://downloads.sentry-cdn.com/sentry-cli/1.68.0/sentry-cli-Linux-armv7` according to https://sentry.io/get-cli/ .(This is why I think gitlab is unrelated)
I suggest updating scripts/install.js 's getDownloadUrl as following, though I have not tested it (I don't know how to test it).
```
function getDownloadUrl(platform, arch) {
  const releasesUrl = `${CDN_URL}/${pkgInfo.version}/sentry-cli`;
  let archString = '';
  switch (arch) {
    case 'x64':
      archString = 'x86_64';
      break;
    case 'x86':
    case 'ia32':
      archString = 'i686';
      break;
    case 'arm64':
      archString = 'aarch64';
      break;
+    case 'arm':
+      archString = 'armv7';
+      break;
    default:
      archString = arch;
  }
```

I would appreciate it if this problem is fixed.

femshima:
When arch="arm", set archString="armv7", so that the install script will download the correct binary.

Closes #1023

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -66,6 +66,9 @@ function getDownloadUrl(platform, arch) {
     case 'arm64':
       archString = 'aarch64';
       break;
+    case 'arm':
+      archString = 'armv7';
+      break;
     default:
       archString = arch;
   }


kamilogorek (approved):

kamilogorek:
Thanks!

PR #1074, opened 2021-11-16
feat: add SENTRYCLI_SKIP_DOWNLOAD flag for preventing download

### LINKED ISSUE #1066: Skip downloading binary
Hi, I was wondering how I can prevent the download in my CI jobs where sentry cli is not used? I see approaches like using `--ignore-scripts` but that will affect other packages as well.

I'd suggest adding an environment variable, say `SENTRY_SKIP_DOWNLOAD`, which tells that a download is not needed when set to `1`. WDYT?

SevenOutman:
Close #1066

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -259,6 +259,11 @@ if (process.env.SENTRYCLI_LOCAL_CDNURL) {
 
 npmLog.stream = getLogStream('stderr');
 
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):
Looks reasonable, thanks!

PR #1094, opened 2022-01-11
ref: Box ParseDif::Object value to prevent large enum variant

kamilogorek:
In the latest version of clippy, `large_enum_variant` is turned on by default, which fails with the current implementation.

Two removed fields from `XcodeProjectInfo` were unused.


flub (approved):
kind of silly, but i don't have a better suggestion...

PR #1103, opened 2022-01-20
ref: Rename ini group from dsym to dif with a fallback

kamilogorek:
Rename previously called `dsym` group in our INI config files to `dif`, while still keeping it backward compatible.
`dif` has priority over `dsym` if both exist.


AbhiPrasad (approved):
What's the strategy to update https://github.com/getsentry/sentry-docs/blob/master/src/docs/product/cli/configuration.mdx? Do we wait a couple of releases and then update (since it's backwards compatible)?

kamilogorek:
I will add an appropriate note about this change there.

PR #1104, opened 2022-01-21
fix: Prevent daemonize mode from crashing upload process

kamilogorek:
In macOS 10.13, changes to Objective-C runtime forbid calling ObjC code before and after forking the process. If we want to do that, and we need to, due to moving `sentry-cli` process to the background during XCode artifacts upload, we need to perform all those calls *after* we daemonize ourselves. This forces us to not spawn any processes, nor call any ObjC-compiled code (like cURL) before we actually get to that branch.

The thing that's detaching us is `MayDetach::wrap` via `MayDetach::may_detach` function.

If we call any cURL code or spawn any process before-hand, it will result in the background process crashing with:

```
objc[80549]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[80549]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

This PR does the following:
- disable signal handling for processes spawned via XCode (prevent direct process spawning); we can afford that, as our upload process is a standalone `Build Script`, and not part of the larger mechanism, thus we never expect it to be interrupted.
- disable default built-in crash reporting (prevent process spawning in SDK transport constructors); this part was never used for years in the first place (see: https://github.com/getsentry/sentry-cli/commit/2b6bedbf7652682482e91cc7a526fc4d0b364a09) and it was always meant to be used for us to track CLI issues. Original initialization code was still in place though, so it was executed despite lacking the necessary DSN (it initialized empty client).
- move API initialization inside the wrapped function (prevent calling internal curl code);

ref: https://www.wefearchange.org/2018/11/forkmacos.rst.html
ref: http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ predicates = "2.0.0"
 tempfile = "3.1.0"
 
 [features]
-default = ["with_crash_reporting"]
+default = []
 managed = []
 with_crash_reporting = []
 
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -283,8 +283,7 @@ fn setup() {
     }
     #[cfg(not(feature = "with_crash_reporting"))]
     {
-        static LOGGER: Logger = Logger;
-        log::set_logger(&Logger);
+        log::set_logger(&Logger).unwrap();
     }
 }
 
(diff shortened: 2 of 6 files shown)


mitsuhiko (approved):

marandaneto:
@kamilogorek is it related to https://github.com/getsentry/sentry-react-native/issues/1424 and the `--force_foreground` issue?

kamilogorek:
Yes, that's exactly that.

PR #1107, opened 2022-01-25
feat: Allow for adding sourcemaps refs to non-minified JS sources

kamilogorek:
This flag allows for adding sourcemaps references to non-minified sources. This is useful for compile-to-JS languages/frameworks, that does not perform minification, but still produce source-maps.

From what I discussed with @lobsterkatie, this is the behavior that Next.js has with it's server-side code, that's compiled with SWC from TypeScript down to JavaScript.
The problem there is that we don't have a control over filenames, thus we cannot use the `.min.js` branch for detecting whether code is minified - https://github.com/getsentry/sentry-cli/blob/405a6fee86107daac0122afc717b131428c70e47/src/utils/sourcemaps.rs#L218 and compiled code is not detected as such by the heuristics in `might_be_minified`,

Still to be discussed whether it's a reasonable solution.

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -86,9 +86,9 @@ declare module '@sentry/cli' {
      */
     sourceMapReference?: boolean;
     /**
-     * Enables adding sourcemaps references to non-minified sources.
+     * Enables adding sourcemap references to non-minified sources.
      * This is useful for compile-to-JS languages/frameworks,
-     * that does not perform minification, but still produce source-maps.
+     * that do not perform minification, but still produce source-maps.
      */
     processNonminified?: boolean;
     /**
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -86,6 +86,12 @@ declare module '@sentry/cli' {
      */
     sourceMapReference?: boolean;
     /**
+     * Enables adding sourcemaps references to non-minified sources.
+     * This is useful for compile-to-JS languages/frameworks,
+     * that does not perform minification, but still produce source-maps.
+     */
+    processNonminified?: boolean;
+    /**
      * When paired with the rewrite option this will remove a prefix from uploaded files.
      * For instance you can use this to remove a path that is build machine specific.
      */
--- a/src/commands/releases.rs
+++ b/src/commands/releases.rs
@@ -289,9 +289,9 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                     .help("Wait for the server to fully process uploaded files."))
                 .arg(Arg::with_name("process_nonminified")
                     .long("process-nonminified")
-                    .help("Enables adding sourcemaps references to non-minified sources.{n}\
+                    .help("Enables adding sourcemap references to non-minified sources.{n}\
                            This is useful for compile-to-JS languages/frameworks, \
-                           that does not perform minification, but still produce source-maps.")
+                           that do not perform minification, but still produce source-maps.")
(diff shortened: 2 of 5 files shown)


js/index.d.ts
```diff
@@ -85,6 +85,12 @@ declare module '@sentry/cli' {
      * This prevents the automatic detection of sourcemap references.
      */
     sourceMapReference?: boolean;
+    /**
+     * Enables adding sourcemaps references to non-minified sources.
+     * This is useful for compile-to-JS languages/frameworks,
+     * that does not perform minification, but still produce source-maps.
```
iker-barriocanal, line 91:
```suggestion
     * Enables adding sourcemap references to non-minified sources.
     * This is useful for compile-to-JS languages/frameworks,
     * that do not perform minification, but still produce source-maps.
```

src/commands/releases.rs
```diff
@@ -287,6 +287,12 @@ pub fn make_app<'a, 'b: 'a>(app: App<'a, 'b>) -> App<'a, 'b> {
                 .arg(Arg::with_name("wait")
                     .long("wait")
                     .help("Wait for the server to fully process uploaded files."))
+                .arg(Arg::with_name("process_nonminified")
+                    .long("process-nonminified")
+                    .help("Enables adding sourcemaps references to non-minified sources.{n}\
+                           This is useful for compile-to-JS languages/frameworks, \
+                           that does not perform minification, but still produce source-maps.")
```
iker-barriocanal, line 294:
```suggestion
                    .help("Enables adding sourcemap references to non-minified sources.{n}\
                           This is useful for compile-to-JS languages/frameworks, \
                           that do not perform minification, but still produce source-maps.")
```

src/utils/sourcemaps.rs
```diff
@@ -176,10 +177,21 @@ impl SourceMapProcessor {
     pub fn new() -> SourceMapProcessor {
         SourceMapProcessor {
             pending_sources: HashSet::new(),
+            process_nonminified: false,
             sources: HashMap::new(),
         }
     }
 
+    /// Set whether we should add sourcemap references to non-minified sources.
+    /// This is useful for compile-to-JS languages/frameworks, that does not perform minification,
```
iker-barriocanal, line 186:
```suggestion
    /// This is useful for compile-to-JS languages/frameworks, that do not perform minification,
```

src/utils/sourcemaps.rs
```diff
@@ -176,10 +177,21 @@ impl SourceMapProcessor {
...
+    /// but still produce sourcemaps.
+    ///
+    /// Defaults to `false`.
+    pub fn process_nonminified(&mut self, allow: bool) -> &mut Self {
```
iker-barriocanal, line 190: Why return `&mut Self`?

iker-barriocanal (approved):

src/utils/sourcemaps.rs
kamilogorek: Builders convention that we use in most places in the CLI. Allows you to chain creating of complex structs. eg.

```rust
DifUpload::new(org.clone(), project.clone())
    .wait(matches.is_present("wait"))
    .search_paths(matches.values_of("paths").unwrap_or_default())
    .allow_zips(!matches.is_present("no_zips"))
    .filter_ids(ids);
```

iker-barriocanal: Makes sense :D

kamilogorek:
Let close this for now

Issue #1093, opened 2021-12-29
Possible workaround for Xcode 13 Plist issue


armcknight:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
Cocoa 7.7.0

Xcode version?
13+

### Steps to Reproduce

1. Follow iOS onboarding instructions
2. Choose "Without BitCode" path
3. Add _Run Script_ Build Phase to Xcode target
4. Try to run the app in simulator

### Expected Result

App target builds and runs successfully, with results displaying in dashboard afterwards.

### Actual Result

Build fails due to an error parsing the plist and inability to find the `CFBundleName`.

As we note in our [docs](https://docs.sentry.io/platforms/apple/dsym/):

> Starting with Xcode 13, newly generated projects do not automatically create a required Info.plist configuration file. If this is your situation, please read Info.plist Is Missing in Xcode 13 — Here’s How To Get It Back

however, it took me a while to track down that this was actually what was causing my build to fail

### Recommendation

By default, the Plist that Xcode generates sets the `CFBundleName` to `$(PRODUCT_NAME)`. If we read `PRODUCT_NAME` environment variable from Xcode that is passed to the _Run Script_ Build Phase executable (see also: https://github.com/getsentry/sentry-cli/issues/1087), we should be able to avoid this error for the majority of cases, and all cases where someone is just spinning up a test Xcode project app to try out Sentry.

armcknight:
@getsentry/team-mobile

philipphofmann:
Thanks, @armcknight, for the issue. I think many users chose to upload the dSYMs with the Run Script Phase in Xcode when trialing Sentry. Fixing this issue will prevent a bad first impression of Sentry for potential new users, and therefore I think it is essential to fix this.

PR #1111, opened 2022-01-26
feat: Use default xcode values for plist struct

kamilogorek:
Resolves https://github.com/getsentry/sentry-cli/issues/1093


armcknight (approved):
✨

PR #1117, opened 2022-02-07
meta: Update symbolic to 8.6.0

Swatinem:

PR #1118, opened 2022-02-07
build: Update symbolic to 8.6.0

jan-auer:
Fixes a panic when inspecting debug files larger than 4GB.


jan-auer:
Duplicate of #1117

PR #1123, opened 2022-02-17
feat: Add checksum validation for installed binaries

kamilogorek:
Checksum validation is performed only for newly downloaded binaries, not cached ones. It also bails out if a checksum was not found due to an incorrect name or some CI misconfiguration. It also contains a new `SENTRYCLI_CHECKSUM_VALIDATION` flag that allows you to skip it in case something goes wrong but you still want to proceed.
Those fail-safe mechanisms are there only in case something goes wrong, and we don't want to break people's stuff.

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -150,6 +150,43 @@ function getTempFile(cached) {
     .slice(2)}.tmp`;
 }
 
+function validateChecksum(tempPath, name) {
+  let storedHash;
+  try {
+    const checksums = fs.readFileSync(path.join(__dirname, '../checksums.txt'), 'utf8');
+    const entries = checksums.split('\n');
+    for (let i = 0; i < entries.length; i++) {
+      const [key, value] = entries[i].split('=');
+      if (key === name) {
+        storedHash = value;
+        break;
+      }
+    }
+  } catch (e) {
+    npmLog.info('Checksums are only available for npm published package. Skipping validation.');
+    return;
+  }
+
+  if (!storedHash) {
+    npmLog.info(`Checksum for ${name} not found, skipping validation.`);
+    return;
+  }
+
+  const currentHash = crypto
+    .createHash('sha256')
+    .update(fs.readFileSync(tempPath))
+    .digest('hex');
+
+  if (storedHash !== currentHash) {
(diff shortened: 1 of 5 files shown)


scripts/install.js
```diff
@@ -217,6 +254,9 @@ function downloadBinary() {
           .on('error', e => reject(e))
           .on('close', () => resolve());
       }).then(() => {
+        if (process.env.SENTRYCLI_CHECKSUM_VALIDATION !== '1') {
```
AbhiPrasad, line 257: let's explicitly put skip in so it's clear what it's being used for?

```suggestion
        if (process.env.SENTRYCLI_SKIP_CHECKSUM_VALIDATION !== '1') {
```

scripts/install.js
```diff
@@ -150,6 +150,43 @@ function getTempFile(cached) {
     .slice(2)}.tmp`;
 }
 
+function validateChecksum(tempPath, name) {
+  let storedHash;
+  try {
+    const checksums = fs.readFileSync(path.join(__dirname, '../checksums.txt'), 'utf8');
+    const entries = checksums.split('\n');
```
AbhiPrasad, line 157: Should we split up `fs` operation and the checksum string operations into two separate try catches? I think if the read file operation fails we should throw an error.

AbhiPrasad (approved):

scripts/install.js
kamilogorek: It will fail when used locally on purpose, as there are no checksums in the repo, so we cannot throw then

scripts/install.js
```diff
@@ -150,6 +150,43 @@ function getTempFile(cached) {
...
+    for (let i = 0; i < entries.length; i++) {
+      const [key, value] = entries[i].split('=');
+      if (key === name) {
+        storedHash = value;
+        break;
+      }
+    }
+  } catch (e) {
+    npmLog.info('Checksums are only available for npm published package. Skipping validation.');
```
kamilogorek, line 166:
```suggestion
    npmLog.info('Checksums are generated when the package is published to npm. They are not available directly in the source repository. Skipping validation.');
```

Issue #1126, opened 2022-02-25
Sentry crashing without proper debug information when . xcodeproj is not valid


grzegorzkrukowski:
### Environment

How do you use Sentry?
Sentry CLI - directly pushed to GIT repository. I am running it on GitHub workers - tried both macOS and Linux - both failing

Which SDK and version?
sentry-cli 1.73.0

### Steps to Reproduce

I lately changed our CLI setup to keep `Package.resolved` files in the repository for caching purposes. 
This makes my repository have empty `.xcodeproj` that just has `Package.resolved` in it - no more files from typical xcodeproj exists there.

With that state - SentryCLI is failing, differently for different environments:
- macOS - crashing with `thread 'unnamed' panicked at 'Command `xcodebuild -list -json -project`
- ubuntu - crashing with `error: No such file or directory (os error 2)`

### Expected Result

I would expect Sentry to just ignore discovering InfoPlist file in that case - maybe with a warning, but not crashing.
If that's expected behavior - the errors should be more readable - especially for Ubuntu - took me half a day to figure out what's actually happening there.
Also would be nice to just have a flag to skip this auto-discovery if it cannot be silently skipped without errors.

### Actual Result

Sentry failed to upload my pre-generated dSYMs.

kamilogorek:
Can you run the command with `SENTRY_LOG_LEVEL=debug` and provide the log please?

grzegorzkrukowski:
Hey @kamilogorek - here you go - I had to hide all private stuff with ***
It's really easy to reproduce by just creating a `.xcodeproj` and deleting everything beside `Package.resolved` in it.

```
  DEBUG   2022-02-28 13:34:22.762848 +01:00 sentry-cli version: 1.73.0, platform: "darwin", architecture: "x86_64"
  INFO    2022-02-28 13:34:22.763676 +01:00 sentry-cli was invoked with the following command line: "sentry-cli" "--auth-token" "*****************" "upload-dif" "dSYMs" "--org" "*******" "--project" "******" "--log-level=debug"
  INFO    2022-02-28 13:34:22.764024 +01:00 Issuing a command for Organization: ******** Project: *******
thread 'unnamed' panicked at 'Command `xcodebuild -list -json -project *****.xcodeproj` failed to produce a valid JSON output.': src/utils/xcode.rs:120

   0: backtrace::backtrace::trace

Abort trap: 6
```

PR #1127, opened 2022-02-28
ref: Dont panic on malformed xcodeproj directories

kamilogorek:
There are scenarios like the one lined below, where `xcodeproj` directories does not contain all the required files for `xcodebuild` to give us a proper data. In that case, instead of panicking, we can bail out with a warning, and let the command proceed, as it's possible to do that without associating dSYMs just fine - https://github.com/getsentry/sentry-cli/blob/765f26ac55987c6ceeeae5fdc2e6befbc2abc7ea/src/commands/upload_dif.rs#L302-L329

Fixes https://github.com/getsentry/sentry-cli/issues/1126

### DIFF
--- a/src/utils/xcode.rs
+++ b/src/utils/xcode.rs
@@ -101,10 +101,10 @@ fn get_xcode_project_info(path: &Path) -> Result<Option<XcodeProjectInfo>, Error
     }
 
     if projects.len() == 1 {
-        return match XcodeProjectInfo::from_path(&projects[0]) {
+        match XcodeProjectInfo::from_path(&projects[0]) {
             Ok(info) => Ok(Some(info)),
             _ => Ok(None),
-        };
+        }
     } else {
         Ok(None)
     }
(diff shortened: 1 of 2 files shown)


src/utils/xcode.rs
```diff
@@ -115,12 +122,17 @@ impl XcodeProjectInfo {
...
-        rv.project.path = path.as_ref().canonicalize()?;
-        Ok(rv.project)
+
+        match serde_json::from_slice::<Output>(&p.stdout) {
+            Ok(mut rv) => {
+                rv.project.path = path.as_ref().canonicalize()?;
+                Ok(rv.project)
+            }
+            Err(e) => {
+                warn!("Command `xcodebuild -list -json -project {}` failed to produce a valid JSON output. Your .xcodeproj might be malformed.", path.as_ref().display());
```
philipphofmann, line 132: I would start with what is wrong and then go more into detail.
```suggestion
                warn!("Your .xcodeproj might be malformed. Command `xcodebuild -list -json -project {}` failed to produce a valid JSON output.", path.as_ref().display());
```

philipphofmann (commented):
From an Apple perspective, this change makes sense to me. What is the incentive for you to do this change? Did it come up somewhere? 
The only thing to add for me for the Rust perspective is that I don't see a test for this? Can we add one?

kamilogorek:
See https://github.com/getsentry/sentry-cli/issues/1126 for the original cause of this change. As for tests, we don't have any setup tbh. But well, it's Rust, so as long as it compiles, it will be fine now.

philipphofmann (approved):
LGTM from the Apple side, but I would get someone with Rust knowledge to look at the code as well.

PR #1141, opened 2022-03-10
ci: Change stale GitHub workflow to run once a day

kamilogorek:
Running it 32 times a day is a waste of resources and unnecessarily pollutes the actions list.


AbhiPrasad (approved):
🚀

Issue #1142, opened 2022-03-12
Support multiple custom headers


kamilogorek:
We currently support only a single value.

ref: https://github.com/getsentry/sentry-cli/issues/308#issuecomment-1065520776

kamilogorek:
Unfortunatelly it cannot be done in v1 due to clash with `sentry-cli releases files upload --header`, but `--headers` will be renamed to `--file-header` in v2:

- https://github.com/getsentry/sentry-cli/commit/ef7e0e0089734c64944d9981f0d4f70abd50b415
- https://github.com/getsentry/sentry-cli/commit/65a2f68f3456bffc39be918c7f19115e4852173d

kamilogorek:
`2.0.0-beta.0` has been released 🎉

PR #970, opened 2021-06-02
support custom request header from env / config

bkotsopoulossc:
closes: https://github.com/getsentry/sentry-cli/issues/308

As discussed in the issue above, there are use cases where being able to attach a custom header to every request is useful.

This introduces support for a new env var `CUSTOM_HEADER`, as well as a config block under defaults called `custom_header`, that allows you to specify a string that is added directly as a request header to all outgoing requests. Note that this is not a key/value pair - this expects the format `key: value` as a string.

**Testing:**

Uploading a lib somewhere where we don't have permission:

```
$ /docker-entrypoint.sh --log-level DEBUG upload-dif -o 31 -p my_proj /usr/lib/libssl.so.1.1
...
  DEBUG   2021-06-02 18:17:31.558003600 +00:00 > GET /api/0/organizations/31/chunk-upload/ HTTP/1.1
  DEBUG   2021-06-02 18:17:31.558079800 +00:00 > Host: sentry.io
  DEBUG   2021-06-02 18:17:31.558097 +00:00 > Accept: */*
  DEBUG   2021-06-02 18:17:31.558111100 +00:00 > Connection: TE
  DEBUG   2021-06-02 18:17:31.558121500 +00:00 > TE: gzip
  DEBUG   2021-06-02 18:17:31.558214100 +00:00 > User-Agent: sentry-cli/1.65.0
...
```

Setting the env var:
```
$ CUSTOM_HEADER="x-request-id: 12345" /docker-entrypoint.sh --log-level DEBUG upload-dif -o 31 -p my_proj /usr/lib/libssl.so.1.1
...
  DEBUG   2021-06-02 18:17:53.609051800 +00:00 > GET /api/0/organizations/31/chunk-upload/ HTTP/1.1
  DEBUG   2021-06-02 18:17:53.609131500 +00:00 > Host: sentry.io
  DEBUG   2021-06-02 18:17:53.609157300 +00:00 > Accept: */*
  DEBUG   2021-06-02 18:17:53.609178300 +00:00 > Connection: TE
  DEBUG   2021-06-02 18:17:53.609194300 +00:00 > TE: gzip
  DEBUG   2021-06-02 18:17:53.609216900 +00:00 > User-Agent: sentry-cli/1.65.0
  DEBUG   2021-06-02 18:17:53.609239700 +00:00 > x-request-id: 12345
...
```

Setting the config:
```
$ printf "[defaults]\ncustom_header = x-request-id: 987\n" > ~/.sentryclirc
[...]

PR #1146, opened 2022-03-15
feat: Allow for specifying global --header options to config requests

kamilogorek:
Back in the https://github.com/getsentry/sentry-cli/pull/970 `CUSTOM_HEADER` env variable and `defaults.custom_header` config option were introduced. This however was limited to a single header being passed.

In this PR, I add `--header/-h` flag as a global option that can be provided multiple times, and will be propagated down to every request.

This still plays nicely with the old solution and fallbacks to their values if configured.

Note: In v2, I'll change `CUSTOM_HEADER` to `SENTRY_HEADER` and `defaults.custom_header` to `default.header` to make it more unified.

Closes https://github.com/getsentry/sentry-cli/issues/1142

cc @mathew-cf

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -420,9 +420,9 @@ impl Api {
         handle.transfer_encoding(self.config.allow_transfer_encoding())?;
 
         let env = self.config.get_pipeline_env();
-        let custom_header = self.config.get_custom_header();
+        let headers = self.config.get_headers();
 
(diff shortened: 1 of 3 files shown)


kamilogorek:
This will clash with `sentry-cli releases files upload --header` flag, so this change won't make it to v1.

PR #1147, opened 2022-03-15
ref: Rework find_matching_rev tests to not use our real repo history

kamilogorek:
This change allows us to remove `fetch-depth` from the CI, as the tests are now fully isolated.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -587,10 +587,22 @@ use {
 
 #[test]
 fn test_find_matching_rev_with_lightweight_tag() {
+    let dir = tempdir().expect("Failed to generate temp dir.");
+    git_initialize_repo(dir.path());
+
+    git_create_commit(
+        dir.path(),
+        "foo.js",
+        b"console.log(\"Hello, world!\");",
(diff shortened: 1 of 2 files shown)


src/utils/vcs.rs
```diff
@@ -587,10 +587,22 @@ use {
 
 #[test]
 fn test_find_matching_rev_with_lightweight_tag() {
+    let dir = tempdir().expect("Failed to generate temp dir.");
+    git_initialize_repo(dir.path());
+
+    git_create_commit(
+        dir.path(),
+        "foo.js",
+        b"console.log(\"Hello, world!\");",
```
NisanthanNanthakumar, line 596: @kamilogorek why is this a bytestring?

NisanthanNanthakumar: oh I saw further down that it's just any content.

NisanthanNanthakumar (approved):

src/utils/vcs.rs
kamilogorek: I copied it from your tests 😄

PR #1151, opened 2022-03-17
ref: Dont run install script using node binary

kamilogorek:
Instead of relying on some random `node`, which is decided by the `npm` on runtime (eg. `npm` adds everything from `node_modules/.bin` to `$PATH` for its lifecycle-scripts), let `#!/usr/bin/env node` from scripts itself decide. `./install/script.js` has all `-x` chmod flags set correctly, so it won't be a problem executing it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "sentry-cli": "bin/sentry-cli"
   },
   "scripts": {
-    "install": "node scripts/install.js",
+    "install": "./scripts/install.js",
     "fix": "npm-run-all fix:eslint fix:prettier",
     "fix:eslint": "eslint --fix bin/* scripts/**/*.js js/**/*.js",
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):
Good change. We should enforce this in the SDKs as well.

victoriaNine:
I'm having the same issue on Windows, it can no longer be installed.

> error command C:\WINDOWS\system32\cmd.exe /d /s /c ./scripts/install.js
error '.' is not recognized as an internal or external command

kamilogorek:
Oh, excuse me. Fixing now (although gh has some issues so might not be able to release atm)

kamilogorek:
@beheh @victoriaNine released patch in `1.74.2`

victoriaNine:
Perfect, working again. Thanks for the quick fix 🙌

Issue #179, opened 2017-12-08
Make organization slug and other necessary options position independent


john1jan:
I am creating a release for my project using release new command

`sentry-cli releases new 1.5`

But getting this error

error: An organization slug is required (provide with --org)

Where can I set organization slug. If I use **--org** it 

**sentry-cli releases new 1.5 --org ORGANIZATION**
```
`error: Found argument '--org' which wasn't expected, or isn't valid in this context`

USAGE:
    sentry-cli releases new [OPTIONS] <VERSION>
```

jan-auer:
Hi @john1jan, this is slightly confusing but you currently have to put the `--org` parameter at a specific position: 

```
sentry-cli releases --org <org> new --project <proj> <name>
```

Of course it should not matter where to put the options, so let's leave this open until we fix it.

PatNeedham:
@jan-auer where does the `org` field appear on the Sentry dashboard for an existing project? When logged in and viewing a specific project at url https://sentry.io/$1/$2, does `$1` map to org slug and `$2` map to project slug?

jan-auer:
@PatNeedham correct, the org and project slugs appear in the URL: `https://sentry.io/org-slug/project-slug/`. Besides that, we currently show the org slug in settings at the top.

oliverbenns:
Sorry but this is pretty poor design. Now I'm trying all sorts of combinations to get the upload-sourcemaps script to work which does not follow the above ordering.
```
sentry-cli releases --org "org_name" files --project "proj_name" "release_name" upload-sourcemaps ./sourcemaps
```

AlirezaNoorali:
I have the same problem. assume that my project name is **myapp** and my org name is **myorg** now how should I send the command? what is releases ?

mir3z:
This is very unintuitive and documentation does not provie a single example of correct usage of these options in cli.

tonyo:
As an alternative, there's an option of using environment variables, e.g.:
`SENTRY_ORG=org-name SENTRY_PROJECT=project-name sentry-cli releases new 1.0`
This might be more intuitive at the moment.

XyrisKennEnrian:
This critical part of syntax example is not present in the documentation ("org_slug" & "proj_slug", and the project URL doesn't match what is described here. Left to guess at the author's intention.

silavsale:
> @jan-auer & @PatNeedham & @john1jan :
[...]

PR #1152, opened 2022-03-17
ref: Make org and project flags position independent

kamilogorek:
`.global` is only propagating down, not up the command structure, so it will work for us perfectly.

Fixes https://github.com/getsentry/sentry-cli/issues/179


AbhiPrasad (approved):

kamilogorek:
> Argument long must be unique

Ayayaya. In Clap v3 it will be overwritten. I pulled this change from my v2 branch, but it looks like might not work out.

kamilogorek:
This will clash with `sentry-cli releases new --project` flag, so this change won't make it to v1.

Issue #836, opened 2020-09-28
WARN section out of bounds (from goblin)


Paul-van-Klaveren:
[sentry-cli 1.58.0 via Homebrew]
```
$ sentry-cli --auth-token *** upload-dif --org *** --project *** appDsyms
> Found 45 debug information files
  WARN    2020-09-28 11:35:15.178463 +02:00 section #9 size 197146 out of bounds  (from goblin)
  WARN    2020-09-28 11:35:15.335268 +02:00 section #1 size 677524 out of bounds  (from goblin)
> Prepared debug information files for upload
> Uploaded 45 missing debug information files
> File upload complete:

PENDING …
…
```
"  WARN   " _in black with a red background_ sounds ominous, yet the scripts finishes saying it uploaded the files. As a user, should I be concerned with the warning?

jan-auer:
You do not need to be concerned about this warning at this place. 

@kamilogorek can you look into customizing log levels for some of our downstream crates like we do [in Relay](https://github.com/getsentry/relay/blob/b516a63b733a8c249baba85e853c837bcef3fc32/relay/src/setup.rs#L73-L76)? I think we may need to suppress everything that is not an error from goblin, as we're handling these things in `symbolic` already.

kamilogorek:
@jan-auer will try to find some spare time to tackle this

glorysaber:
+1, I get so many it buries every other message.

kamilogorek:
Should be fixed in `1.74.3`

PR #1169, opened 2022-03-30
ref: Filter non-error goblin logs

kamilogorek:
Fixes https://github.com/getsentry/sentry-cli/issues/836

Unfortunately, we couldn't go relay-log route of using `RUST_LOG` env var mentioned by @jan-auer in the original issue, as we are using a custom logger and not `env_logger` crate.


src/utils/logging.rs
```diff
@@ -82,6 +86,18 @@ impl log::Log for Logger {
     fn flush(&self) {}
 }
 
+fn should_skip_log(record: &log::Record) -> bool {
+    let level = record.metadata().level();
+    let target = record.target();
+
+    // We only want errors from goblin
```
AbhiPrasad, line 93: Can we leave a quick reason why for future readers in the comment?

AbhiPrasad (approved):
Makes sense 👍

PR #1172, opened 2022-03-31
sentry-cli v2 RC [rust-part]

kamilogorek:
Minor and non-important changes were removed from the list below.

Related JavaScript code: https://github.com/getsentry/sentry-cli/pull/1173

Features:
- Implement global `--quiet`/`--silent` with appropriate helpers
- Allow for specifying global `--header` options to config requests

Improvements:
- Update to clap v3
- Migrate error handling from failure to anyhow
- Unify env setup and logging level handling in config _(breaking - all output goes to stderr now)_
- Unify progressbar-related code _(breaking - all output goes to stderr now)_
- Make `--auth-token` a global flag
- Make `ignore-empty` for `set-commits` a default behavior
- Rename `CUSTOM_HEADER` to `SENTRY_HEADER` and `defaults.custom_headers` to `http.header` _(breaking)_

Deprecations:
- Remove deprecated and hidden flags from commands
- Remove `difutil id` subcommand
- Remove `react-native codepush` subcommand
- Remove `react-native-gradle` and `react-native-xcode` commands
- Remove `upload-dsym` command
- Remove `monitors` command
- Remove `bash-hook` command in favor of future https://github.com/getsentry/sentry-cli-bash-hook
- Remove crash reporting related code (as it was never used)


src/api.rs
```diff
@@ -211,53 +210,42 @@ impl fmt::Display for SentryError {
     }
 }
 
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("project was renamed to '{0}'\n\nPlease use this slug in your .sentryclirc or sentry.properties or for the --project parameter")]
```
AbhiPrasad, line 214: Why the double newlines `\n\n`?

```suggestion
#[error("project was renamed to '{0}'\n\nPlease use this slug in your .sentryclirc file, sentry.properties file or in the CLI --project parameter")]
```

src/commands/difutil.rs
```diff
@@ -8,27 +8,27 @@ macro_rules! each_subcommand {
         $mac!(difutil_bundle_sources);
         $mac!(difutil_find);
         $mac!(difutil_check);
-        $mac!(difutil_id);
     };
 }
 
-pub fn make_app<'a, 'b: 'a>(mut app: App<'a, 'b>) -> App<'a, 'b> {
+pub fn make_app(mut app: Command) -> Command {
```
AbhiPrasad, line 14: should we rename all references from `app` -> `command` to match the new clap API? We can do that in a follow up PR as well. So `make_app` -> `make_command`

src/utils/logging.rs
```diff
@@ -1,30 +1,66 @@
...
+lazy_static! {
+    static ref QUIET_MODE: AtomicBool = AtomicBool::new(false);
+}
+
+pub fn quiet_mode() -> bool {
+    QUIET_MODE.load(Ordering::Relaxed)
+}
+
+pub fn set_quiet_mode(yes: bool) {
+    QUIET_MODE.store(yes, Ordering::Relaxed);
```
AbhiPrasad, line 23: nit:

```suggestion
pub fn set_quiet_mode(is_quiet: bool) {
    QUIET_MODE.store(is_quiet, Ordering::Relaxed);
```

src/utils/logging.rs
```diff
@@ -1,30 +1,66 @@
...
 use lazy_static::lazy_static;
+use log::max_level;
 use parking_lot::RwLock;
 
+// Quiet mode decides whether to print something based on `--quiet` flag.
+lazy_static! {
+    static ref QUIET_MODE: AtomicBool = AtomicBool::new(false);
+}
+
+pub fn quiet_mode() -> bool {
```
AbhiPrasad, line 18:
```suggestion
pub fn is_quiet_mode() -> bool {
```

src/utils/logging.rs
```diff
@@ -1,30 +1,66 @@
...
+pub fn set_quiet_mode(yes: bool) {
+    QUIET_MODE.store(yes, Ordering::Relaxed);
+}
+
+// NOTE: Remove `allow`s after first use.
+#[allow(unused_macros)]
+macro_rules! quiet_println {
+    ($($tt:tt)*) => {{
+        if !crate::utils::logging::quiet_mode() {
+            println!($($tt)*);
```
AbhiPrasad, line 31: is this macro to prevent allocations? If it is - super clever!!

src/commands/info.rs
```diff
@@ -82,7 +78,7 @@ pub fn execute(matches: &ArgMatches<'_>) -> Result<(), Error> {
     let errors =
         project.is_none() || org.is_none() || config.get_auth().is_none() || info_rv.is_err();
 
-    if !matches.is_present("quiet") {
+    if !quiet_mode() {
```
AbhiPrasad, line 81: Why do we have quiet mode and the debug logger? Shouldn't just debug logging with different levels be enough here? So then we just have a single option to configure?

js/releases/index.js
```diff
@@ -95,10 +95,6 @@ class Releases {
       commitFlags.push('--ignore-missing');
     }
 
-    if (options.ignoreEmpty) {
-      commitFlags.push('--ignore-empty');
```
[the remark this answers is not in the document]
kamilogorek: No, it's fine as it is. We log appropriate messages for it, and I don't believe there is a valid use case for ever breaking a build due to this cli call. Situation is different with `ignore-missing`, as there you actually need to specify what commits you wanted to assign, and if they are not found, there is indeed something wrong.

src/api.rs
kamilogorek: It used to be 2 `\n\n` and I didn't want to change it unnecessarily. But now that I'll use your improved wording, I can bump it too.

src/utils/logging.rs
kamilogorek: I took this convenvtion from clap, but I agree it's better with `is_quiet`

src/utils/logging.rs
kamilogorek: I don't believe it prevents it, as quiet_mode is decided during runtime.
It's so that I can put `quiet_println!` in place of `println!` in most places. (it will be runner up PR)

src/commands/info.rs
kamilogorek: We had a discussion about it on slack. Basically, all logging (including progress bars) go to `stderr`, where CLI output goes to `stdout`. The logging level only controls the verbosity of the logging, but it will not prevent anything from being printed to `stdout`. This way we can easily pipe the output to other places, while still allowing some commands like `sentry-cli info`, `sentry-cli releases info <name>` to have useful exit codes, and make them work as a conditional guard in shell scripts, without polluting the `stdout`.

kamilogorek:
> Looks good! Would like it if we removed the as u64 casting for the progress bar, and I'm sure there are allocations we can make more efficient - but all nits.

`indicatiff` expects `u64` and we cannot do anything about it, but I at least unified where we do the cast.

src/utils/logging.rs
AbhiPrasad: Cool, sounds good!

AbhiPrasad (approved):
think we are good to go!

PR #1173, opened 2022-03-31
sentry-cli v2 RC [js-part]

kamilogorek:
Minor and non-important changes were removed from the list below.

Related Rust code: https://github.com/getsentry/sentry-cli/pull/1172

Improvements:
- Remove `SENTRY_NO_PROGRESS_BAR` flag in favor of `SENTRYCLI_NO_PROGRESS_BAR` _(breaking)_
- Drop support for Node v10 and older and modernize JS codebase _(breaking)_

### DIFF
--- a/js/releases/index.js
+++ b/js/releases/index.js
@@ -50,7 +50,7 @@ class Releases {
    * @returns {Promise} A promise that resolves when the release has been created.
    * @memberof SentryReleases
    */
-  new(release, options) {
+  async new(release, options) {
     const args = ['releases', 'new', release].concat(helper.getProjectFlagsFromOptions(options));
     return this.execute(args, null);
   }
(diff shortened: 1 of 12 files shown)


js/releases/index.js
```diff
@@ -50,7 +50,7 @@ class Releases {
    * @returns {Promise} A promise that resolves when the release has been created.
    * @memberof SentryReleases
    */
-  new(release, options) {
+  async new(release, options) {
```
AbhiPrasad, line 53: why are we adding `async` to this function? `this.execute` should just return a promise directly right, why do we need to wrap it in another promise?

AbhiPrasad (approved):

js/releases/index.js
AbhiPrasad: ahh TIL that makes sense

PR #1176, opened 2022-04-04
2.x prerelease fixes

kamilogorek:
Bunch of fixes prior to beta release.


src/commands/react_native_xcode.rs
```diff
@@ -80,7 +80,9 @@ pub fn make_command(command: Command) -> Command {
         .arg(
             Arg::new("args")
                 .value_name("ARGS")
-                .multiple_occurrences(true)
+                .takes_value(true)
+                .multiple_values(true)
+                .index(2)
```
AbhiPrasad, line 85: should we link to the clap issue report here?

src/utils/system.rs
```diff
@@ -121,7 +121,11 @@ pub fn print_error(err: &Error) {
     if let Some(clap_err) = err.downcast_ref::<clap::Error>() {
         clap_err.exit();
     }
-    eprintln!("{:?}", err);
+
+    eprintln!("{} {}", style("error:").red(), err);
+    err.chain()
+        .skip(1)
+        .for_each(|cause| eprintln!("  {} {}", style("caused by:").dim(), cause));
```
AbhiPrasad, line 128: Should we extract this into a helper? And use whenever we want to log an error out?

src/commands/info.rs
```diff
@@ -80,34 +80,40 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-                    }
+    if is_quiet_mode() {
+        return if errors {
+            Err(QuietExit(1).into())
+        } else {
+            Ok(())
+        };
+    }
+
+    println!("Sentry Server: {}", config.get_base_url().unwrap_or("-"));
```
AbhiPrasad, line 91: If we use `quiet_println` we can avoid the early return above right?

AbhiPrasad (approved):

src/utils/system.rs
kamilogorek: It's only intended for `anyhow` errors, which will always bubble up to this place

src/commands/info.rs
kamilogorek: We can, but I'm not sure if using `quiet_println` in such huge blocks is any more readable 👀

AbhiPrasad: yeah that's fair, let's leave it as is!

PR #1178, opened 2022-04-06
[v2] Rework upload-dif, difutil and releases commands

kamilogorek:
All introduced changes are backward compatible through hidden aliases, so there are no breaking changes in this beta release.

- feat: Change `difutil` command into `debug-files` with `dif/difutil` aliases
- feat: Extract `releases deploys` subcommand to a separate `deploys` command
- feat: Extract `releases files` subcommand to a separate `files` command
- feat: Extract `releases upload-sourcemaps` to a separate `sourcemaps` command
- feat: Integrate `upload-dif` with `debug-files upload` command
- ref: Remove `ReleaseContext` in favor of being explicit


src/commands/deploys.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            deploy.finished = Some(Utc::now());
+        }
+        if let Some(started_str) = matches.value_of("started") {
+            deploy.started = Some(get_timestamp(started_str)?);
+        }
+    }
+
+    let org = config.get_org(matches)?;
+    let deploy = api.create_deploy(&org, &version, &deploy)?;
```
AbhiPrasad, line 109: nit: It's confusing to me to redeclare the variable here, can we make the difference more clear?

src/commands/files_delete.rs
```diff
@@ -0,0 +1,55 @@
...
+pub fn execute(matches: &ArgMatches) -> Result<()> {
+    let config = Config::current();
+    let release = config.get_release_with_legacy_fallback(matches)?;
+    let org = config.get_org(matches)?;
+    let project = config.get_project(matches).ok();
+    let api = Api::current();
+
+    if matches.is_present("all") {
+        if api.delete_release_files(&org, project.as_deref(), &release)? {
+            println!("All files deleted.");
```
AbhiPrasad, line 37: shouldn't we account for the error case when we don't delete all the files? Same with similar functions.

src/commands/files_upload.rs
```diff
@@ -0,0 +1,197 @@
...
+    let mut headers = vec![];
+    if let Some(header_list) = matches.values_of("file-headers") {
+        for header in header_list {
+            if !header.contains(':') {
+                bail!("Invalid header. Needs to be in key:value format");
+            }
+            let mut iter = header.splitn(2, ':');
+            let key = iter.next().unwrap();
+            let value = iter.next().unwrap();
+            headers.push((key.trim().to_string(), value.trim().to_string()));
```
AbhiPrasad, line 112: why push onto headers in loop vs. using `.map().collect()`?

src/config.rs
```diff
@@ -285,16 +285,44 @@ impl Config {
             .ok_or_else(|| format_err!("An organization slug is required (provide with --org)"))
     }
 
+    /// Given a match object from clap, this returns the release from it.
+    pub fn get_release(&self, matches: &ArgMatches) -> Result<String> {
+        matches
+            .value_of("release")
+            .map(str::to_owned)
+            .or_else(|| env::var("SENTRY_RELEASE").ok())
```
AbhiPrasad, line 293: nit: why do we convert to `Option` with `ok()`? Can't we just call `or_else` on the `Result` directly?

AbhiPrasad (approved):
nice, looks solid!

src/commands/debug_files_upload.rs
```diff
@@ -0,0 +1,323 @@
...
+                .value_name("TYPE")
+                .multiple_occurrences(true)
+                .possible_values(&[
+                    "dsym",
+                    "elf",
+                    "breakpad",
+                    "pdb",
+                    "pe",
+                    "sourcebundle",
+                    "bcsymbolmap",
```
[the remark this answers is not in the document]
kamilogorek: Unfortunately not in this case, however `other => bail!("Unsupported type: {}", other)` should be sufficient imo.

src/commands/debug_files_upload.rs
```diff
@@ -0,0 +1,323 @@
...
+    let config = Config::current();
+    let (org, project) = config.get_org_and_project(matches)?;
+
+    let ids = matches
+        .values_of("ids")
+        .unwrap_or_default()
+        .filter_map(|s| DebugId::from_str(s).ok());
+
+    info!(
+        "Issuing a command for Organization: {} Project: {}",
```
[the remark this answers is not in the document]
kamilogorek: We can unify this in some form of log in the separate PR. We already do have a debug log for what command was triggered though.

src/commands/deploys.rs
```diff
@@ -0,0 +1,154 @@
...
+                // Backward compatibility with `releases deploys <VERSION>` commands.
+                .arg(Arg::new("version").long("version").hide(true))
+                .arg(
+                    Arg::new("env")
+                        .long("env")
+                        .short('e')
+                        .value_name("ENV")
+                        .required(true)
+                        .help(
+                            "Set the environment for this release.{n}This argument is required.  \
```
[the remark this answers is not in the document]
kamilogorek: This is for clap to automatically align help description spacing:

> Any valid UTF-8 is allowed in the help text. The one exception is when one wishes to include a newline in the help text and have the following text be properly aligned with all the other help text.

src/commands/deploys.rs
```diff
@@ -0,0 +1,154 @@
...
+        .subcommand_required(true)
+        .arg_required_else_help(true)
+        .org_arg()
+        .project_arg(true)
+        .release_arg()
+        .subcommand(
+            Command::new("new")
+                .about("Creates a new release deployment.")
+                // Backward compatibility with `releases deploys <VERSION>` commands.
+                .arg(Arg::new("version").long("version").hide(true))
```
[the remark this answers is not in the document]
kamilogorek: No, we leave all backward compat in v2, but as hidden aliases. It'll make transitions easier.

src/commands/files.rs
```diff
@@ -0,0 +1,57 @@
...
+        $mac!(files_list);
+        $mac!(files_upload);
+    };
+}
+
+pub fn make_command(mut command: Command) -> Command {
+    macro_rules! add_subcommand {
+        ($name:ident) => {{
+            command = command.subcommand(crate::commands::$name::make_command(Command::new(
+                stringify!($name)[6..].replace('_', "-"),
```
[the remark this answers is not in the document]
kamilogorek: It's gone in https://github.com/getsentry/sentry-cli/pull/1179 😄

src/commands/files_delete.rs
kamilogorek: We do account for it, as api returns result, which will throw error with a valid reason thanks to `?` at the end of the call.

src/commands/files_upload.rs
kamilogorek: Mapping over vector of results is imo just annoying and harder to read than simple loop.

src/config.rs
kamilogorek: `map`, and `or_else` return type is `Option`, which forces us to convert.

PR #1179, opened 2022-04-06
[v2] ref: Restructure commands file structure and importing process

kamilogorek:
No functional changes were done. Just internal restructuring.
The PR is based off `commands-rework` branch, thus this is where it points to.

Commands that have no subcommand are a single-file per command.

image


AbhiPrasad (approved):
didn't look too indepth, but I really like the new folder structure! RIP git blame though.

PR #1181, opened 2022-04-08
Add integration tests for basic interactions with most of the commands

kamilogorek:
image


here/api__0.json
```diff
@@ -0,0 +1 @@
+{"version":"0","auth":null,"user":null}
```
AbhiPrasad, line 1: where is this being generated?

src/api.rs
```diff
@@ -1722,7 +1728,13 @@ impl ApiResponse {
     /// non okay response codes into errors.
     pub fn into_result(self) -> ApiResult<Self> {
         if let Some(ref body) = self.body {
-            debug!("body: {}", String::from_utf8_lossy(body));
+            let body = String::from_utf8_lossy(body);
+            debug!("body: {}", body);
+
+            // Internal helper for making it easier to write integration tests.
+            if let Ok(dir) = env::var("SENTRY_DUMP") {
```
AbhiPrasad, line 1735: do we need to document `SENTRY_DUMP`? Also maybe we make the env variable more specific to sentry cli?

src/commands/login.rs
```diff
@@ -37,6 +39,12 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             .unwrap_or("<unknown>")
     );
 
+    // It's not currently possible to easily mock I/O with `trycmd`,
+    // but verifying that `execute` is not panicking, is good enough for now.
+    if env::var("TEST").is_ok() {
```
AbhiPrasad, line 44: `TEST` is also pretty generalized, maybe we choose a more specific env variable name?

We should also debug log so it's clear why we are skipping the rest of `execute`.

This also applies to the rest of the usages of `env::var("TEST")`

src/api.rs
```diff
@@ -1864,6 +1876,26 @@ fn log_headers(is_response: bool, data: &[u8]) {
     }
 }
 
+fn dump_response(mut dir: String, url: &str, body: String) {
```
AbhiPrasad, line 1879: shouldn't this return an error vs. emitting a panic that bubbles up?

tests/integration/debug_files/upload.rs
```diff
@@ -0,0 +1,68 @@
+use crate::integration::{mock_endpoint, register_test, EndpointOptions};
+
+// NOTE: I have no idea why this is timing out on Windows.
+// I verified it manually, and this command works just fine. — Kamil
+#[cfg(not(windows))]
+#[test]
+fn command_debug_files_upload_help() {
+    register_test("debug_files/debug_files-upload-help.trycmd");
```
AbhiPrasad, line 8: windows CI is notoriously slow, esp for I/O operations. We ran into this when trying to run the Node CI in JS on windows: https://github.com/getsentry/sentry-javascript/pull/4616. Maybe we leave a TODO comment on the command as a future performance optimization?

Even opening a GH issue with the list of these disabled tests with `help wanted` might be nice, maybe somebody else can help figure it out for us!

AbhiPrasad (approved):
Nice, good changes!

here/api__0.json
kamilogorek: Whoops. Removed.

src/api.rs
kamilogorek: I'd not do that tbh. It's only a helper for me, to make writing tests for all existing commands easier. I might remove it in the future entirely. Added better comment, renamed it and handled errors.

src/commands/login.rs
kamilogorek: Updated.

tests/integration/debug_files/upload.rs
kamilogorek: Updated

PR #1182, opened 2022-04-11
feat: Add sourcemaps resolve command

kamilogorek:
Based on https://github.com/getsentry/sentry-cli/pull/1181 so it needs to get merged first.

Includes implementation and integration tests. Only the last commit of this PR is relevant.

### DIFF
--- a/src/commands/sourcemaps/resolve.rs
+++ b/src/commands/sourcemaps/resolve.rs
@@ -1,0 +1,129 @@
+use std::fs;
+use std::path::PathBuf;
+
+use anyhow::{format_err, Result};
+use clap::{Arg, ArgMatches, Command};
+use sourcemap::{DecodedMap, Token};
+
+use crate::utils::args::validate_int;
+
+pub fn make_command(command: Command) -> Command {
+    command
+        .about("Resolve sourcemap for a given line/column position.")
+        .arg(
+            Arg::new("path")
+                .value_name("PATH")
+                .help("The sourcemap to resolve."),
+        )
+        .arg(
+            Arg::new("line")
+                .long("line")
+                .short('l')
+                .value_name("LINE")
+                .validator(validate_int)
+                .help("Line number for minified source."),
+        )
+        .arg(
+            Arg::new("column")
+                .long("column")
+                .short('c')
+                .value_name("COLUMN")
+                .validator(validate_int)
+                .help("Column number for minified source."),
+        )
+}
+
+/// Returns the zero indexed position from matches
+fn lookup_pos(matches: &ArgMatches) -> Option<(u32, u32)> {
+    Some((
+        matches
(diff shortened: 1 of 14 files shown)


src/commands/sourcemaps/resolve.rs
```diff
@@ -0,0 +1,107 @@
...
+            Arg::new("column")
+                .long("column")
+                .short('c')
+                .value_name("COLUMN")
+                .validator(validate_int)
+                .help("Column number for minified source."),
+        )
+}
+
+/// Retrurns the zero indexed position from matches
```
AbhiPrasad, line 36:
```suggestion
/// Returns the zero indexed position from matches
```

AbhiPrasad (approved):

src/commands/sourcemaps/resolve.rs
```diff
@@ -0,0 +1,107 @@
...
+    let sourcemap_path = matches
+        .value_of("path")
+        .ok_or_else(|| format_err!("Sourcemap not provided"))?;
+
+    let sm = sourcemap::decode_slice(&fs::read(PathBuf::from(sourcemap_path))?)?;
+
+    let ty = match sm {
+        DecodedMap::Regular(..) => "regular",
+        DecodedMap::Index(..) => "indexed",
+        DecodedMap::Hermes(..) => "hermes",
```
AbhiPrasad, line 91: do we just make returning the type a method on `DecodedMap` so we don't have to come back and update this match statement when we add new one's in the future?

src/commands/sourcemaps/resolve.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+    if let Some(source) = token.get_source() {
+        println!("  source file: {:?}", source);
+    } else {
+        println!("  source file: not found");
+    }
+    println!("  source line: {}", token.get_src_line());
+    println!("  source column: {}", token.get_src_col());
+    println!("  minified line: {}", token.get_dst_line());
+    println!("  minified column: {}", token.get_dst_col());
```
AbhiPrasad, line 62: what's with the spaces at the start of lines here?

src/commands/sourcemaps/resolve.rs
kamilogorek: We'd have to change that directly in `rust-sourcemap`, but it changes so rarely that I'd not bother tbh.

src/commands/sourcemaps/resolve.rs
kamilogorek: It was meant to make the distinction between the lookup input and output more clear:

image

kamilogorek: Also added better indent formatting (code ugly af, but it works :p)

image

PR #1193, opened 2022-04-13
fix: Remove fetch.FetchError usage in favor of catch-all clause

kamilogorek:
We can do that, as according to https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md

> All operational errors other than aborted requests are rejected with a FetchError.

And because we don't have abort logic, we don't need to handle that.


philipphofmann (approved):
LGTM 🚀

Issue #1204, opened 2022-04-22
Install script fails in pnp environment


tobilen:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
node / yarn berry (3.1.1)
@sentry/nextjs 6.19.6

### Steps to Reproduce

1. Create next project
2. Set to latest yarn version (`yarn set version latest`)
3. Add sentry for nextjs (`yarn add @sentry/nextjs`)
4. Add "@sentry/cli" as dependency to "@sentry/nextjs", since its not declared in https://github.com/getsentry/sentry-javascript/blob/master/packages/nextjs/package.json
to do this, go into `yarnrc.yml` and add a block like this:
```yml
packageExtensions:
  "@sentry/nextjs@*":
    dependencies:
      "@sentry/cli": "*"
```
5. run `yarn`

### Expected Result
postinstall script from the @sentry/cli package gets executed successfully

### Actual Result

postinstall script fails:
```log
# This file contains the result of Yarn building a package (@sentry/cli@npm:2.0.3)
# Script name: install

info sentry-cli Downloading from https://downloads.sentry-cdn.com/sentry-cli/2.0.3/sentry-cli-Darwin-universal
Error: Unable to download sentry-cli binary from https://downloads.sentry-cdn.com/sentry-cli/2.0.3/sentry-cli-Darwin-universal.
Error message: Cannot find package 'node-fetch' imported from /Users/tlengsholz/projects/tbml/.yarn/unplugged/@sentry-cli-npm-2.0.3-80d4720ded/node_modules/@sentry/cli/scripts/install.js
Did you mean to import node-fetch-npm-3.2.3-0f62fc8780-6f702b2683.zip/node_modules/node-fetch/src/index.js?
Error code: ERR_MODULE_NOT_FOUND
```

### Additional context
This seems to be caused by this change: https://github.com/getsentry/sentry-cli/commit/1db9023270f1a9ff277cfa259f68306221fded83#diff-cd32fe9aae3dd316d9dc36709ea1259b7c9aa73acc1e60c4d486ec20be7e43f4R14

yarn berry uses the pnp system for dependecy resolution, and passing a string like this will not cause it to find the dependency. I believe refactoring the line like this should solve the problem:
```diff
- const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args));
[...]

PR #1206, opened 2022-04-25
ref: Use node-fetch@v2 for the simplicity sake

kamilogorek:
We don't use anything from v3, to be honest, and v2 will have all the necessary fixes anyway:

> Critical bug fixes will continue to be published for v2. – https://github.com/node-fetch/node-fetch

Fixes https://github.com/getsentry/sentry-cli/issues/1204

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "https-proxy-agent": "^5.0.0",
-    "node-fetch": "^3.2.3",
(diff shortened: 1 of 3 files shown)


AbhiPrasad (approved):

KrzysztofZawisla:
It should be reverted as it will help to omit the punycode deprecation warning

@AbhiPrasad

PR #1213, opened 2022-04-28
feat: Add organizations list command

kamilogorek:
Adds `organization` with the basic `list` command that lists all orgs for the authenticated command with some details about that organizations (id, name, slug, date created, 2fa requirement, early adopter).


tests/integration/_cases/organizations/organizations-no-subcommand.trycmd
```diff
@@ -0,0 +1,25 @@
+```
+$ sentry-cli organizations
+? failed
```
AbhiPrasad, line 3: Not ideal that we just write `failed` - shouldn’t we leave a message about why we have an error (missing required sub command?)

AbhiPrasad (approved):

tests/integration/_cases/organizations/organizations-no-subcommand.trycmd
kamilogorek: We kinda do. It will print the help page with usage example. This is how clap implements it, and we do the same for every command. `failed` here just means non-0 exit code.

PR #1235, opened 2022-05-17
feat: Add "sourcemap explain" command to debug production sourcemaps

kamilogorek:
Most of the work was done around integration tests really. The verification code is rather straightforward.


src/api.rs
```diff
@@ -2599,3 +2698,16 @@ pub struct GitCommit {
     pub message: Option<String>,
     pub id: String,
 }
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct ProcessedEvent<'a> {
```
sl0thentr0py, line 2703: not that familiar with lifetimes in structs yet, why do we need a lifetime here and the `Cow`s below?

src/commands/sourcemaps/explain.rs
```diff
@@ -0,0 +1,404 @@
...
+        if artifacts.is_empty() {
+            error("Release has no artifacts uploaded");
+            tip("https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/#verify-artifacts-are-uploaded");
+            return Err(QuietExit(1).into());
+        }
+        Ok(artifacts)
+    })?
+}
+
+fn find_matching_artifact(artifacts: &[Artifact], abs_path: &str) -> Result<Artifact> {
```
sl0thentr0py, line 116: this is the only function here I found really hard to follow because I don't know what `abs_path` usually looks like and why we add a `~` later. Maybe add a comment with an example?

sl0thentr0py (approved):
just a couple of things I didn't understand but otherwise looks good to my rust noob okboomer eyes ^^

Issue #1250, opened 2022-06-14
`upload-dif` generates a lot of persistent garbage in %TMP% on Windows


jorgenpt:
### Environment

How do you use Sentry? Sentry SaaS

Which SDK and version? Unreal integration, sentry-cli 1.74.3

### Steps to Reproduce

Every build we do

```bat
sentry-cli.exe upload-dif --include-sources --org <org> --project <project> <Project>\Binaries\Win64
sentry-cli.exe upload-dif  --org <org> --project <project> Engine\Binaries\Win64 
```

### Expected Result

`sentry-cli.exe` does not consume a monotonically increasing amount of disk space

### Actual Result

In the week since I configured a custom `%TMP%` env var before we run `sentry-cli.exe`, one of our builders has generated ~25GB of chaff in 5343 files. The files are all named like UUIDv4 files (e.g. `0a35d3da-e4c7-48cb-88c4-54d6cb5be1ae`) without an extension, and range from 700kb to 11mb.

Based on the filenames, I'm guessing these are built using [fs::TempFile](/getsentry/sentry-cli/blob/4ac17f363ab4b8d8140079a7b23767dd0b080555/src/utils/fs.rs#L51), but it does have a `Drop` implementation that seems to try to clean up. 

Our `sentry-cli.exe` invocations do not error out.

kamilogorek:
I tried to reproduce this issue locally on Windows VM, however unfortunately with no luck.
Would you want to try and debug it locally, and see if it will give us some hints?

The only thing we need to update is https://github.com/getsentry/sentry-cli/blob/4ac17f363ab4b8d8140079a7b23767dd0b080555/src/utils/fs.rs#L90-L94

```patch
impl Drop for TempFile {
    fn drop(&mut self) {
-        fs::remove_file(&self.path).ok();
+        if let Err(e) = fs::remove_file(&self.path) {
+            println!("Failed to remove {}: {:?}", &self.path.display(), e)
+        }
    }
}
```

Then recompile and run with `./target/debug/sentry-cli.exe -include-sources --org <org> --project <project> <Project>\Binaries\Win64` or `cargo run upload-dif --include-sources --org <org> --project <project> <Project>\Binaries\Win64`

jorgenpt:
@kamilogorek, seems like it is failing due to access denied on these files -- is something still keeping one open?

```log
[...]

PR #1259, opened 2022-06-24
fix: Make sure that TempFile is dropped on Windows

kamilogorek:
Fixes https://github.com/getsentry/sentry-cli/issues/1250


kamilogorek:
Suprseded by https://github.com/getsentry/sentry-cli/pull/1256 and https://github.com/getsentry/sentry-cli/pull/1263

PR #1262, opened 2022-06-27
fix: Set bash as the interpretter for bash-hook command

kamilogorek:
Shell script produced by `bash-hook` was missing a shebang, which caused it to be executed by `sh` in some environments. And because `sh` has no process substitution (https://www.shellcheck.net/wiki/SC3001) it failed with a syntax error.

Fixes https://github.com/getsentry/sentry-cli/issues/1260

### DIFF
--- a/src/bashsupport.sh
+++ b/src/bashsupport.sh
@@ -1,3 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
 _SENTRY_TRACEBACK_FILE="___SENTRY_TRACEBACK_FILE___"
 _SENTRY_LOG_FILE="___SENTRY_LOG_FILE___"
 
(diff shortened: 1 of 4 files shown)


mitsuhiko (changes requested):
that won't work since you need to source it in bash. It might be better to detect non bash use and fail quickly.

kamilogorek:
@mitsuhiko not sure what's reliable way of detecting used interpreter

armenzg (changes requested):
Checking the value of `$SHELL` is sufficient.

kamilogorek:
How would `$SHELL` tell you that you are not using `sh`? 🤔

kamilogorek:
This wont work, so cleaning it up for now.

PR #1269, opened 2022-07-04
fix: move dist option to SentryCliUploadSourceMapsOptions

ikenfin:
Hello!

It seems that the `dist` is wrongly placed at main settings, while (according to [this sourcecode](https://github.com/getsentry/sentry-cli/blob/master/js/releases/options/uploadSourcemaps.js#L10)) it need to be placed in source maps settings (_passing dist option to new SentryCli(path, opts) doesnt do anything, while passing it to uploadSourceMaps works_).

This PR moves `dist` from `SentryCliOptions` to `SentryCliUploadSourceMapsOptions`

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -39,11 +39,6 @@ declare module '@sentry/cli' {
      */
     vcsRemote?: string;
     /**
-     * Unique identifier for the distribution, used to further segment your release.
-     * Usually your build number.
-     */
-    dist?: string;
-    /**
      * If true, all logs are suppressed.
      */
     silent?: boolean;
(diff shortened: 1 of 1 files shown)


kamilogorek:
Uff, my bad. Thanks for the fix. Funny enough, we had it merged somewhat incorrectly in webpack plugin, so it works there just fine 👀 

https://github.com/getsentry/sentry-webpack-plugin/blob/395f80bf530b37fae1ade9da6edff6b5f467d2d2/index.d.ts#L19

image

image

ikenfin:
Yeah, it seems that in webpack plugin they passing the whole options object to [uploadSourcemap method](https://github.com/getsentry/sentry-webpack-plugin/blob/395f80bf530b37fae1ade9da6edff6b5f467d2d2/src/index.js#L489), while in vite plugin we keep settings splitted for sentry-cli submodules for sourcemaps, deploy and setCommits using sentry-cli types mostly:)

Issue #1286, opened 2022-07-14
Unexpected argument parsing


vancegillies:
### Environment
Any I think
### Steps to Reproduce

1. run command like `sentry-cli --auth-token token react-native appcenter DEPLOYMENT android ./build/CodePush --deployment Staging --dist 7 -org org --project app` Notice the single `-org` that should be `--org` 
2. see this error 
```
> Fetching latest AppCenter deployment info
> Processing react-native AppCenter sourcemaps
error: No such file or directory (os error 2)
```
3. be really confused because other issue mentioning this error are about java args or metro workers
4. pull this repo and change `react-native/appcenter.rs` line 136 and compile
```rs
for path in matches.values_of("paths").unwrap() {
        println!("WTF is this >>>>>>>>>>>>> :{}", path);
        for entry in (fs::read_dir(path)?).flatten() {
```
5. run same command and see
```
> Fetching latest AppCenter deployment info
> Processing react-native AppCenter sourcemaps
WTF is this >>>>>>>>>>>>> :./build/CodePush
WTF is this >>>>>>>>>>>>> :org
error: No such file or directory (os error 2)
```
6. Realise `-org` issue in command
7. feel dumb and cry.

### Expected Result

Would be nice if it gave you the same error as when it finds args it doesn't recognise
```
error: Found argument '--anything' which wasn't expected, or isn't valid in this context
```
or maybe output the path it errored on would have been super helpful

### Actual Result

Gives a file system error because it some how parses the unknown arg as a path
```
> Fetching latest AppCenter deployment info
> Processing react-native AppCenter sourcemaps
error: No such file or directory (os error 2)
```

kamilogorek:
Hey, thanks for the detailed report. Unfortunately we cannot make it as nice as

> error: Found argument '--anything' which wasn't expected, or isn't valid in this context

as this is coming directly from args parsing library. We need to allow values like `-dir` to be handled, as this is valid directory name in most systems.

And because `paths` are multi-value and position-independent arguments, they are not easily distinguishable.
I updated the error message here, which should let people find the problem way faster - https://github.com/getsentry/sentry-cli/pull/1287

Cheers!

vancegillies:
@kamilogorek I'm sure that will help future idiots like myself, cheers!

PR #1287, opened 2022-07-14
ref: Print better error when processing appcenter paths

kamilogorek:
**Before:**

```
> Fetching latest AppCenter deployment info
> Processing react-native AppCenter sourcemaps
error: No such file or directory (os error 2)
```

**After:**

```
> Fetching latest AppCenter deployment info
> Processing react-native AppCenter sourcemaps
error: Failed processing path: "./src/missing_dir"
  caused by: No such file or directory (os error 2)
```

Fixes https://github.com/getsentry/sentry-cli/issues/1286

### DIFF
--- a/src/commands/react_native/appcenter.rs
+++ b/src/commands/react_native/appcenter.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::ffi::OsStr;
 use std::fs;
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::{Arg, ArgMatches, Command};
 use console::style;
 use if_chain::if_chain;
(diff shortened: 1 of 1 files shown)


src/commands/react_native/appcenter.rs
```diff
@@ -134,7 +134,10 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
     let mut processor = SourceMapProcessor::new();
 
     for path in matches.values_of("paths").unwrap() {
-        for entry in (fs::read_dir(path)?).flatten() {
+        let entries = fs::read_dir(path)
+            .map_err(|e| anyhow!(e).context(format!("Failed processing path: \"{}\"", &path)))?;
```
Swatinem, line 138: there is `with_context` for this: https://docs.rs/anyhow/latest/anyhow/trait.Context.html#example

Swatinem (approved):
do we need to error our early? why can’t we just print the error and continue?

kamilogorek:
I didnt want to change the behavior in patch release.

As for `with_context`, I couldn't get it working for `IE::Error` :<

Issue #1304, opened 2022-08-01
Print warnings when skipping files due to settings


vaind:
CLI currently does not print any warning when it skips the files prior to upload.

This makes it hard to notice that some files are not uploaded, e.g. due to the default of `dif.max_item_size`, which the user may not even know about.

_Originally posted by @kamilogorek in https://github.com/getsentry/sentry-unity/issues/920#issuecomment-1201175427_

PR #1305, opened 2022-08-04
ref: Log warning when debug source excedes item size limit

kamilogorek:
This will produce the warning below:

```
WARN    2022-08-04 12:32:50.920489 +02:00 Source exceded maximum item size limit (116303). /Applications/Xcode.app/usr/include/c++/v1/chrono
WARN    2022-08-04 12:32:50.961811 +02:00 Source exceded maximum item size limit (103509). /Applications/Xcode.app/usr/include/c++/v1/atomic
WARN    2022-08-04 12:32:50.980145 +02:00 Source exceded maximum item size limit (157727). /Applications/Xcode.app/usr/include/c++/v1/type_traits
WARN    2022-08-04 12:32:51.003650 +02:00 Source exceded maximum item size limit (169525). /Applications/Xcode.app/usr/include/c++/v1/string
WARN    2022-08-04 12:32:51.056923 +02:00 Source exceded maximum item size limit (112450). /Applications/Xcode.app/usr/include/c++/v1/vector
```

Question is, whether we should leave it as `WARN`, which is on by default, or make it `DEBUG`, which is opt-in.
I'm not sure how common this case is, and don't want the logs to get cluttered too much.

cc @bruno-garcia 

resolves https://github.com/getsentry/sentry-cli/issues/1304
ref https://github.com/getsentry/sentry-unity/issues/920#issuecomment-1201410874

### DIFF
--- a/src/utils/dif_upload.rs
+++ b/src/utils/dif_upload.rs
@@ -1063,17 +1063,25 @@ fn process_symbol_maps<'a>(
 /// Default filter function to skip over bad sources we do not want to include.
 pub fn filter_bad_sources(entry: &FileEntry) -> bool {
     let max_size = Config::current().get_max_dif_item_size();
+    let path = &entry.abs_path_str();
 
     if entry.name_str().ends_with(".pch") {
         // always ignore pch files
-        false
-    } else if let Ok(meta) = fs::metadata(&entry.abs_path_str()) {
+        return false;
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
I’m okay with making this a warning, as it prompts people to fix their config, or makes it very obvious why they don’t get source context for that file

PR #1047, opened 2021-10-18
Retrieve the events list for a project

dcariotti:
This PR adds a new command:
```
Manage events on Sentry.

USAGE:
    sentry-cli events [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help                     Prints help information
        --log-level <LOG_LEVEL>    Set the log output verbosity. [possible values:
                                   trace, debug, info, warn, error]
    -o, --org <ORG>                The organization slug
    -p, --project <PROJECT>        The project slug
    -t, --tags                     Include tags into the list.
    -u, --user                     Include user' info into the list.

SUBCOMMANDS:
    help    Prints this message or the help of the given subcommand(s)
    list    List all events for an project.
```

Using `events -o <org> -p <project> list` it shows all the events associated with that project of that organization. Using `-t` extend the table with the tags, and using `-u` extend the table with the user informations.

PR #1307, opened 2022-08-05
Retrieve the events list for a project for v2

dcariotti:
Sorry to re-open a PR for the same feat, but I can't re-open #1047. This PR adds
```
Manage events on Sentry.

USAGE:
    sentry-cli events [OPTIONS] <SUBCOMMAND>

OPTIONS:
        --auth-token <AUTH_TOKEN>    Use the given Sentry auth token.
    -h, --help                       Print help information
        --header <KEY:VALUE>         Custom headers that should be attached to all requests
                                     in key:value format.
        --limit <LIMIT>              Limit of requests [default: 10]
        --log-level <LOG_LEVEL>      Set the log output verbosity. [possible values: trace, debug,
                                     info, warn, error]
        --max-rows <MAX-ROWS>        Max of rows for a table.
    -o, --org <ORG>                  The organization slug
    -p, --project <PROJECT>          The project slug.
        --quiet                      Do not print any output while preserving correct exit code.
                                     This flag is currently implemented only for selected
                                     subcommands. [aliases: silent]
    -t, --tags                       Include tags into the list.
    -u, --user                       Include user's info into the list.

SUBCOMMANDS:
    help    Print this message or the help of the given subcommand(s)
    list    List all events in your organization.
```

As requested on [this comment](https://github.com/getsentry/sentry-cli/pull/1047#issuecomment-1095021293) I added two flags:
- `limit` is the limit of API requests
- `max_rows` is the limit of events the table will print (default is all)

I'm not really sure about the last flag but I think it is the closets thing to "per-page" as `cursor` moves 100 items at a time.

---

@kamilogorek sorry if I could only work on it now but I was a bit busy :sweat_smile:

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1414,6 +1414,42 @@ impl Api {
         Ok(rv)
     }
 
+    /// List all events associated with an organization and a project
+    pub fn list_organization_project_events(
+        &self,
+        org: &str,
+        project: &str,
+    ) -> ApiResult<Vec<Event>> {
+        let mut rv = vec![];
+        let mut cursor = "".to_string();
+
+        loop {
+            let resp = self.get(&format!(
+                "/projects/{}/{}/events/?cursor={}",
+                PathArg(org),
+                PathArg(project),
+                QueryArg(&cursor)
+            ))?;
+
+            if resp.status() == 404 || (resp.status() == 400 && !cursor.is_empty()) {
+                if rv.is_empty() {
+                    return Err(ApiErrorKind::OrganizationNotFound.into());
+                } else {
+                    break;
+                }
+            }
+            let pagination = resp.pagination();
+            rv.extend(resp.convert::<Vec<Event>>()?.into_iter());
+            if let Some(next) = pagination.into_next_cursor() {
+                cursor = next;
+            } else {
+                break;
+            }
+        }
+
+        Ok(rv)
+    }
+
     /// List all repos associated with an organization
     pub fn list_organization_repos(&self, org: &str) -> ApiResult<Vec<Repo>> {
         let mut rv = vec![];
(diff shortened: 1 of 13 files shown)


src/commands/events/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+pub fn make_command(mut command: Command) -> Command {
+    macro_rules! add_subcommand {
+        ($name:ident) => {{
+            command = command.subcommand(crate::commands::events::$name::make_command(
+                Command::new(stringify!($name).replace('_', "-")),
+            ));
+        }};
+    }
+
+    command = command
```
kamilogorek, line 23: Those flags should belong to `list` command itself.

src/commands/events/list.rs
```diff
@@ -0,0 +1,65 @@
...
+        matches
+            .value_of("max-rows")
+            .unwrap()
+            .parse::<usize>()
+            .unwrap()
+    } else {
+        events.len()
+    };
+
+    for event in &events[..max_rows] {
```
kamilogorek, line 43: This will panic for `max_rows > events.len()`

src/api.rs
```diff
@@ -2307,6 +2351,81 @@ pub struct Project {
     pub team: Option<Team>,
 }
 
+#[derive(Debug, Deserialize)]
```
kamilogorek, line 2354: We already have all those types in sentry-sdk crate.

kamilogorek (commented):
I'll try to patch everything today

kamilogorek:
Fixed everything, applied necessary changes and added basic cli tests. Thanks for your contribution!

src/api.rs
dcariotti: Hi @kamilogorek I found [`Event`](https://docs.rs/sentry/0.27.0/sentry/types/protocol/v7/struct.Event.html) but I don't think you're talking about this one because it is not good for this JSON response. 
For `EventUser` you were right because we can use `sentry::protocol::User` instead of `EventUserOption`.
For `EventTags` I suggest to keep the `Vec<EventTag>` due the JSON serialization. If we use the same way used for sentry-sdk we have to refactor `tags` in `Vec<BTreeMap<String, String>>` (or `HashMap`) but we always have a BTreeMap (or HashMap) with a tuple `("key", "...")` and a `("value", "...")` cause the nature of the response.

kamilogorek (approved):

PR #1325, opened 2022-09-13
feat(profiling): add profiling to debug_files commands

viglia:


kamilogorek:
It looks like you rebased something on top of latest `master` branch and my recent commit is showing up here. Can you fix that please?

.github/workflows/build.yml
```diff
@@ -35,10 +35,20 @@ jobs:
...
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+
 
       - name: Rename Binary
         run: mv target/*/release/sentry-cli sentry-cli-Linux-${{ matrix.arch }}
 
+      - name: Upload Symbols
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        run: sentry-cli-Linux-${{ matrix.arch }} upload-dif sentry-cli-Linux-${{ matrix.arch }} --org <FILL IN ORG> --project <FILL IN PROJECT>
```
kamilogorek, line 47: NOTE: Either `<FILL IN ORG>` and `<FILL IN PROJECT>` or use same as on line 105

Cargo.toml
```diff
@@ -95,3 +96,4 @@ winapi = "0.3.9"
 [profile.release]
 opt-level = 3
 panic = "abort"
+debug = true
```
kamilogorek, line 99: Is this to output debug symbols?

Cargo.toml
```diff
@@ -80,12 +80,13 @@ mac-process-info = "0.2.0"
 osascript = "0.3.0"
 unix-daemonize = "0.1.2"
 
-[target."cfg(unix)"]
+#[target.'cfg(target_family = "unix")']
 
-[target."cfg(unix)".dependencies]
+[target.'cfg(target_family = "unix")'.dependencies]
```
kamilogorek, line 85: Do we need to be that explicit with `target_family` here instead of using just `cfg(unix)`?

Cargo.toml
```diff
@@ -80,12 +80,13 @@ mac-process-info = "0.2.0"
...
 openssl-probe = "0.1.5"
 signal-hook = "0.3.14"
 crossbeam-channel = "0.5.5"
+sentry = { git = "https://github.com/getsentry/sentry-rust", version = "0.27.0", default-features = false, features = ["anyhow", "curl", "profiling", "frame-pointer"] }
```
kamilogorek, line 89: Specifying url should be unnecessary here.

kamilogorek (commented):
As for the tracing code itself, I need to ping @Swatinem as I've never used it in Rust.

src/commands/mod.rs
```diff
@@ -247,7 +248,23 @@ pub fn execute() -> Result<()> {
             .join(" ")
     );
 
-    run_command(&matches)
+    let _guard = sentry::init((
+        env!("SENTRY_DSN").to_string(),
```
kamilogorek, line 252: This should definitely be optional, as people may want to build it without enabling our SDK.

.github/workflows/build.yml
viglia: ops! thanks for noticing!

scripts/build-in-docker.sh
```diff
@@ -15,7 +15,10 @@ DOCKER_RUN_OPTS="
 
 docker run \
   ${DOCKER_RUN_OPTS} \
-  cargo build --release --target=${TARGET} --locked
+  /bin/bash -c "rustup toolchain install nightly;
+  rustup component add rust-src --toolchain nightly;
+  cargo +nightly run -Z build-std --target ${TARGET}; 
+  RUSTFLAGS=\"-C force-frame-pointers=yes\" cargo build --release --target=${TARGET} --locked"
```
Swatinem, line 21: In the CI job you use `override` which means `cargo +nightly` is the same as `nightly`, but I think here the second command is actually using stable Rust vs nightly.
Not quite sure how exactly `build-std` is supposed to work, I would need to look that up.

Swatinem (commented):
I believe this example gives us a nice improvement opportunity for the global sentry API (wrapping a closure in a transaction)

I’m also not sure if some of these commands panic instead of returning an error, @kamilogorek can enlighten us here. I don’t think we currently finish a transaction when a panic occurs, and I’m also not sure that would work correctly the way that transactions can be shared by multiple hubs, or if that would even be desirable from a product point of view.

Cargo.toml
viglia: yes, correct!

later on, though, when I build the binary I strip out the debug information (see the CI job) so it won't affect the final binary size.

Cargo.toml
viglia: it's not "necessary". I simply sticked to the reference [doc](https://doc.rust-lang.org/reference/conditional-compilation.html#target_family).

If you like the _shortcut_ syntax more we can revert that. Personally though I like to be explicit when writing things. It's a bit more verbose but for people reading it, it might be easier to get what's going on.

Cargo.toml
viglia: Without the `url` it's going to pull the latest published `crate` I think.

I'm not sure made it to a release yet, plus I'm about to merge a PR to align to a new standard format for profiles across all the SDKs so that is def. not in any release yet.

We can coordinate this with @Swatinem to make sure we make a release and at that point we can use a specific version instead of pulling from `master` ?

scripts/build-in-docker.sh
viglia: When I ran my experiments (both on Intel Mac and Linux VM) that seemed to work fine.

I'd run the second command without first enabling the nightly toolchain but simply using the stable one.

src/commands/debug_files/bundle_sources.rs
```diff
@@ -67,6 +67,9 @@ fn get_canonical_path<P: AsRef<Path>>(path: P) -> Result<PathBuf> {
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
+    let tx_ctx = sentry::TransactionContext::new("bundle_sources", "create source bundle for debug file");
+    let transaction = sentry::start_transaction(tx_ctx);
```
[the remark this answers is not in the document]
viglia: yes, good point! 👍🏻

src/commands/mod.rs
viglia: yes, good suggestion!

I've made the whole `profiling` integration optional with a specific feature 👍🏻

.github/workflows/ci.yml
```diff
@@ -38,6 +38,8 @@ jobs:
 
       - name: Run Clippy
         uses: actions-rs/cargo@v1
+        env:
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
```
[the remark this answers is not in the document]
viglia: yes, without `SENTRY_DSN` env variable, if you run `cargo clippy` with `--all-features` it should fail with the following error:

```
environment variable `SENTRY_DSN` not defined
```

due to this: `env!("SENTRY_DSN").to_string()`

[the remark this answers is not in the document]
Swatinem: Thats bad, because it means it will fail for *everyone* (who hasn’t set that env var).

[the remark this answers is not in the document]
viglia: only for everyone that wants to compile it with `profiling` feature (which I've added in one of the last commit).

If you don't use the profiling feature then no error at all.

Honestly this is the best trade-off I could find to still have the `DSN` in the binary without exposing it as clear text on Github.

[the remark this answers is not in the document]
Swatinem: DSNs are not sensitive data. They are shipped longside, or embedded in your executable anyway.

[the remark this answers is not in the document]
viglia: ok, so is it safe to hardcode it directly?

@kamilogorek could you confirm I can do that in `sentry-cli`?

Cargo.toml
```diff
@@ -70,6 +70,7 @@ trycmd = "0.13.4"
 default = []
 managed = []
 with_crash_reporting = []
+profiling = ["sentry"]
```
Swatinem, line 73: the SDK is still used for other things in sentry-cli. Only the profiling (and frame pointer) feature needs to be optional in this case.

viglia: I think that `sentry` get imported in line `43` in the "normal" dependency.

This is more for the optional one at line `90` specific to Unix.

I can't use the `"package-name?/feature-name"` syntax to simplify this because as you said `sentry` as a dependency is already used regardless of profiling and I would end up adding profiling even when I should not.

Not sure if there is a better way to handle this case but I'm open for suggestions.

Cargo.toml
kamilogorek: https://github.com/getsentry/sentry-rust/pull/504 is already approved, so lets merge it, release new `sentry-rust` and replace it with regular cargo index as it was previously.

Cargo.toml
kamilogorek: Usually, the rust convention is to use the shorter form, so I'd prefer to leave it as it was

kamilogorek:
> I’m also not sure if some of these commands panic instead of returning an error, @kamilogorek can enlighten us here.

Yes, it can panic, and we have a hook for that in place as well https://github.com/getsentry/sentry-cli/blob/0c4c6de12c26663fcce94422c8ab8113a9779714/src/utils/system.rs#L140-L170

.github/workflows/ci.yml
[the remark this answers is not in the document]
kamilogorek: I'm fine putting DSN here, however, we need to ask around to create a permanent project for that, and not something temporary/just for testing, as it'll be embedded here forever (for that version).

kamilogorek:
I'd revert all the changes from `Wraps command in a closure to make sure the transaction ends regardless of the results` commit tbh, it's very noisy.

Cleaner solution IMO (at least for now), would be to modify `debug_files/mod.rs`, where we actually register and execute those commands. This way commands are self-contained, and profiling is only added as sort of "instrumentation" of those commands.

.github/workflows/ci.yml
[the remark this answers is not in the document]
viglia: @kamilogorek I just had a quick slack chat with Arpad and it seems we're fine passing it as an env instead of hardcoding it in the code.

> we need to ask around to create a permanent project for that

isn't the `sentry-cli` under the `Sentry` organization just that?

Cargo.toml
viglia: 👍🏻

Cargo.toml
viglia: I'll have to hold on for a bit for that.

I can merge that yet as the related [Relay](https://github.com/getsentry/relay/pull/1462), Sentry and Snuba PRs have yet to be merged.

Once those are merged I'll be able to merge the `sentry-rust`, than I can use that as you suggested.

src/commands/mod.rs
```diff
@@ -247,6 +247,19 @@ pub fn execute() -> Result<()> {
             .join(" ")
     );
 
+    #[cfg(feature = "profiling")]
+    let _guard = sentry::init((
+        env!("SENTRY_DSN").to_string(),
+        sentry::ClientOptions {
+            release: sentry::release_name!(),
+            user_agent: crate::constants::USER_AGENT.into(),
+            traces_sample_rate: 0.05,
```
kamilogorek, line 256: I wonder if its worth it to move `traces_sample_rate` and `profiles_sample_rate` to `constants.rs`

src/commands/mod.rs
```diff
@@ -247,6 +247,19 @@ pub fn execute() -> Result<()> {
             .join(" ")
     );
 
+    #[cfg(feature = "profiling")]
+    let _guard = sentry::init((
+        env!("SENTRY_DSN").to_string(),
```
kamilogorek, line 252:
```suggestion
        env!("SENTRY_DSN", "Environment variable `SENTRY_DSN` is required for `profiling` feature").to_string(),
```

kamilogorek: Its imo better than default `error: environment variable SENTRY_DSN not defined`

.github/workflows/ci.yml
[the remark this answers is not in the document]
kamilogorek: It is, just wanted to double-check.

szokeasaurusrex:
@viglia looks like this PR has been open for some time. Are the changes still relevant, and if yes, is the PR ready for review?

viglia:
@szokeasaurusrex, I'm closing it since it's not relevant anymore.

Issue #1306, opened 2022-08-04
`Pagination offset too large (http status: 400)`


nappernick:
### Environment
Github action
platform: linux, architecture: x86_64
node v16.16.0
sentry-cli 2.5.0 (`curl -sL https://sentry.io/get-cli/ | bash`)

How do you use Sentry?
Sentry SaaS (sentry.io)

### Steps to Reproduce
```
SENTRY_RELEASE=$(_______) >> $_______
  
curl -sL https://sentry.io/get-cli/ | bash
sentry-cli releases -o org -p folder new $SENTRY_RELEASE
sentry-cli releases -o org -p folder set-commits --auto $SENTRY_RELEASE
sentry-cli releases -o org -p folder finalize $SENTRY_RELEASE
sentry-cli releases -o org -p folder deploys $SENTRY_RELEASE new -e $SENTRY_ENV
sentry-cli releases -o org -p folder files $SENTRY_RELEASE upload src
sentry-cli releases -o org -p folder files $SENTRY_RELEASE upload-sourcemaps lib --log-level=debug
```

### Expected Result
Successfully upload sourcemaps, so we can deploy to development/production.

### Actual Result
```yml
Finalized release master
Created new deploy _______ for '_______'
> Found 201 release files
> Bundled 201 files for upload
> Uploaded release files to Sentry
> File upload complete (processing pending on server)
> Organization: _______
> Project: _______
> Release: _______
> Dist: None
  DEBUG   2022-08-04 19:56:53.369529646 +00:00 sentry-cli version: 2.5.0, platform: "linux", architecture: "x86_64"
  INFO    2022-08-04 19:56:53.369617948 +00:00 sentry-cli was invoked with the following command line: "sentry-cli" "releases" "-o" "______" "-p" "______" "files" "master" "upload-sourcemaps" "lib" "--ignore-file" ".sentryignore" "--log-level=debug"
> Found 374 release files
> Analyzing 374 sources
> Rewriting sources
> Adding source map references
```
[Full logs](https://github.com/getsentry/sentry-cli/files/9263557/full_debug_logs_sentry_cli.md)
```yml
DEBUG   2022-08-04 19:59:45.964989286 +00:00 request GET https://sentry.io/api/0/projects/______/______/releases/master/files/?cursor=100:200:0
DEBUG   2022-08-04 19:59:45.965000386 +00:00 using token authentication
[...]

PR #1337, opened 2022-09-22
fix: Break out of the loop when we reach cursor limit for list_release_files

kamilogorek:
Workaround for the small regression introduced in https://github.com/getsentry/sentry-cli/pull/1275 that happens in usecases where over 20k of files are uploaded for a single release.

Potential improvement is explained in the comment:
> Instead of breaking out of the loop here when we reach the limit
> (200 pages of 100 items, for a total of 20_000 files), we should send a list of hashes
> to the server, perform an intersection there, and change the upload mechanism
> to leave out only those files that were gave back to us by the server.
> This would also limit number of API requests for deduping from `N=[1:200]` to 1.

Related DIFs functionality that we can reuse here:

1. https://github.com/getsentry/sentry/blob/f9bf16e7d663c780c1e0c0378ae92f2a93aa326b/src/sentry/api/urls.py#L1941-L1945
2. https://github.com/getsentry/sentry/blob/f9bf16e7d663c780c1e0c0378ae92f2a93aa326b/src/sentry/api/endpoints/debug_files.py#L248-L255
3. https://github.com/getsentry/sentry-cli/blob/2cac02b7e3a1139f941c6aa9842ca1119e64bc7d/src/api.rs#L1005-L1030

Fixes https://github.com/getsentry/sentry-cli/issues/1306

### DIFF
--- a/src/utils/xcode.rs
+++ b/src/utils/xcode.rs
@@ -182,12 +182,10 @@ impl XcodeProjectInfo {
     /// Returns the config with a certain name
     pub fn get_configuration(&self, name: &str) -> Option<&str> {
         let name = name.to_lowercase();
-        for cfg in &self.configurations {
-            if cfg.to_lowercase() == name {
-                return Some(cfg);
-            }
-        }
-        None
+        self.configurations
+            .iter()
+            .find(|&cfg| cfg.to_lowercase() == name)
+            .map(|v| v.as_ref())
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/xcode.rs
```diff
@@ -182,12 +182,10 @@ impl XcodeProjectInfo {
     /// Returns the config with a certain name
     pub fn get_configuration(&self, name: &str) -> Option<&str> {
         let name = name.to_lowercase();
-        for cfg in &self.configurations {
-            if cfg.to_lowercase() == name {
-                return Some(cfg);
-            }
-        }
-        None
+        self.configurations
```
kamilogorek, line 185: "Make clippy happy"

Swatinem (approved):

Issue #1276, opened 2022-07-11
Support env. var. that degrades gracefully in case the external world isn't reachable


marandaneto:
Context: https://github.com/getsentry/team-mobile/issues/3
DACI: https://www.notion.so/sentry/Flag-to-disable-asset-uploads-and-release-creation-5aa83935006a4986a290dd1aa442bedb

There are no implementation details or specs, it's just important that one is able to export an env. var. and be able to have a successful response even if sentry.io isn't reachable.

- [ ] Implement env. var. that degrades gracefully.
- [ ] Log out if the env. var. is active, so they don't forget to `unset` once everything is back.
- [ ] Log out if the env. var. isn't exported and let the users know that they can export that env. var. in case of errors.
- [ ] Add to the docs https://docs.sentry.io/product/cli/configuration/

PR #1343, opened 2022-09-29
feat: Add allow failure option

krystofwoldrich:
This enables users having cli as part of their build process (mainly mobile app/sdks) to proceed in the rare times when sentry.io is down.

This is my first rust code, so all comments on how to do it right are appreciated. 

Also I will add a test for this... but first I would like to know if handling the failure like this makes sense.

Closes: https://github.com/getsentry/sentry-cli/issues/1276 

#skip-changelog

Enables:
- https://github.com/getsentry/sentry-react-native/issues/2246

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -5,7 +5,6 @@ use std::process;
 
 use anyhow::{bail, Result};
 use clap::{Arg, ArgMatches, Command};
-use console::style;
 use log::{debug, info, set_logger, set_max_level, LevelFilter};
 
 use crate::api::Api;
@@ -258,18 +257,13 @@ pub fn execute() -> Result<()> {
     let command_result = run_command(&matches);
 
     if Config::current().get_allow_failure(&matches) {
-        match command_result.err() {
-            Some(err) => print_error(&err),
-            None => {
-                eprintln!();
-                eprintln!("{}", style("The allow_failure options is active, but command exited successfully. Consider removing this option.").yellow());
-            }
+        if let Some(err) = command_result.as_ref().err() {
+            println!("Command failed, however, the \"allow-failure\" flag was present. Exiting gracefully.");
+            println!("Original error:");
+            print_error(err);
         }
         Ok(())
     } else {
-        if let Some(_err) = command_result.as_ref().err() {
-            eprintln!("{}", style("The command failed. Consider adding allow_failure option if this blocks your process.").yellow());
-        }
         command_result
     }
 }
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -183,6 +183,7 @@ fn app() -> Command<'static> {
           Arg::new("allow_failure")
               .long("allow_failure")
               .global(true)
+              .hide(true)
               .help("Always return 0 exit code."),
         )
 }
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -181,7 +181,7 @@ fn app() -> Command<'static> {
         )
         .arg(
           Arg::new("allow_failure")
-              .long("allow_failure")
+              .long("allow-failure")
               .global(true)
               .hide(true)
               .help("Always return 0 exit code."),
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -254,17 +254,17 @@ pub fn execute() -> Result<()> {
             .join(" ")
     );
 
-    let command_result = run_command(&matches);
-
-    if Config::current().get_allow_failure(&matches) {
-        if let Some(err) = command_result.as_ref().err() {
-            println!("Command failed, however, the \"allow-failure\" flag was present. Exiting gracefully.");
-            println!("Original error:");
-            print_error(err);
+    match run_command(&matches) {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            if Config::current().get_allow_failure(&matches) {
+                print_error(&e);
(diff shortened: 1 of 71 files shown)


krystofwoldrich:
@kamilogorek @mitsuhiko @marandaneto @lobsterkatie @jan-auer 
I'm tagging all people I've seen had something to do with this feature.

Would this be acceptable solution, or is it too aggressive? (Keep in mind the reason to do this is to allow people to build their mobile apps in cases where command from CLI is failing. So they don't have to remove whole Sentry integration.)

src/commands/mod.rs
```diff
@@ -178,6 +179,12 @@ fn app() -> Command<'static> {
                 .global(true)
                 .help("Do not print any output while preserving correct exit code. This flag is currently implemented only for selected subcommands."),
         )
+        .arg(
+          Arg::new("allow_failure")
```
bruno-garcia, line 183: To keep consistent with the other commands, use - instead of _ to split words:
```suggestion
          Arg::new("allow-failure")
```

bruno-garcia:
my 2 cents: I like it. You'll be noisy/annoying if eveyrthing is working well (exit 0) but the flag is on. Folks should remove it. But if the command fails, we tell them how to unblock themselves.

Could additonally add more context:

kamilogorek:
I'm still opposed to this change. There are multiple ways to achieve this behavior already, and building it into the CLI is not necessary.

As for the PR:
- `allow-failure` instead of `allow_failure` which Bruno already pointed out
- this flag should be mostly used for our needs, not encouraged and effectively hidden with `.hide(true)`
- current behavior is way too noisy, it should not emit anything when the command is successful, otherwise this "optional" failure pass becomes a "you must pass, or I'll yell at you". I'd not emit anything at all in either of the cases
- due to things mentioned above, all integration tests are quite messy now

marandaneto:
> I'm still opposed to this change. There are multiple ways to achieve this behavior already, and building it into the CLI is not necessary.
> 
> As for the PR:
> 
> * `allow-failure` instead of `allow_failure` which Bruno already pointed out
> * this flag should be mostly used for our needs, not encouraged and effectively hidden with `.hide(true)`
> * current behavior is way too noisy, it should not emit anything when the command is successful, otherwise this "optional" failure pass becomes a "you must pass, or I'll yell at you". I'd not emit anything at all in either of the cases
> * due to things mentioned above, all integration tests are quite messy now

There was a [DACI](https://www.notion.so/sentry/Flag-to-disable-asset-uploads-and-release-creation-5aa83935006a4986a290dd1aa442bedb) that we all have agreed on about doing this directly into the CLI.

So I'd just address the code review points related to naming, noisy, etc.

krystofwoldrich:
Okay, I'll adjust it.

I realize it's noisy, if we hide it then not advertising it makes sense, but what about warning about it when you have it set and the command succeeds?

I think that might be useful, the goal is to use this only in an emergency.

marandaneto:
> Okay, I'll adjust it.
> 
> I realize it's noisy, if we hide it then not advertising it makes sense, but what about warning about it when you have it set and the command succeeds?
> 
> I think that might be useful, the goal is to use this only in an emergency.

If the command succeeds, we should not log anything, everything worked as it should.

krystofwoldrich:
@marandaneto  I assumed we didn't want people to set this flag and use it forever.

kamilogorek:
> ... we all have agreed on about doing this directly into the CLI.

Me, from DACI:
> which I don't like tbh, but again, I don't like this feature at all anyway ¯\_(ツ)_/¯

👀

If anything, I'd make it a hidden global flag, that would not be documented in the CLI itself, and would only log if something failed _but_ the flag was set, eg. 

```
Command failed, however, the "allow-failure" flag was present. Exiting gracefully.
Original error:
  {err}
```

vaind:
> here it's generated by the bot

Hmm, I wasn't aware of that. the changelog check is unnecessary in this repo I guess?

krystofwoldrich:
Okay, now it should be better. :) @kamilogorek

src/config.rs
```diff
@@ -441,6 +441,15 @@ impl Config {
             false
         }
     }
+
+    pub fn get_allow_failure(&self, matches: &ArgMatches) -> bool {
+        matches.is_present("allow_failure")
+            || if let Ok(var) = env::var("SENTRY_ALLOW_FAILURE") {
```
marandaneto, line 447: Nice, it's much easier to document the `SENTRY_ALLOW_FAILURE` env. var. for plugins.

marandaneto (approved):
Thanks for doing this @krystofwoldrich 
I will let @kamilogorek do the final approval since I don't know Rust :P

kamilogorek (approved):

kamilogorek:
Updated code, tests and merged.

mattjohnsonpint:
Hey, looks like we could use this documented here also?  https://docs.sentry.io/product/cli/configuration/#configuration-values

kamilogorek:
Tbh I was always against this option in the first place, and not sure if we should document it publicly, as people can easily shoot themselves in the foot with it. It's there to be mostly used by 3rd party wrappers/plugins.

krystofwoldrich:
We use it in RN Plugins and it's documented [here](https://docs.sentry.io/platforms/react-native/troubleshooting/#failing-build-due-to-sentry-cli) for that use-case. As there is no easy way to opt-out of the plugin when something goes wrong.

bruno-garcia:
PR to add to the docs: https://github.com/getsentry/sentry-docs/pull/11318

kamilogorek:
Thanks @bruno-garcia

Issue #1342, opened 2022-09-28
sentry-cli info exits with error code 1 with no visible error


kittenswolf:
### Environment

```
kittenswolf$ ./sentry-cli --version
sentry-cli 2.6.0

kittenswolf$ uname -a
Linux 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
`sentry-cli` is the `sentry-cli-Linux-x86_64` binary from version 2.6.0. Tested in Docker and WSL.

### Steps to Reproduce

1. Create any token (internal integration token or account API token)
2. Call `sentry-cli info` with said token (e.g. `SENTRY_AUTH_TOKEN=REDACTED SENTRY_URL=REDACTED ./sentry-cli info`)
3. Observe exit code (with `echo $?`)

### Expected Result

I expected the exit code to be 0 when no error is visible.

### Actual Result

```
kittenswolf$ SENTRY_AUTH_TOKEN=REDACTED SENTRY_URL=REDACTED ./sentry-cli info
Sentry Server: REDACTED
Default Organization: -
Default Project: -

Authentication Info:
  Method: Auth Token
  User:
  Scopes:
    - project:releases
kittenswolf$ echo $?
1
```

The program exits with an error code of 1, which makes [Drone](https://github.com/harness/drone) think the command has failed. There is no obvious reason why it should exit with an error code.

With `--log-level=debug`:

```

kittenswolf$ SENTRY_AUTH_TOKEN=REDACTED SENTRY_URL=REDACTED ./sentry-cli info --log-level=debug
  DEBUG   2022-09-28 20:01:04.797780474 +02:00 sentry-cli version: 2.6.0, platform: "linux", architecture: "x86_64"
  INFO    2022-09-28 20:01:04.797825884 +02:00 sentry-cli was invoked with the following command line: "./sentry-cli" "info" "--log-level=debug"
  DEBUG   2022-09-28 20:01:04.798220018 +02:00 request GET https://REDACTED/api/0/
  DEBUG   2022-09-28 20:01:04.798251478 +02:00 using token authentication
  DEBUG   2022-09-28 20:01:04.798280438 +02:00 retry number 0, max retries: 0
  DEBUG   2022-09-28 20:01:04.892199548 +02:00 > GET /api/0/ HTTP/1.1
[...]

PR #1344, opened 2022-10-03
feat: Improve info command and introduce no-defaults flag

kamilogorek:
Closes https://github.com/getsentry/sentry-cli/issues/1342


src/commands/info.rs
```diff
@@ -19,14 +18,24 @@ pub struct AuthStatus {
 
 #[derive(Serialize, Default)]
 pub struct ConfigStatus {
```
kamilogorek, line 20: Removed `HashMap` here to preserve reproducible json values order for integration tests.

src/commands/info.rs
```diff
@@ -75,9 +88,15 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let config = Config::current();
     let (org, project) = config.get_org_and_project_defaults();
+    let org = org.filter(|s| !s.is_empty());
```
kamilogorek, line 91: Dont here instead of in `config.rs` to not alter the current `config-status-json` behavior. If I'd done it there, json would return `null` instead of the current `''`.

src/commands/info.rs
```diff
@@ -88,32 +107,33 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+
+    if !matches.is_present("no_defaults") {
+        println!(
+            "Default Organization: {}",
+            org.unwrap_or_else(|| "-".into())
+        );
+        println!("Default Project: {}", project.unwrap_or_else(|| "-".into()));
+    }
+
+    println!();
```
kamilogorek, line 119: Whitespace changes only.

Swatinem (approved):

PR #1345, opened 2022-10-03
feat: Add support for Portable PDB debug files

Swatinem:
This adds support to the `debug-files` subcommand to check, find and upload Portable PDB files.

The actual upload still depends on properly exposing this in sentry.


deauna (approved):
improve

kamilogorek (approved):

PR #1346, opened 2022-10-03
fix: paths argument for debug-files find should take a value

kamilogorek:
NOTE: `takes_name` implicitly sets `Arg::takes_value(true)`

Whoops.

#no-changelog


kamilogorek:
Ugh, I need to disable danger. Changelogs in CLI are collected automatically.

PR #1349, opened 2022-10-05
List issues

boozec:
Hi all :wave:
with this PR I suggest to add the `sentry issues list` command to show the list of issues bases on a query. This is similar to `events list` but showed as below:

```
sentry-cli issues list --query="is:unresolved"
+------------+-----------+------------------------+-----------------------------+------------+-------+
| Issue ID   | Short ID  | Title                  | Last seen                   | Status     | Level |
+------------+-----------+------------------------+-----------------------------+------------+-------+
| 1          | SE-API-1  | Example                | 2022-10-05T00:00.856565Z    | unresolved | error |
+------------+-----------+------------------------+-----------------------------+------------+-------+
```

#skip-changelog


src/api.rs
```diff
@@ -2813,6 +2861,18 @@ pub struct ProcessedEventUser {
     pub ip_address: Option<String>,
 }
 
+#[derive(Clone, Debug, Deserialize)]
```
kamilogorek, line 2864: You should be able to use `#[serde(rename_all = "camelCase")]` instead I believe.

kamilogorek: Can you also move it near `IssueFilter` and `IssueChanges` so it's grouped together?

src/api.rs
```diff
@@ -1572,6 +1572,54 @@ impl Api {
         Ok(rv)
     }
 
+    /// List all issues associated with an organization and a project
+    pub fn list_organization_project_issues(
+        &self,
+        org: &str,
+        project: &str,
+        max_pages: usize,
+        query: String,
```
kamilogorek, line 1581: Query is optional, so it should be `Option<String>`

src/api.rs
```diff
@@ -1572,6 +1572,54 @@ impl Api {
...
+    ) -> ApiResult<Vec<Issue>> {
+        let mut rv = vec![];
+        let mut cursor = "".to_string();
+        let mut requests_no = 0;
+
+        loop {
+            requests_no += 1;
+
+            let resp = self.get(&format!(
```
kamilogorek, line 1590: Add query conditionally based on `if let Some(query) = query { ... }`

src/commands/issues/list.rs
```diff
@@ -0,0 +1,82 @@
...
+                .help("Query to pass at the request. An example is \"is:unresolved\""),
+        )
+}
+
+pub fn execute(matches: &ArgMatches) -> Result<()> {
+    let config = Config::current();
+    let org = config.get_org(matches)?;
+    let project = config.get_project(matches)?;
+    let pages = *matches.get_one("pages").unwrap();
+    let query = match matches.get_one::<String>("query") {
```
kamilogorek, line 40: Once query is changed to `Option`, there'll be no need to match it here.

tests/integration/_cases/issues/issues-help.trycmd
```diff
@@ -0,0 +1,33 @@
...
+
+Commands:
+  list       List all issues in your organization.
+  mute       Bulk mute all selected issues.
+  resolve    Bulk resolve all selected issues.
+  unresolve  Bulk unresolve all selected issues.
+  help       Print this message or the help of the given subcommand(s)
+
+Options:
+  -o, --org <ORG>                The organization slug
```
kamilogorek, line 17: Previously there was no `list` command so it made sense to keep those flags here. Now however, `status`, `id` and `all` should belong to `mute`, `unmute` and `resolve` commands respectively, otherwise this `help` page is not accurate. Unfortunately, moving them from `issues/mod.rs` to those 3 separate commands will break existing behaviour, as `--all` would not work when specified before subcommand, now it does :(

tests/integration/issues/list.rs
```diff
@@ -0,0 +1,19 @@
+use crate::integration::{mock_endpoint, register_test, EndpointOptions};
+
+#[test]
+fn command_issues_list_help() {
+    register_test("issues/issues-list-help.trycmd");
+}
+
+#[test]
```
kamilogorek, line 8: Would it be possible to add some "working" tests as well? Maybe with flags behavior that you added as well? :)

kamilogorek (changes requested):
Excuse me, I missed that PR entirely. Will merge it once all patches are in place. Thanks for the contribution!

src/api.rs
boozec: Why don't we keep the `query: String` and leave the  `QueryArg` below?

src/api.rs
boozec: Great! Thx

tests/integration/_cases/issues/issues-help.trycmd
boozec: I noticed that the CLI already has something similar for `files delete`. But, it this case, `--all` must be defined after the command as shown here:

```
$ ./target/debug/sentry-cli files --all delete
error: unexpected argument '--all' found

  note: argument '--allow-failure' exists

Usage: sentry-cli files [OPTIONS] <COMMAND>

For more information, try '--help'.
```

Do you think it might be OK to break the existing behaviour to keep a consistency with `files` or do you suggest to create separate subcommands where `mute`, `unmute` and `resolve` keep `status`, `id` and `all` args before the command name?

src/api.rs
kamilogorek: It's more about the intent rather than correctness. Whenever someone comes and read this code in the future, they should be able to see from the function definition itself that query is optional and the api will function without it as well.

tests/integration/_cases/issues/issues-help.trycmd
kamilogorek: I don't think there's much we can do here tbh. We cannot break backwards compatibility and it's not possible to add the same flag multiple times (with the current one being hidden), as clap will complain about the override.
We can leave it as is for now.

tests/integration/issues/list.rs
```diff
@@ -17,3 +19,32 @@ fn doesnt_fail_with_empty_response() {
...
+            "GET",
+            "/api/0/projects/wat-org/wat-project/issues/?cursor=",
+            200,
+        )
+        .with_response_file("issues/get-issues.json")
+        .with_matcher(Matcher::PartialJson(json!({
+            "query": "is:resolved",
+        }))),
+    );
+    register_test("issues/issues-display-with-query.trycmd");
```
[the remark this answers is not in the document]
boozec: Okay, but the query is not filtered and I do not know why

```diff
---- expected: stdout
++++ actual:   stdout
        1 +   INFO    2023-07-19 16:05:44.849431768 +02:00 Loaded config from [CWD]/.sentryclirc
        2 +   DEBUG   2023-07-19 16:05:44.849736671 +02:00 sentry-cli version: [VERSION], platform: "linux", architecture: "x86_64"
        3 +   INFO    2023-07-19 16:05:44.849835116 +02:00 sentry-cli was invoked with the following command line: "[CWD]/target/debug/sentry-cli" "issues" "list" "--query" "is:resolved"
        4 +   DEBUG   2023-07-19 16:05:44.850717023 +02:00 request GET http://127.0.0.1:1234/api/0/projects/wat-org/wat-project/issues/?query=is:resolved&cursor=
        5 +   DEBUG   2023-07-19 16:05:44.850790381 +02:00 using token authentication
        6 +   DEBUG   2023-07-19 16:05:44.850849782 +02:00 retry number 0, max retries: 0
        7 +   DEBUG   2023-07-19 16:05:44.859579115 +02:00 > GET /api/0/projects/wat-org/wat-project/issues/?query=is:resolved&cursor= HTTP/1.1
        8 +   DEBUG   2023-07-19 16:05:44.859636333 +02:00 > Host: 127.0.0.1:1234
        9 +   DEBUG   2023-07-19 16:05:44.859652934 +02:00 > Accept: */*
       10 +   DEBUG   2023-07-19 16:05:44.859668343 +02:00 > Connection: TE
       11 +   DEBUG   2023-07-19 16:05:44.859683101 +02:00 > TE: gzip
       12 +   DEBUG   2023-07-19 16:05:44.859699101 +02:00 > User-Agent: sentry-cli/[VERSION]
       13 +   DEBUG   2023-07-19 16:05:44.865290468 +02:00 > Authorization: Bearer lolnope***
       14 +   DEBUG   2023-07-19 16:05:44.865346623 +02:00 < HTTP/1.1 200 OK
       15 +   DEBUG   2023-07-19 16:05:44.865374055 +02:00 < connection: close
       16 +   DEBUG   2023-07-19 16:05:44.865396797 +02:00 < content-type: application/json
       17 +   DEBUG   2023-07-19 16:05:44.865418939 +02:00 < content-length: 620
       18 +   DEBUG   2023-07-19 16:05:44.865521662 +02:00 response status: 200
       19 +   DEBUG   2023-07-19 16:05:44.865587426 +02:00 body: [
...
       44 + ]
       45 +
       46 +   DEBUG   2023-07-19 16:05:44.865816316 +02:00 Response dumped to: dump/api__0__projects__wat-org__wat-project__issues.json
   1   47 | +------------+-----------+---------------------------------------------------------+-----------------------------+------------+-------+
   2   48 | | Issue ID   | Short ID  | Title                                                   | Last seen                   | Status     | Level |
   3   49 | +------------+-----------+---------------------------------------------------------+-----------------------------+------------+-------+
       50 + | 4242424243 | SEN-CLI-L | ProgrammingError: column users_user.role does not exist | 2023-07-18T00:12:01.222387Z | unresolved | error |
   4   51 | | 4242424242 | SEN-CLI-H | N+1 Query                                               | 2023-07-18T00:10:01.222387Z | resolved   | info  |
       52 + | 4242424241 | SEN-CLI-1 | NameError: name 'jobs' is not defined                   | 2023-07-18T00:00:01.222387Z | ignored    | error |
   5   53 | +------------+-----------+---------------------------------------------------------+-----------------------------+------------+-------+
       54 +   DEBUG   2023-07-19 16:05:44.866845690 +02:00 skipping update nagger because session is not attended
```

the endpoint has a query `/api/0/projects/wat-org/wat-project/issues/?query=is:resolved&cursor=` but the output is not filtered. Am I using a wrong issues-file?

tests/integration/_cases/issues/issues-help.trycmd
boozec: Okay :\

src/api.rs
boozec: It makes really sense, thank you for the suggestion :) check it out

tests/integration/issues/list.rs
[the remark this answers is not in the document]
kamilogorek: Well, it's just a mock, so whatever you put in the json will be returned, and you return this exact response.
You can make an assumption that as long as this url with query is hit, it will return correct data, as its backends responsibility to do so. If url wont be hit, test wont pass.

tl;dr - create a new mock with prefiltered data

kamilogorek (approved):

kamilogorek:
Lovely, thanks! :)

Issue #1354, opened 2022-10-12
JS layer for sentry-cli should allow passing multiple headers


thibmeu:
At the moment, JS layer above [sentry-cli](https://www.npmjs.com/package/@sentry/cli) allow passing at most one header, as follow

```javascript
import Sentry from '@sentry/cli'

const cli = new Sentry(undefined, {
  // Note the singular for custom header. there is only one
  customHeader: 'Authorization: Bearer deadbeef',
})
console.log(await cli.info('my-release'))
```
The above configuration makes sentry-cli use the custom header by setting `CUSTOM_HEADER` environment variable when invoking sentry-cli.

sentry-cli supports passing multiple headers via `--header key:value`. There are situation for which setting multiple headers is good, and limiting sentry-cli JS layer makes it less useful.
At the moment, I understand passing multiple headers has to be done as follow
```javascript
import Sentry from '@sentry/cli'

const cli = new Sentry(undefined, {
  // Nothing here
})
console.log(await cli.execute('--header', 'key1:value1', '--header', 'key2:value2', 'info', 'my-release'))
```

I suggest JS layer to define the below interface
```javascript
import Sentry from '@sentry/cli'

const cli = new Sentry(undefined, {
  // headers is a Record<string, string>
  headers: {
    'Authorization': 'Bearer deadbeef',
  },
})
console.log(await cli.info('my-release'))
```

Lekensteyn:
This relates to #1142 which is about adding support for multiple headers to the CLI tool.

thibmeu:
Thanks @kamilogorek for the review and the merge.

Are there information about the release process available? I'd be interested to know when this change would land in the upstream npm package.

kamilogorek:
Will update it early next week, as there are a few things remaining to tackle before that. Will keep you posted.

kamilogorek:
@thibmeu `2.8.0` is out.

thibmeu:
Thanks @kamilogorek. That's was super fast!

PR #1355, opened 2022-10-12
feat: Add `headers` option on sentry cli js interface (#1354)

thibmeu:
Allow multiple headers to be passed to sentry-cli when usingy the JS interface.

Fixes https://github.com/getsentry/sentry-cli/issues/1354

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -47,6 +47,11 @@ declare module '@sentry/cli' {
      * This value will update `CUSTOM_HEADER` env variable.
      */
     customHeader?: string;
+    /**
+     * Headers added to every outgoing network request.
+     * This value does not set any env variable, and is overriden by `customHeader`.
+     */
+    headers?: Record<string, string>;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


js/index.d.ts
```diff
@@ -47,6 +47,11 @@ declare module '@sentry/cli' {
      * This value will update `CUSTOM_HEADER` env variable.
      */
     customHeader?: string;
+    /**
+     * Headers added to every outgoing network request.
+     * This value does not set any env variable, and is overriden by `customHeader`.
```
Lekensteyn, line 52:
```suggestion
     * This value does not set any env variable, and is overridden by `customHeader`.
```

Lekensteyn (approved):
This seems the right option. `CUSTOM_HEADER` is limited to one header only, and the `--header` option was explicitly added to support multiple headers. See #1142 (and your issue #1354).

Should the docs (https://github.com/getsentry/sentry-docs) also be updated?

kamilogorek (approved):

kamilogorek:
That should do the job. Thanks!

PR #1361, opened 2022-10-16
Update Cargo.toml - Tests would not run

apps4u:
could not build WSL 2 Ubuntu, had to add ```yaml``` to ```insta``` package features


kamilogorek:
We are still on older version of insta, which has no `yaml` feature, as it was built in. If you want to update it, please bump `insta` version first. Thanks!

PR #1369, opened 2022-10-23
fix: Windows binary location

### LINKED ISSUE #1368: On Windows, the binary is copied to an incorrect path
With the latest version of the CLI installed, the directory layout looks like below. This means the binary is at `./js/..sentry-cli.exe`:

image

This is no doubt due to this code:
https://github.com/getsentry/sentry-cli/blob/6f9173687fdfd56810145c5c727ff76978784814/js/helper.js#L9-L11

timfish:
It appears that `path.resolve` already caters for path separator differences so you can safely use forwards slashes and everything works.

Closes #1368

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -7,7 +7,7 @@ const childProcess = require('child_process');
  * @type {string}
  */
 let binaryPath = eval(
-  "require('path').resolve(__dirname, require('os').platform() === 'win32' ? '..\\sentry-cli.exe' : '../sentry-cli')"
+  "require('path').resolve(__dirname, require('os').platform() === 'win32' ? '../sentry-cli.exe' : '../sentry-cli')"
 );
 
 /**


kamilogorek:
Ah, nice cache, you are correct:

> On Windows, both the forward slash (/) and backward slash (\) are accepted as path segment separator

https://nodejs.org/api/path.html#pathsep

timfish:
Yes, the cli was working perfectly well on Windows even with the dodgy file name. 

However `@sentry/nextjs` does a check to see if the binary exists (ie. has been successfully downloaded) and this would not have worked on Windows because a) it was not checking for `.exe` b) the executable file was not where it was expected to be!

https://github.com/getsentry/sentry-javascript/blob/4b137b72e256b3773d9a8b500e857ce3ee7208a0/packages/nextjs/src/config/webpack.ts#L459-L473

kamilogorek (approved):

PR #1374, opened 2022-10-25
fix: Replace usage of `eval` to obfuscate binary path from bundlers

timfish:
`eval` is not ideal because it causes Rollup to display warnings. 

As per my [test repo](https://github.com/timfish/vercel-nft-test), it's reasonably easy to obfuscate paths from `@vercel/nft` without using `eval` as the evaluator can only simplify so far.

This PR adds a test to check that the binaries are not picked up by `@vercel/nft`.

For the purposes of demonstrating the newly added test, the first commit replaces the `eval` usage with `path.resolve()` which will cause the test to fail as the binary is picked up. 

A subsequent commit will add the mildly obfuscated path generation.

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -6,16 +6,11 @@ const childProcess = require('child_process');
  * Absolute path to the sentry-cli binary (platform dependent).
  * @type {string}
  */
-let binaryPath = eval(
-  "require('path').resolve(__dirname, require('os').platform() === 'win32' ? '..\\sentry-cli.exe' : '../sentry-cli')"
+let binaryPath = require('path').resolve(
+  __dirname,
+  '..',
+  process.platform === 'win32' ? 'sentry-cli.exe' : 'sentry-cli'
 );
-
-/**
- * NOTE: `eval` usage is a workaround for @vercel/nft detecting the binary itself as the hard dependency
- *       and effectively always including it in the bundle, which is not what we want.
- * ref: https://github.com/getsentry/sentry-javascript/issues/3865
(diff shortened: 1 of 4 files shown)


lforst (approved):
nice 👍

timfish:
I guess if this needs to be released as `1.x` I should have branched off the `1.x` branch!

lforst:
> I guess if this needs to be released as `1.x` I should have branched off the `1.x` branch!

Probably. Is it possible for you to still change the base?

lobsterkatie (approved):
Nice work figuring out a successful obfuscation strategy. @lforst and @Lms24 and I tried like the dickens for like an hour one day and couldn't manage to trick it.

> I guess if this needs to be released as 1.x I should have branched off the 1.x branch!

P.S. Does anyone remember why the webpack plugin is stuck on sentry-cli 1.x? @kamilogorek, maybe? In any case, I think it actually needs to be merged into both 1.x and 2.x, right? Because eventually we'll upgrade, and we'll want to have it there.

timfish:
> Does anyone remember why the webpack plugin is stuck on sentry-cli 1.x?

https://github.com/getsentry/sentry-webpack-plugin/pull/383#issuecomment-1255110206

timfish:
> Nice work figuring out a successful obfuscation strategy

I have "history" with it's predecessor `@vercel/webpack-asset-relocator` so when nft was releases I spent far too long looking through the strategies it uses and experimented with the evaluation it performs and was left suitably impressed!

PR #1375, opened 2022-10-28
feat: Replace usage of `eval` to obfuscate binary path from bundlers [backport to 1.x]

timfish:
This is a backport of #1374 to the 1.x branch.

I had to disable the `@vercel/nft` test on node.js v8 because it doesn't support versions that old

I also allowed a few Rust clippy lint failures which is no doubt caused by this branch being built with a newer version of Rust than was intended. I chose to disable them rather than follow the lint fix suggestions since this branch will never get merged back into master and it didn't seem appropriate to change the code at this point!

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -1,21 +1,29 @@
 'use strict';
 
+const path = require('path');
 const childProcess = require('child_process');
+
+/**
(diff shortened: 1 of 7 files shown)


lforst (approved):

lforst:
Oof didn't squash when merging 😭

PR #1386, opened 2022-11-04
feat(dist): Add limit validation

olksdr:
Make sure we limit the length of the incoming distribution name, which should not be longer than 64 characters, which is enforced on the DB side.

This change just adds the validation as closer to the user as possible.

related: https://github.com/getsentry/relay/pull/1556

### DIFF
--- a/src/utils/args.rs
+++ b/src/utils/args.rs
@@ -56,7 +56,7 @@ pub fn validate_distribution(v: &str) -> Result<(), String> {
             "Invalid distribution name. Distribution must not contain leading or trailing spaces."
                 .to_string(),
         )
-    } else if v.len() > 64 {
+    } else if bytecount::num_chars(v.as_bytes()) > 64 {
         Err(
             "Invalid distribution name. Distribution name must not be longer than 64 characters."
                 .to_string(),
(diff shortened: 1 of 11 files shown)


src/utils/args.rs
```diff
@@ -50,6 +50,22 @@ fn validate_release(v: &str) -> Result<(), String> {
...
+pub fn validate_distribution(v: &str) -> Result<(), String> {
+    if v.trim() != v {
+        Err(
+            "Invalid distribution name. Distribution must not contain leading or trailing spaces."
+                .to_string(),
+        )
+    } else if v.len() > 64 {
+        Err(
+            "Invalid distribution name. Distribution name must not be longer than 64 characters."
+                .to_string(),
```
jjbayer, line 62: This gets me thinking, how do we handle unicode characters? They might exceed string length 64 even if the number of characters is lower. I assume that the limit enforced on sentry side is applied to the number of characters, not the number of bytes.

kamilogorek:
Looks good. Can you update unicode handling mentioned above? As for lint errors, they should be resolved on master now.

olksdr:
@kamilogorek  done, PTAL!

olksdr:
I've canceled the CI since there is a test which does not progress.

This test is stuck on my mac laptop as well: 
```
test integration::send_event::command_send_event_raw has been running for over 60 seconds
```

kamilogorek:
No worries, fixed it

kamilogorek (approved):

Issue #1391, opened 2022-11-08
Extend node support to >=10


mydea:
Currently, sentry-cli v2.x works on node >= 12, while v1.x requires node >= 8. 

In order to be able to bump sentry-cli from 1.x to 2.x in https://github.com/getsentry/sentry-webpack-plugin / https://github.com/getsentry/sentry-javascript-bundler-plugins, where we want to support node >= 10, it would be very helpful to extend sentry-cli node support to 10 as well.

Otherwise, we'll have to keep adding security fixes etc. to sentry-cli 1.x, which is _possible_ but also not ideal.

kamilogorek:
Isn't the whole idea behind `sentry-javascript-bundler-plugins`, to stop relying on `sentry-cli` itself? cc @lforst @vladanpaunovic @smeubank

Lms24:
>Isn't the whole idea behind sentry-javascript-bundler-plugins, to stop relying on sentry-cli itself? cc @lforst @vladanpaunovic @smeubank

Eventually yes, however, for time reasons we opted to still use the CLI for now as we need to prioritize feature parity with the webpack plugin over getting rid of the CLI dependency. We also discussed trying to compile parts of the rust code to WASM or building a "CLI light" as alternatives. 

There's more to the story though. If we can make cli v2 compatible with the webpack plugin, we could fix the security vulnerability present in v1 as well. Meaning, we could drop v1 entirely eventually(?)

IIRC we said, we'd want to support Node >= 10 since we're probably gonna stop supporting Node 8 in JS SDK v8, meaning  all plugins and the SDKs would have the same minimum Node version (10). Does that sound reasonable?

kamilogorek:
> IIRC we said, we'd want to support Node >= 10 since we're probably gonna stop supporting Node 8 in JS SDK v8, meaning all plugins and the SDKs would have the same minimum Node version (10). Does that sound reasonable?

I was talking about this repo specifically, not the SDK/webpack-plugin.

If that's not huge deal, I'd prefer to keep supporting `1.x` tag instead of degrading minimum version of the `master` branch.

smeubank:
> Isn't the whole idea behind sentry-javascript-bundler-plugins, to stop relying on sentry-cli itself?

more important to me to support more JS bundler than it is to remove dep on sentry-cli that is an internal sentry problem

mydea:
Some more context on this:

[...]

PR #1396, opened 2022-11-14
chore: Setup volta config

mydea:
This adds a volta config, in line with the volta setup for the sentry-javascript repo, to ensure everyone gets the exact same dev experience.

I extracted this out of https://github.com/getsentry/sentry-cli/pull/1392, as it is actually unrelated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,5 +48,9 @@
     "testPathIgnorePatterns": [
       "<rootDir>/src"
     ]
+  },
+  "volta": {
+    "node": "12.22.12",
+    "yarn": "1.22.19"
   }
 }


package.json
```diff
@@ -48,5 +48,9 @@
     "testPathIgnorePatterns": [
       "<rootDir>/src"
     ]
+  },
+  "volta": {
+    "node": "12.22.12",
```
kamilogorek, line 53: Will be out-of-sync once we merge https://github.com/getsentry/sentry-cli/issues/1391 so lets wait for `#1391`.

Issue #1067, opened 2021-11-08
Add tag support to the bash-hook send event command


stevenbdf:
Hey!

I'm using sentry-cli to track issues in my deployment scripts ran on the AWS Amplify Console. I have a `.yml` file where I setup Sentry CLI and run the `eval "$(sentry-cli bash-hook --cli 'sentry-cli -t url:test')"` command to setup sentry in the runner. 

It works fine, but I want to add extra data to the errors and I tought I could use [Sentry Tags](https://docs.sentry.io/product/cli/send-event/#extra-data) using the `-t` parameter, but I found that there's no way I can send parameters to the bash-hook command, located in the `src/bashsupport.sh` script.

I tried to add this feature request by my own, I forked the repo but to be honest I don't have any experience in Rust for building packages. So I will appreciate if somebody could add this simple but important feature.

kamilogorek:
Can you describe your usecase more broadly? It's indeed not possible right now, however, because it's a generic bash hook, you'd get the same tag for every captured error within your script.
I'm asking, because we already do support `environment` and `release` values, and I wonder if maybe that's what you are looking for?

koreno:
Hey, we're also looking for this feature! please merge #1398

PR #1398, opened 2022-11-15
feat: Bash hook tags and extra data

I159:
Allow sending tags and extra data with the bash-hook. Tags and extra data are gathered from an environment using reserved environment variables.

Example:

```
#!/bin/bash

export SENTRY_DSN=https://1c08*****************************86f@o2******6.ingest.sentry.io/56***17
eval "$(sentry-cli bash-hook -e extra_a:123-ars -t tag_a: qwfp-45)"
ls non_existing_dir

```
Closes: [#1067](https://github.com/getsentry/sentry-cli/issues/1067)

### DIFF
--- a/src/bashsupport.sh
+++ b/src/bashsupport.sh
@@ -18,6 +18,30 @@ _sentry_exit_trap() {
   exit $_exit_code
 }
 
+_sentry_gather_tags() {
+    TAGS=' '
+    for tag in $SENTRY_TAGS
+    do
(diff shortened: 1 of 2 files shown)


kamilogorek:
Thanks for the PR! I'm fine adding `tags` to this feature, however with `extra`, we are moving away from them, in favor of structured `context` fields. And because here, both `tags` and `extra` are in `key:value` format anyway, we can just leave `tags` and it should have the exact same capabilities.
Would you mind leaving that one only?

PR #1414, opened 2022-12-07
fix: detect SENTRYCLI_NO_PROGRESS_BAR instead of SENTRY_NO_PROGRESS_BAR

MaxLeiter:
According to the [CHANGELOG](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#removed-apis), `SENTRY_NO_PROGRESS_BAR` was replaced with `SENTRYCLI_NO_PROGRESS_BAR`, but a reference was left in the progress bar code.

### DIFF
--- a/src/utils/progress.rs
+++ b/src/utils/progress.rs
@@ -8,7 +8,7 @@ use crate::utils::logging;
 pub use indicatif::{ProgressDrawTarget, ProgressStyle};
 
 pub fn is_progress_bar_visible() -> bool {
-    env::var("SENTRY_NO_PROGRESS_BAR") != Ok("1".into())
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):
Good catch. It was updated inside `install.js` script, but missed this one. Thanks!

kamilogorek:
Oh wait, I was too fast with the check. `SENTRYCLI_` vars are used for installation part, usually through npm/yarn.
`SENTRY_` vars are used for the binary usage itself. What should be updated instead is the changelog, which should be more specific and say what the change refers to.

In the rust code itself, we use `SENTRY_`, see: https://docs.sentry.io/product/cli/configuration/

PR #1415, opened 2022-12-07
Prevent Sentry from attempting to access unbound variables

LeonMelis:
Bash has a nifty feature `set -u` to error on unbound variables. This is also trappable (like with `set -e`), so Sentry can catch this.

However, Sentry itself may attempt to access arguments via `$1` and `$2`, which may not be set, which will trigger a bash error if `set -u` is used. This will create an error output: `./my_script.sh: line 42: $1: unbound variable`.

The proposed fix here will check for existence of `$1...$n` via the argument count `$#`, without accessing `$1...$n`, making it possible to use Sentry for bash scripts with `set -u`

### DIFF
--- a/src/bashsupport.sh
+++ b/src/bashsupport.sh
@@ -21,10 +21,10 @@ _sentry_exit_trap() {
 _sentry_err_trap() {
   local _exit_code="$?"
   local _command="${BASH_COMMAND:-unknown}"
-  if [ "x$1" != x ]; then
+  if [ $# -ge 1 ] && [ "x$1" != x ]; then
     _command="$1"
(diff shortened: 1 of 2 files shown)


LeonMelis:
Maybe someone from Sentry should have a look at this, before my PR is closed automatically due to inactivity...

kamilogorek:
@LeonMelis I missed that one as I'm on vacation the whole of December, will take a look at it after the new year. Cheers.

kamilogorek:
Makes perfect sense, thanks!

kamilogorek (approved):

PR #1442, opened 2023-01-26
chore: apply clippy fixes for Rust v1.67

vaind:


kamilogorek:
All will be done in https://github.com/getsentry/sentry-cli/pull/1440

PR #1463, opened 2023-02-07
feat: extract and upload embedded Portable PDB from PE

vaind:
* closes https://github.com/getsentry/symbolic/issues/750
* Depends on https://github.com/getsentry/symbolic/pull/757 and a new `symbolic` release v11.1.0

If CLI finds a PE file with embedded PPDB, it extracts the file and uploads it separately. This way, we don't need on any custom handling in the `symbolicator` and `symbolic-debuginfo` (feature flags) and also we don't upload the PE file just for the sake of the .pdb.

The following shows DIF uploads for `samples\Sentry.Samples.Console.Basic\Sentry.Samples.Console.Basic.csproj` with `<DebugType>embedded</DebugType>`
image


Swatinem (approved):
code lgtm.
Though I do not fully understand the screenshot you posted.

Why do we suddenly get two files? Shouldn’t the portable pdb have everything in it? Or is it rather that sentry-cli creates a *separate* sourcebundle because you used `--include-sources` and it was collecting all the local files referenced from the portable pdb?

vaind:
> Why do we suddenly get two files? Shouldn’t the portable pdb have everything in it? Or is it rather that sentry-cli creates a separate sourcebundle because you used --include-sources and it was collecting all the local files referenced from the portable pdb?

Yes, that run was without "embedded sources" (`<EmbedAllSoruces>` flag in the project) so the CLI looked for sources manually and created a source bundle

Swatinem:
Perfect, that makes sense! Will it do the same when you actually have embedded sources though?

vaind:
> Perfect, that makes sense! Will it do the same when you actually have embedded sources though?

No, it's just a single file. I was trying to find my yesterday's attempt but there were too many so wasn't sure which one, so ran it again now:
image

BTW it doesn't show that it includes sources, because Sentry isn't updated with the lastest symbolic v11 - https://github.com/getsentry/sentry/pull/43832

src/utils/dif.rs
```diff
@@ -335,11 +335,28 @@ impl<'a> DifFile<'a> {
...
                     features.symtab = features.symtab || object.has_symbols();
                     features.debug = features.debug || object.has_debug_info();
                     features.unwind = features.unwind || object.has_unwind_info();
                     features.sources = features.sources || object.has_sources();
+                };
+
+                for object in archive.get().objects().filter_map(Result::ok) {
+                    add_object_features(&object);
+
+                    // Combine features with an embedded Portable PDB, if any.
```
Swatinem, line 349: the reason we are doing this is because otherwise the object does not have any features at all, and we would not upload it at all?

src/utils/dif_upload.rs
```diff
@@ -123,14 +124,11 @@ impl<'data> DifMatch<'data> {
             Object::parse(unsafe { &*b }).map(|object| ParsedDif::Object(Box::new(object)))
         })?;
 
-        // Even though we could supply the debug_id here from the object we do not, the
-        // server will do the same anyway and we actually have control over the version of
-        // the code running there so can fix bugs more reliably.
```
Swatinem, line 128: I believe moving this comment to line ~200 below would make sense.

Swatinem (approved):

src/utils/dif.rs
Swatinem: Oh, I saw the new testcase that should report sources in `dif check`

vaind: yes, but that gave me another idea... I have to check whether this recent change doesn't cause the .dll itself to be uploaded now (uselessly)

src/utils/dif_upload.rs
vaind: done

src/utils/dif.rs
vaind: ok, works fine (not do what I was afraid of). Updated the comment to clarify why it's here.

Issue #1482, opened 2023-02-18
Skipping upload sources logic needs to change


mattjohnsonpint:
In a .NET Portable PDB file, if we use `sentry-cli dif upload --include-sources`, we should upload source code files as long as they aren't already embedded in the PDB.

The logic for determining that is here:

https://github.com/getsentry/sentry-cli/blob/cf74876640792c8d194a1acb169cc777d3011823/src/utils/dif_upload.rs#L1159-L1164

However, that is skipping uploading *any* source files, if there are *any* sources embedded.  In reality, there could be (for example) 10 actual source files on disk and only 1 of them embedded in the PDB, so the other 9 never get uploaded.

It *could* be that the other files have SourceLink URLs, but not necessarily.

It can also happen that a source generator could create code during compilation, and that generated code would get embedded in the PDB - which makes sense because it's not actually on disk.  In such a case, we'd want to have both the embedded source code, and the files on disk that were not embedded, all be available to show in Sentry.

I believe we should check each file that is referenced in the PDB.  If it's embedded then we can skip it.  If not, then we have to go get it from disk.  We can't skip all source files because just one was embedded.

This was discovered through customer interaction here:
https://discord.com/channels/621778831602221064/1076234788916183162

vaind:
This should be a quick fix on the CLI part, I would, however, wait for a new `symbolic` release with the source-link support so we can handle those files properly right from the get-go.

Additionally, we should check `symbolicator` how the situation is handled when there may be sources in multiple DIFs matching a single debug ID.

vaind:
@mattjohnsonpint can we get a small dll that we could include in the repo to test against? I've tried against https://nuget.info/packages/Microsoft.Orleans.Core/7.1.0 but I think that will also hit the licensing roadblock

mattjohnsonpint:
Yep. I'm on vacation, but I'll put one together when I get back on Wednesday. Thanks.

mattjohnsonpint:
Here's a small app that reproduces the issue.

[SrcGenSampleApp.zip](https://github.com/getsentry/sentry-cli/files/10834373/SrcGenSampleApp.zip)

It's basically just the getting started with source generators sample from here: https://learn.microsoft.com/dotnet/csharp/roslyn-sdk/source-generators-overview

The PDB at `SrcGenSampleApp/bin/Release/net6.0/SrcGenSampleApp.pdb` references two source files:
[...]

PR #1486, opened 2023-02-21
fix: include sources referenced but not embedded by an object file

vaind:
* Closes https://github.com/getsentry/sentry-cli/issues/1482
* Depends on symbolic v12 (and https://github.com/getsentry/symbolic/pull/764 to be part of it).

This changes the bundle-source collection behaviour to include sources that are referenced by an object file but not embedded directly in it, even if the object file contains some embedded sources alongside referenced ones.


mattjohnsonpint:
Seems like this could be merged now?

kamilogorek (approved):

PR #1489, opened 2023-02-22
chore: update clap to v4

vaind:
Following the migration guard from the changelog: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28

This was even more painful than the deprecations in v3.2

The biggest change was that all args with v4 now take a value by default, instead of being flags.


kamilogorek:
Tbh I tried to do this one day and gave up, as I've seen no real value in moving to v4 😅

vaind:
> Tbh I tried to do this one day and gave up, as I've seen no real value in moving to v4 😅

Well I've been too fed up with the trailing spaces in the help files constantly being removed when I change something (and my editor persistently removing it) so I tried updating... not exactly as easy as I've hoped for, but it's "done", at least...

kamilogorek (approved):
I really hope no subcommand was missed in the tests 🤞

PR #1494, opened 2023-02-28
ref: also distribute sentry-cli as a pip package

asottile-sentry:
this packages up this binary as something that's `pip install`able -- hopefully making the setup process easier and not relying on the uptime of github packages or the sentry cdn for installation

how this works:
1. build a sdist and wheel from source (mostly to demonstrate that building from source works for _unsupported_ platforms)
2. download the prebuilt artifacts that we distribute through github releases
3. stitch the prebuilt binaries into wheels, adjusting the platform tag to match

(this saves building everything twice, and ensures that identical binaries are present independent of how it installs)

I've also pushed this to `release/just-testing` so I can validate that the build works (it should show in this PR too!)

```console
$ ./venv/bin/pip install y/sentry_cli-2.13.0-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl 
Processing ./y/sentry_cli-2.13.0-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
Installing collected packages: sentry-cli
Successfully installed sentry-cli-2.13.0
$ ./venv/bin/sentry-cli --version
sentry-cli 2.13.0
```

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -63,6 +63,9 @@ targets:
         checksums:
           - algorithm: sha256
             format: hex
+  - name: pypi
+  - name: sentry-pypi
+    internalPypiRepo: getsentry/pypi
 requireNames:
   - /^sentry-cli-Darwin-x86_64$/
   - /^sentry-cli-Darwin-arm64$/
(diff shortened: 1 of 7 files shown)


mitsuhiko (approved):
Longer term would be nice not to have a setup.py here but I'm totally okay with this for now

Issue #1495, opened 2023-03-02
2.14.0 breaking change


h3rmanj:
### Environment

`getsentry/sentry-cli:2.14.0` on docker hub.

### Steps to Reproduce

Running these in my CI

```
    - sentry-cli releases new $VERSION
    - sentry-cli releases files $VERSION upload-sourcemaps --url-prefix '~/assets' build/assets --rewrite
    - sentry-cli releases finalize $VERSION
    - sentry-cli releases set-commits --auto $VERSION || echo 'Found no commits to assign to release.'
    - sentry-cli releases deploys $VERSION new -e $SENTRY_ENV
```

### Expected Result

Succeess

### Actual Result

`sentry-cli releases files $VERSION upload-sourcemaps --url-prefix '~/assets' build/assets --rewrite` command gives this error:  
`error: A release slug is required (provide with --release)`

### Workaround

Pinned my cli version to `2.13.0`

kamilogorek:
cc @vaind it apparently broke with https://github.com/getsentry/sentry-cli/pull/1489 as we dont have integration tests for legacy APIs :(

@h3rmanj in the meantime you can fix this by replacing

```
sentry-cli releases files $VERSION upload-sourcemaps --url-prefix '~/assets' build/assets --rewrite
```

with

```
sentry-cli sourcemaps upload build/assets --release $VERSION --url-prefix '~/assets'
```

(`--rewrite` is a default in v2, so no need to set it)

fdintino:
Which part is the legacy API? The [current sentry docs](https://docs.sentry.io/product/cli/releases/#finalizing-releases) have the release version passed as a positional argument, not with `--release`.

fubhy:
This broke our github ci pipeline too.

pgraemer:
same here.

eXaminator:
I'm with @kamilogorek. This also broke our pipeline and we also did it as described in the docs. So I hope this will be fixed soon and if this API is legacy / deprecated it should be documented as such.

kamilogorek:
re docs: https://github.com/getsentry/sentry-cli/issues/1429

This is a "legacy" API, which is supposed to be fully compatible so that we don't have to update the docs right away. This issue is a bug and will be fixed asap, no need to migrate to another API in v2 :)

avg07:
> cc @vaind it apparently broke with #1489 as we dont have integration tests for legacy APIs :(
> 
> @h3rmanj in the meantime you can fix this by replacing
[...]

PR #1496, opened 2023-03-02
revert: clap v4 update & test releases files upload-sourcemap

vaind:
This reverts #1489 to workaround the regression in #1495

Actual fix should follow later.
This also adds a test case to verify the regression doesn't happen again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,21 +330,23 @@ dependencies = [
 
 [[package]]
 name = "clap"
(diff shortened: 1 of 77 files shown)


kamilogorek (approved):

kamilogorek:
NOTE: Backward compatibility needs to be preserved for `releases files` and `releases deploys` (second one lacks the test).

PR #1500, opened 2023-03-03
chore: reapply clap update to v4 & fix releases subcommands

vaind:
Follows up on #1496 - for some reason, the "version" arg wasn't `global(true)` and yet it still got propagated down the road to subcommands - an "undefined behavior" that stopped working in clap v4

- reapply: clap-rs v4 update
- fix: `releases files` and `releases deploys` version propagation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,23 +330,21 @@ dependencies = [
 
 [[package]]
 name = "clap"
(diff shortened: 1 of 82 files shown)


kamilogorek (approved):
Hopefully tests will catch everything this time 🙏

PR #1528, opened 2023-03-15
wip: feat(view-hierarchy): Allow dartsymbols.json file upload

narsaynorath:
There's a debug file generated by the Flutter obfuscation process that we want to make use of. We'd like to update `sentry-cli` to detect this file and submit it as a debug file. The corresponding sentry PR is here: https://github.com/getsentry/sentry/pull/45837

I was trying to follow the changes previously added for `il2cpp` in https://github.com/getsentry/sentry-cli/pull/1248/ but my changes diverged as I stepped through some of the code. In its current state, there are issues calculating a uuid when it detects a dartsymbols.json file. I'm not exactly sure how to resolve this so I'm seeking some guidance here.

### DIFF
--- a/src/utils/dif_upload.rs
+++ b/src/utils/dif_upload.rs
@@ -81,6 +81,7 @@ enum ParsedDif<'a> {
     BcSymbolMap(BcSymbolMap<'a>),
     UuidMap(UuidMapping),
     Il2Cpp,
+    DartSymbols,
 }
 
 impl<'slf, 'data: 'slf> AsSelf<'slf> for ParsedDif<'data> {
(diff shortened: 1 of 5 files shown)


src/utils/dif_upload.rs
```diff
@@ -184,6 +185,21 @@ impl<'data> DifMatch<'data> {
         })
     }
 
+    fn from_dartsymbols(uuid: DebugId, name: String, data: ByteView<'static>) -> Result<Self> {
+        // TODO(Nar): This was a guess, but I want to just read the data as JSON without modifications
```
kamilogorek, line 189: Not sure why you need this function at all if you are mapping it to a raw enum anyway?

src/utils/dif_upload.rs
```diff
@@ -775,9 +800,13 @@ fn collect_auxdif<'a>(
         .file_stem()
         .map(|stem| stem.to_string_lossy())
         .unwrap_or_default();
+
+    // TODO(Nar): Can't seem to get past this parse call
```
kamilogorek, line 804: It's trying to parse the path of the file using https://github.com/getsentry/rust-debugid which is not implemented for dartsymbols. But from what I see, you shouldnt go through the `auxdif` at all, as you'll be collecting those files directly. This however, requires you to first implement a new format at https://github.com/getsentry/symbolic/blob/master/symbolic-debuginfo/src/lib.rs

kamilogorek:
This whole implementation needs to branch-out to other repositories first (starting with `symbolic`) to add handling of this file format, as unfortunately it's not that trivial to just add it here.

PR #1536, opened 2023-03-21
feat(auth): Support DSN auth for the API client

evanpurkhiser:
This is in support of https://github.com/getsentry/sentry-cli/issues/1537

Explicitly does not add this as a global cli parameter

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1812,6 +1812,10 @@ impl ApiRequest {
                 debug!("using token authentication");
                 self.with_header("Authorization", &format!("Bearer {token}"))
             }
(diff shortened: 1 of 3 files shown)


kamilogorek (approved):

evanpurkhiser:
Should I add a changelog entry?

kamilogorek:
We usually generate changelog automatically based on the commit messages, and those are nice so far so you should be fine without doing this by hand. You can also edit minor PRs, like the one about wording change and add `#skip-changelog` to the description, which will exclude it from the changelog (even after it was already merged).

Issue #1537, opened 2023-03-21
Crons: Support DSN auth for monitor checkins


evanpurkhiser:
So the situation that currently exists is: There are currently 2 ways to identify your monitors, and 3 ways to do monitor checkins.

1. Your monitor can be identified by a globally unique UUID. This is currently called the `monitor_id` We are trying to phase this out in favor of the second means of identification, `monitor_slug`
2. The monitor slug is unique **within an organization** it is also user configurable. There are various reasons we want to prefer this, the main being it allows for automatic creation of monitors using existing names in your codebase (think celery integrations, where the names of the celery tasks can be used to identify the monitors)

Then for checkins

1. We support checkins via ingest. Long term **this would be the ideal way for monitors to checkin via the CLI.** In the future relay will have monitor ingest endpoints that can be used by the CLI with **only DSN auth.** Alternatively the CLI could send events via envelopes.
2. We have 2 existing “legacy” API endpoints, these endpoints are part of the sentry app API and are being deprecated in favor of all checkins being ingested through relay.
3. `/orgaizations/{orgSlug}/monitors/{monitor_id|monitor_slug}/checkin` — This endpoint provides organization context, allowing the monitor slug to be looked up **no matter if DSN auth or Token auth** is being used.
4. `/monitors/{monitor_id|monitor_slug}/checkin` — This endpoint provides **no org context.** Meaning when Token Auth is used **we are not able to lookup a monitor via it’s slug.** This is a problem as we want to suggest users ONLY use monitor slugs (which can be configured) and do NOT use monitor GUIDs. So in order for this endpoint to accept a monitor slug and be able to determine the organization context **DSN Auth MUST be used.** Unfortunately, this endpoint is **also the endpoint being used in the sentry-cli.**

So we have two options

1. Update the monitor command to **require an organization** which would allow for token auth to be used.
    1. This is not the ideal solution, since in the future **we will use ingestion which will only support DSN auth.** So forcing a user to update their CLI usage to pass an organization slug, where down the line we will update the usage to NOT require a org slug and instead use DSN auth is not ideal.
2. Update the monitor command to **use DSN Auth.**
    1. This is good as it future proofs the interface for the user to make checkins with the CLI.
    2. Also many of our users are confused about what scopes their API token needs for handling monitor checkins, switching to DSN Auth **completely mitigates this**. Which would be nice

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cli/pull/1536
[...]

PR #1540, opened 2023-03-21
ref(monitors): Prefer slugs over guids

evanpurkhiser:
Updates tests, help strings, and some small bits of logic to prefer calling monitor identifiers 'slugs' instead of simply IDs. This is in line with what we're doing for #1537

I would like to wait for the follow up (support DSN auth) to be ready before merging this, as you will ONLY be able to use legacy GUID/ID identifiers for monitors when using toke auth


src/commands/monitors/run.rs
```diff
@@ -13,10 +12,10 @@ pub fn make_command(command: Command) -> Command {
     command
         .about("Wraps a command")
         .arg(
-            Arg::new("monitor")
-                .help("The monitor ID")
-                .required(true)
-                .value_parser(Uuid::parse_str),
+            Arg::new("monitor_slug")
+                .value_name("monitor-slug")
```
kamilogorek, line 16: This is a breaking change and we need to preserve a hidden alias that'd still support UUID.

kamilogorek (changes requested):

src/commands/monitors/run.rs
evanpurkhiser: all slugs on the backend match existing UUIDs, it is fully backwards compatible.

evanpurkhiser: See https://github.com/getsentry/sentry/issues/45166

kamilogorek: Ah, perfect 😌

PR #1543, opened 2023-03-22
test: Allow auth method to be specified in tests

evanpurkhiser:
Prior to this there was now way to specify that you wanted to use DSN auth or token auth. This makes things more clear.


evanpurkhiser:
Closed in favor of https://github.com/getsentry/sentry-cli/pull/1545

PR #1541, opened 2023-03-22
ref(crons): Prefer DSN based auth

evanpurkhiser:
Updates the cron monitors run command to prefer using DSN auth, including spitting out a warning when DSN Auth is not used.

Requires https://github.com/getsentry/sentry-cli/pull/1543

PR #1545, opened 2023-03-24
ref(crons): Prefer DSN based auth

kamilogorek:
Replaces https://github.com/getsentry/sentry-cli/pull/1541 and https://github.com/getsentry/sentry-cli/pull/1543

### DIFF
--- a/tests/integration/monitors/run.rs
+++ b/tests/integration/monitors/run.rs
@@ -6,10 +6,24 @@ fn command_monitors_run() {
         EndpointOptions::new("POST", "/api/0/monitors/foo-monitor/checkins/", 200)
             .with_response_file("monitors/post-monitors.json"),
     );
+
     if cfg!(windows) {
         register_test("monitors/monitors-run-win.trycmd");
     } else {
         register_test("monitors/monitors-run.trycmd");
+    }
+}
+
+#[test]
+fn command_monitors_run_token_auth() {
+    let _server = mock_endpoint(
+        EndpointOptions::new("POST", "/api/0/monitors/foo-monitor/checkins/", 200)
+            .with_response_file("monitors/post-monitors.json"),
+    );
+    if cfg!(windows) {
+        register_test("monitors/monitors-run-token-auth-win.trycmd").env("SENTRY_DSN", "");
+    } else {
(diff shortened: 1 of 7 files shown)


Swatinem (approved):

kamilogorek:
@evanpurkhiser feel free to merge it once you test it out

evanpurkhiser (approved):

tests/integration/monitors/run.rs
```diff
@@ -6,13 +6,27 @@ fn command_monitors_run() {
...
 }
 
+#[test]
+fn command_monitors_run_token_auth() {
+    let _server = mock_endpoint(
+        EndpointOptions::new("POST", "/api/0/monitors/foo-monitor/checkins/", 200)
+            .with_response_file("monitors/post-monitors.json"),
+    );
+    if cfg!(windows) {
+        register_test("monitors/monitors-run-token-auth-win.trycmd").env("SENTRY_DSN", "");
```
evanpurkhiser, line 24: Ahh, I was trying to clear it out like this but that wasn't working since even having a blank value was reading it :-)

There must be some logic somewhere that treats empty as None in the new code

kamilogorek: Yup, it's here - https://github.com/getsentry/sentry-cli/blob/ac6532fc1aa69ff9f9d2bf74f4a6fe0b09f67c89/src/config.rs#L407-L416
Empty string fails to `.parse()` as `Dsn` type.

mattjohnsonpint:
Some explanation would be useful.  What is "DSN based auth"? And does it only apply to certain Sentry CLI operations, or anything?  For example, does this mean I could use a DSN instead of an auth token to upload debug info files?  Thanks.

mattjohnsonpint:
... I'm guessing not.  But at any rate, it DSN based auth it a thing for Sentry CLI, should it be mentioned in docs, such as in [Configuration and Authentication](https://docs.sentry.io/product/cli/configuration/)?  Thanks.

kamilogorek:
DSN auth is not too accurate tbh. It's the same "auth" as you'd do with using `send-event` or `send-envelope`. Which is more of an "identification" than "authorization".

PR #1552, opened 2023-03-30
Skip empty SourceMap uploads

Swatinem:
This avoids uploading empty (but still unique until we pull in https://github.com/getsentry/symbolic/pull/778) artifact / release bundles if all the source maps included in them have already been uploaded before.


kamilogorek (approved):
Updated tests

PR #1555, opened 2023-03-31
feat: Add `forceArtifactBundles` bundles option to JS interface

lforst:
To support our new debug ID process we add the `forceArtifactBundles` which will set the `SENTRY_FORCE_ARTIFACT_BUNDLES` env var in the Sentry CLI child process.

Ref: https://github.com/getsentry/sentry-docs/pull/6551

Some prettier changes snuck in but i think its fine.


kamilogorek (approved):
Note that we'll most likely move to a separate flag instead of envvar, but its fine to have it here for now.

PR #1556, opened 2023-03-31
fix: Fix spawn ENOENT for vite integration

alexgrozav:
Fixes vite plugin integration error.

~~~
alexgrozav@Alexs-MacBook-Pro editor-ui % pnpm run build

> n8n-editor-ui@0.188.0 build /Users/alexgrozav/Workspace/n8n/n8n/packages/editor-ui
> cross-env NODE_OPTIONS="--max-old-space-size=8192" vite build

node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

Error: spawn /Users/alexgrozav/Workspace/n8n/n8n/node_modules/.pnpm/@sentry+cli@2.16.1/node_modules/@sentry/cli/sentry-cli ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:485:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn /Users/alexgrozav/Workspace/n8n/n8n/node_modules/.pnpm/@sentry+cli@2.16.1/node_modules/@sentry/cli/sentry-cli',
  path: '/Users/alexgrozav/Workspace/n8n/n8n/node_modules/.pnpm/@sentry+cli@2.16.1/node_modules/@sentry/cli/sentry-cli',
  spawnargs: [ 'releases', 'propose-version' ],
  cmd: '/Users/alexgrozav/Workspace/n8n/n8n/node_modules/.pnpm/@sentry+cli@2.16.1/node_modules/@sentry/cli/sentry-cli releases propose-version'
}
~~~

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -15,6 +15,7 @@ function getBinaryPath() {
   const parts = [];
   parts.push(__dirname);
   parts.push('..');
+  parts.push('bin');
   parts.push(`sentry-cli${process.platform === 'win32' ? '.exe' : ''}`);
   return path.resolve(...parts);
 }


kamilogorek:
This path is the same one that we use for https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js script, so I'm not sure how that change would make a difference tbh.
This looks like an issue with running (possibly caching a result?) of `postinstall` script, which produces the binary itself, which is then not present during execution.
Do you have a repro repository that we could use to debug this?

alexgrozav:
@kamilogorek You can find the exact usage here: https://github.com/n8n-io/n8n/pull/5870/files 

I had to patch the package in order to get it to work. Perhaps could have something to do with the fact that it's a monorepo? I'm not entirely sure, but it's worth looking into.

js/helper.js
```diff
@@ -15,6 +15,7 @@ function getBinaryPath() {
   const parts = [];
```
netroy, line 15: Could we use `require.resolve` instead here, like this:
```javascript
path.join(require.resolve('@sentry/cli'), `../../bin/sentry-cli${process.platform === 'win32' ? '.exe' : ''}`)
```

kamilogorek: No, unfortunately, we specifically cannot use `require.resolve` due to `@vercel/nft` behavior.

See https://github.com/getsentry/sentry-cli/pull/1207 and a follow-up in https://github.com/getsentry/sentry-cli/pull/1374

> Perhaps could have something to do with the fact that it's a monorepo? I'm not entirely sure, but it's worth looking into.

Yes, that's most likely it, as it's making links between shared packages in the root directory. I'm out for the next few weeks, but will keep that issue in mind.

kamilogorek:
> Perhaps could have something to do with the fact that it's a monorepo? I'm not entirely sure, but it's worth looking into.

Yes, that's most likely it, as it's making links between shared packages in the root directory. I'm out for the next few weeks, but will keep that issue in mind.

PR #1557, opened 2023-03-31
feat(inject): Add --use-debug-ids flag to force Artifact Bundles usage

kamilogorek:
Name is up for discussion.

- remove `SENTRY_FORCE_ARTIFACT_BUNDLES`
- remove `forceArtifactBundles` from JS API in favor of `useDebugIds`
- restore chunk options endpoint method to the old, non-overriding behavior
- add `--use-debug-ids` flag that overrides the server capability

### DIFF
--- a/src/commands/sourcemaps/upload.rs
+++ b/src/commands/sourcemaps/upload.rs
@@ -1,4 +1,3 @@
-use std::env;
 use std::path::PathBuf;
 
 use anyhow::Result;
@@ -6,7 +5,7 @@ use clap::{Arg, ArgAction, ArgMatches, Command};
 use glob::{glob_with, MatchOptions};
 use log::{debug, warn};
 
-use crate::api::Api;
+use crate::api::{Api, ChunkUploadCapability};
 use crate::config::Config;
 use crate::utils::args::validate_distribution;
 use crate::utils::file_search::ReleaseFileSearch;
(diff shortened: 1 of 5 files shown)


lforst (approved):
JS part is approved

Jesse-Box:
@kamilogorek 

Since we print the `Bundle Type: Artifact Bundle` after the upload command has been triggered. Wouldn't it make more sense to name the flag something like: `-type-artifact`?

Another question; is there any other flag a user can use that isn't interoperable with this new flag?

kamilogorek:
> Another question; is there any other flag a user can use that isn't interoperable with this new flag?

Good questions, because if yes, then we can actually model this inside flags structure (the lib we use allows to make one flag prevent use of other flag)

mitsuhiko (approved):

src/commands/sourcemaps/upload.rs
```diff
@@ -180,6 +180,17 @@ pub fn make_command(command: Command) -> Command {
                     Defaults to: `--ext=js --ext=map --ext=jsbundle --ext=bundle`",
                 ),
         )
+        // NOTE: Hidden until we decide to expose it publicly
+        .arg(
+            Arg::new("use_debug_ids")
+                .long("use-debug-ids")
+                .action(ArgAction::SetTrue)
+                .help(
+                    "Use new Artifact Bundles upload, that enables use of Debug ID \
```
iambriccardo, line 189:
```suggestion
                    "Use new Artifact Bundles upload, that enables the use of Debug IDs \
```

iambriccardo (approved):
LGTM

Issue #1514, opened 2023-03-09
`sentry-cli bash-hook` should support tags and context


ckdake:
When using other client libraries, e.g. the Ruby one, we can initialize Sentry locally settings tags/context so that we can group/report in the Sentry UI using them. e.g.:

```ruby
tenant="example"

Sentry.set_tags(tenant: tenant)
Sentry.set_context("tenant", {tenant: tenant})
```

This works great for the rails app, and we'd like to do something similar for our backend jobs using a different framework. It's straightforward enough to use sentry-cli there, e.g.:

```bash
if [[ -v SENTRY_DSN ]]; then
    echo "INFO: SENTRY_DSN is set, so we can use it to catch errors"
    eval "$(sentry-cli bash-hook)"
fi
```

But every event from every tenant is grouped into the same single BashError event in Sentry. It would be much better to be able to split these up for tenant, I'd expect this to work something like:

```bash
TENANT=example

if [[ -v SENTRY_DSN ]]; then
    echo "INFO: SENTRY_DSN is set, so we can use it to catch errors"
    eval "$(sentry-cli bash-hook --tags tenant=$TENANT --context tenant={tenant: $TENANT})"
fi
```

ckdake:
I'd still like this to be considered. Thanks.

kamilogorek:
There has been an attempt implementing this but wasnt taken through the finish line - https://github.com/getsentry/sentry-cli/pull/1398
If anyone wants to pick this up from there, should be rather straightforward to proceed :)

PR #1560, opened 2023-04-03
feat: Adds tags to bash hook

ckdake:
building on https://github.com/getsentry/sentry-cli/pull/1398/files to address https://github.com/getsentry/sentry-cli/issues/1514

### DIFF
--- a/src/commands/bash_hook.rs
+++ b/src/commands/bash_hook.rs
@@ -5,7 +5,7 @@ use std::fs;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
 
-use anyhow::Result;
+use anyhow::{format_err, Result};
 use clap::{builder::ArgPredicate, Arg, ArgAction, ArgMatches, Command};
 use lazy_static::lazy_static;
 use regex::Regex;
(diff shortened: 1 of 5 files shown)


src/commands/bash_hook.rs
```diff
@@ -197,6 +230,34 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         )
         .replace("___SENTRY_LOG_FILE___", &log.display().to_string());
 
+    let mut hook_tags = "".to_string();
```
kamilogorek, line 233: This will break for multiple tags, as it'll be concatenated to `-t foo-t bar-tbaz`

src/commands/bash_hook.rs
```diff
@@ -197,6 +230,34 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    for tag in tags {
+        hook_tags.push_str(&"-t ");
+        hook_tags.push_str(format!("\"{}\"", tag.to_string()));
+    }
+    if hook_tags.is_empty() {
```
kamilogorek, line 238: If `hook_tags` is indeed empty, replacing the placeholder with `&""` will be effectively the same as `&hook_tags.to_string()` :)

src/commands/bash_hook.rs
```diff
@@ -197,6 +230,34 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    } else {
+        script = script.replace(
+            "___SENTRY_TAGS___",
+            &hook_tags.to_string(),
+        );
+    }
+
+    let mut hook_contexts = "".to_string();
+    for context in contexts {
+        hook_contexts.push_str(&"--extra ");
```
kamilogorek, line 249:
```suggestion
        hook_contexts.push_str(&"--contexts ");
```

src/commands/bash_hook.rs
```diff
@@ -197,6 +230,34 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        hook_contexts.push_str(format!("\"{}\"", context.to_string()));
+    }
+    if hook_contexts.is_empty() {
+        script = script.replace("___SENTRY_TAGS___", &"")
```
kamilogorek, line 253:
```suggestion
        script = script.replace("___SENTRY_CONTEXTS___", &"")
```

src/commands/bash_hook.rs
kamilogorek: Both comments are also applicable to `contexts`

kamilogorek:
```
make check
make test
```

If they pass locally, they'll pass here as well :) (maybe except windows tests, as those are run conditionally :p)

ckdake:
Context looks a little tricker due to types, and I really just need tags so I'm biasing towards just getting that working.  

First Rust PR ;)

ckdake:
First Rusting here, so my local env wasn't letting me run `make check`. I had to roll back my `brew install rust` and do `brew install rustup` as a start.    `make check` is now passing.

ckdake:
Those all make sense, I spent too much time tracing around on on what became `&[&String]`.  I had a .iter().map() drafting too, but types were slowing me down there too. Thanks!

kamilogorek (approved):
Lovely. Thanks for contribution! :)

PR #1574, opened 2023-04-11
bugfix: uniquify GHA upload-step names, to avoid race condition

bukzor-sentryio:


ashwoods (approved):

.github/workflows/build.yml
```diff
@@ -1,9 +1,10 @@
-name: Release Build
+nbinariesame: Release Build
```
mattgauntseo-sentry, line 1:
```suggestion
name: Release Build
```

.github/workflows/build.yml
```diff
@@ -41,7 +42,7 @@ jobs:
 
       - uses: actions/upload-artifact@v3.1.1
         with:
-          name: ${{ github.sha }}
+          name: ${{ github.sha }}-binaries
```
mattgauntseo-sentry, line 45: Should this be `linux` if we're going with `macos` in the following job?

mattgauntseo-sentry (changes requested):

Issue #1564, opened 2023-04-04
Crons: Add Support for Environments


gaprl:
With [Crons: Add support for multiple environments#42788](https://github.com/getsentry/sentry/issues/42788) we now enable check-ins to have an environment, such as `production` or `development`, and we need to update the Sentry CLI to enable such feature.

PR #1599, opened 2023-05-02
feat(crons): Allow specifying checkin environment

evanpurkhiser:
Fixes https://github.com/getsentry/sentry-cli/issues/1564


src/api.rs
```diff
@@ -2503,6 +2505,8 @@ pub struct UpdateMonitorCheckIn {
     pub status: Option<MonitorCheckinStatus>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub duration: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub environment: Option<String>,
```
kamilogorek, line 2509: Why is it `Option` if we have a default value?

evanpurkhiser: It's not absolutely required to send, this reflects that should the API be used anywhere else

kamilogorek (approved):

PR #1616, opened 2023-05-19
Update `sentry` SDK to fix sending raw envelopes

Swatinem:


src/utils/event.rs
```diff
@@ -81,6 +81,8 @@ where
     let client = Client::from_config((
         dsn,
         apply_defaults(ClientOptions {
+            // TODO: do we want to eventually set `debug` based on the CLI log-level?
```
kamilogorek, line 84: It'll already tell you some details now (mostly about whether it actually went out), but would be nice to set it here too imo.

kamilogorek (approved):

Issue #1605, opened 2023-05-09
Sourcemaps upload with --url-prefix behavior is strange


Memfisrain:
### Environment

We have sentry **on-premise** version **22.5.0**
**nodejs** version **16**
**sentry-cli** version **2.17.5**

### Steps to Reproduce

As part of releasing frontend applications, we create Sentry release, set commits, upload source maps and create deployment. The problem is that we have different environment and different CDN for stagings / preprod / prod.

When source maps uploaded first time to release, everything works good. Let's say, I deploy my code on **stg25** , source maps are uploaded using the following command:

`sentry-cli releases -o myOrg -p appjs-sdk files 1e2e77d8 upload-sourcemaps ./dist --url-prefix '~/stg25/appjs-sdk'`

If I open **appjs-sdk** project in Sentry, then go to source maps configuration and select release **1e2e77d8** I will find source files together with source maps. Their names look like: `~/stg25/appjs-sdk/bootstrap-4b2cd0de.js`, which is expected for me.

If after that, I decided to deploy this to preprod, source maps are uploaded using the following command:
`sentry-cli releases -o myOrg -p appjs-sdk files 1e2e77d8 upload-sourcemaps ./dist --url-prefix '~/preprod/appjs-sdk'`

If I open **appjs-sdk** project in Sentry, then go to source maps configuration and select release **1e2e77d8** I **will not** find sources and files with names `~/preprod/appjs-sdk/bootstrap-4b2cd0de.js` which is unexpected for me. Even in Sentry tutorial video https://youtu.be/hEwpjRHRpC8?t=593 source maps are uploaded with different `--url-prefix` and they become visible on UI.

The problem is that we have a production issue in Sentry, and source maps are not applied to it.

Moreover, I see that some of our releases contain source maps and sources for many staging, preprod and prod, so it seems that the issue if flaky.

### Expected Result
Uploading source maps with different `--url-prefix` results in showing them on UI under the release artifact and will apply to Sentry issues.

### Actual Result
Uploading source maps with different `--url-prefix` seems has no effect after first source maps were uploaded with any `--url-prefix`

Memfisrain:
It seems, that I found out the issue with debug mode, in logs I spotted that it fetches the artifacts from release:

[...]

PR #1619, opened 2023-05-22
ref(types): Add dedupe to available TS types

kamilogorek:
Ref https://github.com/getsentry/sentry-cli/issues/1605#issuecomment-1554427232


kamilogorek:
That's fineee. It's green.

PR #1625, opened 2023-05-23
Revert "feat(crons): Replace API checkins w/ send_envelope (#1618)"

evanpurkhiser:
This reverts commit a8db0f069be87590ed12b88b4083b17030c534cb.

Going to re-implement this without dropping support for the auth tokens


kamilogorek:
We decided to leave the commit in place and build on top of it to restore the functionality.

PR #1629, opened 2023-05-26
Test case insensitive when finding repo by name

maxnowack:


kamilogorek:
Although `git` itself *is* case-sensitive, I think it makes sense here for a better UX.

kamilogorek (approved):

Issue #1569, opened 2023-03-28
Hermes bytecode bundle and source maps are not automatically uploaded on iOS  


krystofwoldrich:
### Description

Starting RN `0.71` [react-native docs](https://reactnative.dev/docs/sourcemaps) recommend using `SOURCEMAP_FILE` env to enable source maps on iOS. 

`sentry-cli` doesn't recognize that Hermes source maps were generated (`Warning: build produced no sourcemaps.`) and doesn't upload any artifacts.

At the moment `sentry-cli` only recognizes source maps generated using the following flag `EXTRA_PACKAGER_ARGS="--sourcemap-output $DERIVED_FILE_DIR/main.jsbundle.map"`, but this way doesn't enable Hesmes compiler source maps.

Context: Generating Hermes source map has a side effect. https://github.com/facebook/react-native/issues/34043#issuecomment-1488180858

PR #1667, opened 2023-07-05
Add Debug Id reference to Hermes Bundle from the Source Map

krystofwoldrich:
- Subtask of https://github.com/getsentry/sentry-react-native/issues/2947
- Closes: https://github.com/getsentry/sentry-cli/issues/1569

This PR adds support for RN Hermes bundles with debug IDs. As we can't read Hermes byte code we need to fill the debug id from the referenced source map to the headers of the Hermes bundle.

This PR also updates Xcode Call functionality to be able to intercept `hermesc` and `compose-source-maps.js` calls by RN tooling. This is needed for recording the path of the Hermes Bundle and final Composed Hermes source maps. 

Xcode Call has now also ability to copy debug if from the packager source map to the composed Hermes source map. This way the debug ids will work with both JSC and Hermes builds without directly changing the Hermes bundle. Metro packager plugin will inject the debug id into the plain JS the same we do it on the web. 

Note: In the future, we might be able to inject debug id directly into the Hermes bundle but for now, these changes are needed.
- https://github.com/facebook/hermes/discussions/1122


src/utils/sourcemaps.rs
```diff
@@ -162,18 +178,36 @@ fn guess_sourcemap_reference(sourcemaps: &HashSet<String>, min_url: &str) -> Res
...
+                    Some(&new_ext),
+                )));
             }
         }
     }
 
     bail!("Could not auto-detect referenced sourcemap for {}", min_url);
 }
 
+pub struct SourceMapReference {
```
loewenheim, line 193: Please add some documentation to this struct, it's not easy to understand what `url` and `original_url` are.

src/utils/sourcemaps.rs
```diff
@@ -635,8 +669,71 @@ impl SourceMapProcessor {
...
-            if let Some(Some(sourcemap_url)) = self.sourcemap_references.get(&source.url) {
-                source.set_sourcemap_reference(sourcemap_url.to_string());
+            if let Some(Some(sourcemap)) = self.sourcemap_references.get(&source.url) {
+                source.set_sourcemap_reference(sourcemap.url.to_string());
+            }
+        }
+        Ok(())
+    }
+
+    /// Adds debug id reference from the given source map to all minified files.
```
loewenheim, line 679: What is meant by "the given source map" here?

src/utils/sourcemaps.rs
```diff
@@ -635,8 +669,71 @@ impl SourceMapProcessor {
...
+                    .clone()
+                    .unwrap_or(sourcemap_reference.url.clone());
+
+                if !self.debug_ids.contains_key(sourcemap_url) {
+                    debug!(
+                        "{} No debug id found for {} to reference",
+                        style(">").dim(),
+                        sourcemap_url
+                    );
+                    return Ok(());
```
loewenheim, line 701: Do you intentionally return here, skipping all other iterations of the `for` loop?

src/utils/sourcemaps.rs
```diff
@@ -635,8 +669,71 @@ impl SourceMapProcessor {
...
+                }
+
+                let debug_id_header = (
+                    "debug-id".to_string(),
+                    self.debug_ids[sourcemap_url].to_string(),
+                );
+                if source.headers.contains_key(&debug_id_header.0) {
```
loewenheim, line 708: I added methods `debug_id` and `set_debug_id` to the `SourceFile` struct recently that let you do this without worrying about the headers.

src/utils/sourcemaps.rs
```diff
@@ -635,8 +669,71 @@ impl SourceMapProcessor {
...
+                    continue;
+                }
+
+                debug!(
+                    "{} Adding debug id {} reference to {}",
+                    style(">").dim(),
+                    self.debug_ids[sourcemap_url].to_string(),
+                    source.url
+                );
+                source.headers.insert(debug_id_header.0, debug_id_header.1);
```
loewenheim, line 728: See above.

loewenheim (changes requested):
Good work! I have some notes, though, most of them stylistic.

src/commands/react_native/gradle.rs
```diff
@@ -98,22 +96,37 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     processor.rewrite(&[base.to_str().unwrap()])?;
     processor.add_sourcemap_references()?;
+    processor.add_debug_id_references()?;
 
     // TODO: make this optional
```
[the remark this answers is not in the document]
krystofwoldrich: Yes.

src/commands/react_native/xcode.rs
```diff
@@ -247,14 +275,24 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
                 );
                 panic!("{}", format_err);
             });
-            if report.bundle_path.is_none() || report.sourcemap_path.is_none() {
-                println!("Warning: build produced no sourcemaps.");
+            if report.packager_bundle_path.is_none() || report.packager_sourcemap_path.is_none() {
+                println!("Warning: build produced no packager sourcemaps.");
                 return Ok(());
             }
```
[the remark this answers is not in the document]
krystofwoldrich: Nice, I didn't know this syntax.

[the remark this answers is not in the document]
loewenheim: Yeah, it's fairly new :)

src/commands/react_native/xcode.rs
```diff
@@ -275,44 +313,61 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-                    dedupe: false,
-                    chunk_upload_options: chunk_upload_options.as_ref(),
-                })?;
-            }
-            Some(dists) => {
-                for dist in dists {
+        let dist_from_env = env::var("SENTRY_DIST");
+        let release_from_env = env::var("SENTRY_RELEASE");
+
+        if dist_from_env.is_err() && dist_from_env.is_err() && matches.get_flag("no_auto_release") {
```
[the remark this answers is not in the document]
krystofwoldrich: Yes, good catch.

src/commands/react_native/xcode.rs
```diff
@@ -321,10 +376,24 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+
+    let report_file_path = env::var("SENTRY_RN_SOURCEMAP_REPORT").unwrap();
+    let mut f;
+    if std::path::Path::new(&report_file_path).exists() {
+        f = fs::File::open(report_file_path.clone())?;
+    } else {
+        f = fs::File::create(report_file_path.clone())?;
+    }
+    let mut sourcemap_report: SourceMapReport = serde_json::from_reader(&mut f)
+        .unwrap_or_else(|_| SourceMapReport::default());
```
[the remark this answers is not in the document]
krystofwoldrich: Yeah, that's much cleaner.

src/commands/react_native/xcode.rs
```diff
@@ -344,31 +413,118 @@ pub fn wrap_call() -> Result<()> {
...
+        sourcemap_report.packager_bundle_path = bundle_path.map(PathBuf::from);
 
-    let rv = process::Command::new(env::var("SENTRY_RN_REAL_NODE_BINARY").unwrap())
+    // Hermes Compiler
+    // -emit-binary doesn't have to be first in order but all
+    // supported RN 0.65 to 0.72 have it as first argument
+    // and users can't change it
+    } else if args.len() > 1 && args[0] == "-emit-binary" {
+        execute_hermes_compiler = true;
+        let mut iter = args.iter().fuse();
```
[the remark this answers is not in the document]
loewenheim: Yes. `slice::Iter` implements `FusedIterator`, which means `fuse()` is a no-op.

src/commands/react_native/xcode.rs
```diff
@@ -344,31 +413,118 @@ pub fn wrap_call() -> Result<()> {
...
+                        hermes_sourcemap.insert("debug_id".to_string(), debug_id.clone());
+                    } else if let Some(debug_id) = packager_sourcemap.get("debug_id") {
+                        hermes_sourcemap.insert("debugId".to_string(), debug_id.clone());
+                        hermes_sourcemap.insert("debug_id".to_string(), debug_id.clone());
+                    } else {
+                        println!("No debug id found in packager source map, skipping copy to Hermes combined source map.");
+                    }
+
+                    hermes_sourcemap_file = fs::File::create(hermes_sourcemap_path)?;
+                    serde_json::to_writer(&mut hermes_sourcemap_file, &hermes_sourcemap)?;
```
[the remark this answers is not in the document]
krystofwoldrich: Thanks, no, it only make sense when we add the debug id.

src/utils/sourcemaps.rs
krystofwoldrich: The referenced source map. I changed the text to be more clear.

src/utils/sourcemaps.rs
krystofwoldrich: I was, but it doesn't make sense in general. Replaced by `continue`

src/utils/sourcemaps.rs
krystofwoldrich: Ohh, nice. I didn't know.

krystofwoldrich:
@loewenheim Thank you for all the comments, I updated everything besides the `fuse`, I assume it's unnecessary overhead? I just used it as it was in the other if-else branch.

loewenheim:
> @loewenheim Thank you for all the comments, I updated everything besides the `fuse`, I assume it's unnecessary overhead? I just used it as it was in the other if-else branch.

I don't even think it's overhead at runtime, it should literally do nothing.

loewenheim (approved):

PR #1688, opened 2023-07-19
feat(proguard): Release association

buenaflor:
Creates a weak release association with a proguard mapping file. https://github.com/getsentry/team-mobile/issues/27

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -20,7 +20,6 @@ use backoff::backoff::Backoff;
 use brotli2::write::BrotliEncoder;
 use chrono::{DateTime, Duration, FixedOffset, Utc};
 use clap::ArgMatches;
-use console::style;
 use flate2::write::GzEncoder;
 use if_chain::if_chain;
 use lazy_static::lazy_static;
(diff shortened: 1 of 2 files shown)


loewenheim (approved):
Looks good, apart from lint stuff :)

src/api.rs
```diff
@@ -1354,6 +1354,28 @@ impl Api {
...
+        let path = format!(
+            "/projects/{}/{}/files/proguard-artifact-releases",
+            PathArg(org),
+            PathArg(project)
+        );
+        let resp: ApiResponse = self
+            .request(Method::Post, &path)?
+            .with_json_body(data)?
+            .send()?;
+        if resp.status() == 404 || resp.status() == 201 {
```
kamilogorek, line 1372: Why are we doing `Ok(())` on `404`? 👀

src/api.rs
```diff
@@ -1361,16 +1361,18 @@ impl Api {
         data: &AssociateProguard,
     ) -> ApiResult<()> {
         let path = format!(
-            "/projects/{}/{}/files/proguard-artifact-releases",
+            "/projects/{}/{}/files/proguard-artif",
```
kamilogorek, line 1364: Whoops

src/api.rs
buenaflor: Shouldn't be there for sure!

kamilogorek (approved):

PR #1697, opened 2023-07-27
Allow the CLI binary path to be overridden

koddsson:
Hey friends 👋🏻 

What do you think about allowing me to provide my own Sentry CLI binary location?


kamilogorek:
This might be actually a better way to approach this than using `SENTRYCLI_USE_LOCAL` during installation.

kamilogorek (approved):

PR #1716, opened 2023-08-11
Fix integration tests runner

kamilogorek:
Minor fixes for running in all environments + `CONTRIBUTING.md`


tests/integration/_cases/send_event/send_event-raw.trycmd
```diff
@@ -13,8 +13,9 @@ $ sentry-cli send-event --log-level=debug
 > --fingerprint custom-fingerprint
 > --no-environ
 ? success
-  DEBUG   [..] sentry-cli version: [VERSION], platform: [..], architecture: [..]
-  INFO    [..] sentry-cli was invoked with the following command line: "[CWD]/target/debug/sentry-cli[EXE]" "send-event" "--log-level=debug" "--level" "debug" "--timestamp" "1649335000929" "--release" "my-release" "--dist" "my-dist" "--env" "production" "--message" "hello" "--platform" "prod" "--tag" "hello:there" "--extra" "hello:there" "--user" "id:42" "--fingerprint" "custom-fingerprint" "--no-environ"
+  INFO    [..] Loaded config from [CWD]/.sentryclirc
+  DEBUG   [..] sentry-cli version: [VERSION], platform: "darwin", architecture: "arm64"
```
Swatinem, line 17: you should still mask the platform and architecture for these tests, they are failing for me as I am still on `x86_64`.

src/utils/vcs.rs
```diff
@@ -915,6 +915,7 @@ fn git_create_commit(dir: &Path, file_path: &str, content: &[u8], commit_message
             "--author",
             "John Doe <john.doe@example.com>",
             "--quiet",
+            "--no-edit",
```
Swatinem, line 918: For me, its rather `utils::vcs::test_find_matching_rev_with_lightweight_tag` that opens a vim window and then fails because `fatal: no tag message?`.

Maybe my local git is configured to always expect a message for tags, not sure

.sentryclirc
```diff
@@ -0,0 +1 @@
+; this stub prevents global .sentryclirc from being loaded
```
Swatinem, line 1: The `integration::info::command_info_no_token` test is still failing for be because it is picking up credentials from my global config.

src/utils/vcs.rs
Swatinem: I have this in my global git config:

```
[commit]
    gpgsign = true
[tag]
    gpgsign = true
```

Looks like you can only sign tags if they have a message.
Maybe an explicit command line switch to avoid signing them, or provide a default message (aka `-m=$version`) would solve this.

Swatinem (approved):

PR #1794, opened 2023-11-01
chore: reduce realease binary size

vaind:
Sentry CLI is downloaded many times during consumers' CI. Even though the download should be fast in general and the binary size reduction wouldn't have a great effect on individual runs, it does have an effect at the scale it is used at.

Example: macOS arm64 binary size reduction by these changes is about 30 %.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,8 +111,9 @@ crossbeam-channel = "0.5.6"
 [target."cfg(windows)".dependencies]
 winapi = "0.3.9"
 
+# We optimize the release build for size.
 [profile.release]
-opt-level = 3
+opt-level = "z"
 panic = "abort"
 strip = true      # Strip symbols from the binary.
 codegen-units = 1 # Parallel compilation prevents some optimizations.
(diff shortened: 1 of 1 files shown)


loewenheim:
I've been experimenting along similar lines. In my test, for whatever reason, setting codegen-units to 1 actually slightly increased binary size, have you checked whether it has a positive effect?

Also, setting the opt level to `"s"` reduces the size significantly. It carries a performance hit, but that is likely negligible.

Swatinem (approved):
having only one CGU means that the compiler has more inlining opportunities, so probably generates more code.
however with CGU=1 and LTO, it should generate fewer duplicates as well.

vaind:
For me (arm64 macos) the build is smaller with CGU=1 by about 100 KiB.

Setting the opt-level to S reduces the size by further 2 MiB and going for Z means another MiB down.

vaind:
For future reference, here's the size reduction for a release containing this PR (between 25 and 55 % size reduction):
image

PR #1795, opened 2023-11-02
ci: run cargo build on all branches

vaind:
Running the same build workflow on all branches/PRs makes sure you we catch potential errors sooner than on the release branch.

For example: https://github.com/getsentry/sentry-cli/actions/runs/6731317309/job/18295737112#step:3:344 (``error: package `clap_complete v4.4.3` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.69.0``)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,13 @@
-name: Release Build
+name: Build
 
 on:
   push:
     branches:
+      - master
+      - 1.x
       - release/**
+
+  pull_request:
 
 jobs:
   linux:


ashwoods:
This would add aprox 14m to each PR. I think it would be better to harmonize our debug and release builds.  We will take a look into this.

vaind:
> This would add aprox 14m to each PR. I think it would be better to harmonize our debug and release builds. We will take a look into this.

It does not necessarily mean to be a blocker (required check) for the PR to be merged so if you really are in a hurry you could merge the PR before the build ends. On the other hand, for most PRs (i.e. those that don't get merged within 15 minutes of the last commit), it would be a good test that the build isn't getting broken by the PR.

PR #1817, opened 2023-11-13
feat: Expose `downloadBinary` function to install binary

lforst:
We would like to be able to install the CLI manually in downstream dependencies in case postinstall scripts did not get executed.

### DIFF
--- a/js/install.js
+++ b/js/install.js
@@ -37,7 +37,7 @@ function getLogStream(defaultStream) {
   );
 }
 
-const logger = new Logger(getLogStream('stderr'));
+const ttyLogger = new Logger(getLogStream('stderr'));
 
 const CDN_URL =
   process.env.SENTRYCLI_LOCAL_CDNURL ||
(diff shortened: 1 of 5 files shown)


Lms24 (approved):
Nice! I believe this was the actual download logic is more or less C&P from what we had before, correct? 

(fwiw, TS would probably make sense but let's not overcomplicate things)

lforst:
> I believe this was the actual download logic is more or less C&P from what we had before, correct?

Correct. I didn't change anything except for adding the logger thing.

js/install.js
```diff
@@ -0,0 +1,305 @@
...
+      throw new Error(`Unexpected sentry-cli version "${version}", expected "${expected}"`);
+    }
+  });
+}
+
+function downloadBinary(logger = ttyLogger) {
+  if (process.env.SENTRYCLI_SKIP_DOWNLOAD === '1') {
+    logger.log(`Skipping download because SENTRYCLI_SKIP_DOWNLOAD=1 detected.`);
+    return;
+  }
```
Jonir227, line 199: this should return Promise. it makes error on flag 1

lforst: Ah right. Good catch!

Issue #1360, opened 2022-10-15
npm cli install without postinstall


timfish:
`postinstall` creates a whole host of issues which it would be nice to not have to deal with.

I've seen a similar issues when distributing native node binaries and wonder whether some of the solutions I've seen in use there might be an improvement here.

Firstly, you publish an npm package per platform/arch combo (`@sentry/cli-{platform}-{arch}`) and include the cli binary in there. 
- In `package.json` set `main` as the path to the binary and then it can be found later by simply `require.resolve('@sentry/cli-${process.platform}-${process.arch}')` 
- In `package.json` correctly define `os` and `cpu` properties which ensures that they will only install on the correct platform/arch
- You can even set `bin` in these packages if you want them to be directly accessible.

Then in `@sentry/cli` you list all the above as `optionalDependencies`. After this, when installing `@sentry/cli`, npm/yarn will attempt to install all the optional dependencies and only succeed on those where the `os` and `cpu` match. 

Bonus: If your JavaScript code used to find the binary uses complicated enough dynamic paths construction, `@vercel/nft` gives up and wont include them.

This approach is already used by [Parcel](https://www.npmjs.com/package/@parcel/css-darwin-arm64), [Next.js](https://www.npmjs.com/package/@next/swc-darwin-arm64) and [plenty of others](https://www.npmjs.com/search?q=darwin-arm64).

### Downsides?
- It is perhaps abusing `optionalDependencies` but everyone appears to be on this train now so maybe it wont crash?
- Does npm/yarn have to download all the optional dependencies before checking the `os` and `cpu`? 🤔

kamilogorek:
ref: https://github.com/npm/rfcs/blob/528fa298a3529dcae3f58cbf2f9a577ba16b479f/accepted/0000-package-distributions.md

me4502:
Given the size of the sentry-cli binary, in poor network conditions this can be a substantial issue on CI. Currently this breaks the Yarn caching & install model (https://yarnpkg.com/advanced/lifecycle-scripts#a-note-about-postinstall), leading to the cache being bypassed and the entire binary being downloaded on every run.

Would definitely be good to not have postinstall used to download a large binary

krystofwoldrich:
- linking related issue https://github.com/getsentry/sentry-react-native/issues/2688

lforst:
Plan of action:
- [ ] Create packages for all different architectures (just the package.json) @lforst 
[...]

PR #1828, opened 2023-11-20
build: Add placeholder packages for npm binary distributions

lforst:
Adds packages for binary distributions on npm based on the binaries we currently distribute.

Ref https://github.com/getsentry/sentry-cli/issues/1360#issuecomment-1819270364

### DIFF
--- a/npm-binary-distributions/darwin/package.json
+++ b/npm-binary-distributions/darwin/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@sentry/cli-darwin",
   "version": "2.21.5",
-  "description": "TODO",
+  "description": "The darwin distribution of the Sentry CLI binary.",
   "repository": "https://github.com/getsentry/sentry-cli",
   "license": "BSD-3-Clause",
   "engines": {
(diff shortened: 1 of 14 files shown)


szokeasaurusrex (approved):
Looks good (I am trusting that you have checked that the `package.json` are correct; since I am not super familiar with how these work)

npm-binary-distributions/darwin/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@sentry/cli-darwin",
```
szokeasaurusrex, line 2: Do we need to have a separate binary for M1/M2 and Intel Macs?

lforst: The CLI is currently set up to use our universal darwin binary no matter what architecture.

https://github.com/getsentry/sentry-cli/blob/1c5d20f94043ac66631157b2a208d5668541921c/scripts/install.js#L78-L80

Also, I believe mac OS has a compatibility layer for non-arm binaries. But good that you confirmed!

szokeasaurusrex: Yeah, but I believe the Rosetta compatibility layer you are referencing imposes a performance penalty, and also, if it is the first time the user is running a non-arm binary, the user is prompted to install Rosetta and needs to wait some time for the install to complete. These are probably only minor inconveniences, but we should probably be creating a native ARM build (as we already are with the universal binary)

PR #1833, opened 2023-11-23
feat: Publish binary distributions on npm

lforst:
Sets up CI to upload the tarballs of the sentry CLI binary distributions for different architectures on release branches.

Ref: https://github.com/getsentry/sentry-cli/issues/1360#issuecomment-1819270364

We will verify this with an alpha release.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -11,7 +11,35 @@ targets:
       - path: /sentry-cli/latest/
         metadata:
           cacheControl: public, max-age=600
+
+  # Main Sentry CLI package
   - name: npm
+    id: '@sentry/cli'
+    includeNames: /^sentry-cli-\d.*\.tgz$/
+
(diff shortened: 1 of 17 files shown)


lforst:
Ok this seems to work: https://github.com/getsentry/sentry-cli/actions/runs/6970345254/job/18968277409

AbhiPrasad (commented):
Should we create the binary distributions before we do main sentry-cli package.

Also cc @JonasBa, we should use the same approach for profiling-node

lforst:
> Should we create the binary distributions before we do main sentry-cli package.

@AbhiPrasad I assume you mean in the craft config? Good point. Thanks!

AbhiPrasad (approved):
> I assume you mean in the craft config? Good point. Thanks!

yup - thanks for making the change!

PR #1834, opened 2023-11-23
build: Add optional dependencies to main cli package for binary distributions

lforst:
Adds optional dependencies to the main CLI package.

Also adjusts the bump version script to bump the version of these optional dependencies.

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -22,3 +22,9 @@ cargo update -p sentry-cli
 # Do not tag and commit changes made by "npm version"
 export npm_config_git_tag_version=false
 npm version "${TARGET}"
+
+for dir in $SCRIPT_DIR/../npm-binary-distributions/*; do
+    cd $dir
+    npm version "${TARGET}"
+    cd -
+done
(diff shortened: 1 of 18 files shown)


scripts/bump-version.sh
```diff
@@ -21,10 +21,17 @@ cargo update -p sentry-cli
...
+# Bump the binary npm distributions
 for dir in $SCRIPT_DIR/../npm-binary-distributions/*; do
     cd $dir
     npm version "${TARGET}"
     cd -
 done
+
+# Update the optional deps in the main cli npm package
+# Requires jq to be installed - should be installed ootb on github runners
+jq '.optionalDependencies |= map_values("'"${TARGET}"'")' $SCRIPT_DIR/../package.json > package.json.tmp && mv package.json.tmp $SCRIPT_DIR/../package.json
```
AbhiPrasad, line 37: tbh I'd rather we convert this to a node script but 🤷

AbhiPrasad (approved):

scripts/bump-version.sh
lforst: Yeah I also feel conflicted about this, at the same time it felt weird to either call node from this script, or alternatively it felt also weird converting this script into node bc we run cargo & perl 💀

PR #1835, opened 2023-11-23
feat: Add opt-in code path to use binary distributions instead of downloaded CLI

lforst:
Adds a code path to pick up the npm distribution binaries when executing the CLI.

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -92,7 +92,12 @@ function getBinaryPath() {
 
   if (packageName === undefined) {
     throw new Error(
-      'Unsupported operating system or architecture! Sentry CLI does not work on this architecture.'
+      `Unsupported operating system or architecture! Sentry CLI does not work on this architecture.
+
+Sentry CLI supports:
+- Darwin (macOS)
+- Linux and FreeBSD on x64, x86, ia32, arm64, and arm architectures
+- Windows x64, x86, and ia32 architectures`
     );
   }
 
--- a/js/helper.js
+++ b/js/helper.js
@@ -1,7 +1,71 @@
 'use strict';
 
+const os = require('os');
 const path = require('path');
 const childProcess = require('child_process');
+
+const BINARY_DISTRIBUTIONS = [
+  { packageName: '@sentry/cli-darwin', subpath: 'bin/sentry-cli' },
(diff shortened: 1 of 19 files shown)


js/helper.js
```diff
@@ -16,11 +80,50 @@ function getBinaryPath() {
...
+      }
+    });
+
+    // These error messages are heavily inspired by esbuild's error messages: https://github.com/evanw/esbuild/blob/f3d535262e3998d845d0f102b944ecd5a9efda57/lib/npm/node-platform.ts#L150
+    if (otherInstalledDistribution) {
+      throw new Error(`Sentry CLI binary for this platform/architecture not found!
+
+The "${otherInstalledDistribution.packageName}" package is installed, but for the current platform, you should have the "${packageName}" package installed instead. This usually happens if the "@sentry/cli" package is installed on one platform (for example Windows or MacOS) and then the "node_modules" folder is reused on another operating system (for example Linux in Docker).
+
+To fix this, avoid copying the "node_modules" folder, and instead freshly install your dependencies on the target system. If you know what you are doing, you can also configure your package manager to install the right package. For example, yarn has the "supportedArchitectures" feature: https://yarnpkg.com/configuration/yarnrc/#supportedArchitecture.`);
```
AbhiPrasad, line 118:
```suggestion
To fix this, avoid copying the "node_modules" folder, and instead freshly install your dependencies on the target system. You can also configure your package manager to install the right package. For example, yarn has the "supportedArchitectures" feature: https://yarnpkg.com/configuration/yarnrc/#supportedArchitecture.`);
```

js/helper.js
```diff
@@ -16,11 +80,50 @@ function getBinaryPath() {
...
+    parts.push('..');
+    parts.push(`sentry-cli${process.platform === 'win32' ? '.exe' : ''}`);
+    return path.resolve(...parts);
+  }
+
+  const { packageName, subpath } = getDistributionForThisPlatform();
+
+  if (packageName === undefined) {
+    throw new Error(
+      'Unsupported operating system or architecture! Sentry CLI does not work on this architecture.'
```
AbhiPrasad, line 95: Can we list all the architectures we do support here?

AbhiPrasad (approved):

PR #1858, opened 2023-12-06
fix: install the x64 cli on windows-on-arm hosts

MarshallOfSound:
As in title, currently installing the CLI errored out on windows-on-arm hosts because it tries to install a build that I guess doesn't exist.  This is a temporary workaround until ideally and arm64 build for windows is published.


lforst:
This will likely not be enough (unless you're manually using the install script). We likely need to allow/publish this in our npm distribution packages.

Issue #1861, opened 2023-12-06
Switch from libgit2 to gix


scefali:
@asottile-sentry described a bug I encountered using the CLI through Docker [here](https://github.com/libgit2/libgit2/issues/6663)

Swithing to [gix](https://crates.io/crates/gix) would resolve this problem and avoid having to add this hack to Docker to make the CLI be able to detect the current git sha:

```docker
ENV SUDO_UID=1001
```

szokeasaurusrex:
Closed in favor of #1866

PR #1866, opened 2023-12-13
ref: upgrade git2 so safe.directory is implemented more correctly

asottile-sentry:
perhaps easier than  #1861


szokeasaurusrex (approved):
~~perhaps~~ _definitely_ easier than https://github.com/getsentry/sentry-cli/issues/1861

Thanks for suggesting this fix! Looks good to me

Issue #1849, opened 2023-11-30
Vercel build errors in runtime due to Sentry CLI binary for arch not found


mcmxcdev:
### Environment

`@sentry/cli` v2.22.3 on Ubuntu 23.10

### Steps to Reproduce

When attempting to upgrade `@sentry/sveltekit` from 7.76.0 to 7.83.0, we encounter this runtime error on a Vercel preview deployment URL:

```text
2023-11-29T18:38:58.472Z	undefined	ERROR	Error: Sentry CLI binary for this platform/architecture not found!
It seems like none of the "@sentry/cli" package's optional dependencies got installed. Please make sure your package manager is configured to install optional dependencies. If you are using npm to install your dependencies, please don't set the "--no-optional" or "--omit=optional" flags. Sentry CLI needs the "optionalDependencies" feature in order to install its binary.
    at getBinaryPath (/var/task/node_modules/.pnpm/@sentry+cli@2.22.3/node_modules/@sentry/cli/js/helper.js:116:13)
    at Object.<anonymous> (/var/task/node_modules/.pnpm/@sentry+cli@2.22.3/node_modules/@sentry/cli/js/helper.js:129:18)
    at Module._compile (node:internal/modules/cjs/loader:1241:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
    at r.<computed>.e._load (/var/task/___vc/__launcher/__launcher.js:14:2516)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (/var/task/node_modules/.pnpm/@sentry+cli@2.22.3/node_modules/@sentry/cli/js/index.js:4:16)
INIT_REPORT Init Duration: 977.05 ms	Phase: invoke	Status: error	Error Type: Runtime.ExitError
RequestId: 9b0d05a2-8a1e-4975-b480-92d551a0cf9e Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

The issue is most likely related to https://github.com/getsentry/sentry-cli/releases/tag/2.22.3

The package manager in use is pnpm, optional dependencies are not disabled, and the `.npmrc` file looks as follows:

```
engine-strict=true
enable-pre-post-scripts=true
strict-peer-dependencies=true
update-notifier=false
```

[...]

PR #1874, opened 2023-12-27
fix: Manually download binary if optional dependency binary can't be found after installation

lforst:
Fixes https://github.com/getsentry/sentry-cli/issues/1849 to some degree.

With this PR we re-add a post-install script that will manually download the Sentry CLI binary from our CDN if we cannot find the npm binary distributions that would be installed via optional dependencies.

This will improve the reliability for situations where optional dependencies are not installed (like Vercel apparently).

We still require either postinstall scripts to be enabled, OR optional dependencies to be installed.

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -154,7 +154,7 @@ It seems like none of the "@sentry/cli" package's optional dependencies got inst
   return compatibleBinaryPath;
 }
 
-let mockedBinaryPath = undefined;
+let mockedBinaryPath;
 
 /**
  * Overrides the default binary path with a mock value, useful for testing.
--- a/js/helper.js
+++ b/js/helper.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const os = require('os');
+const path = require('path');
+const fs = require('fs');
 const childProcess = require('child_process');
 
 const BINARY_DISTRIBUTIONS = [
@@ -12,6 +14,23 @@ const BINARY_DISTRIBUTIONS = [
   { packageName: '@sentry/cli-win32-x64', subpath: 'bin/sentry-cli.exe' },
   { packageName: '@sentry/cli-win32-i686', subpath: 'bin/sentry-cli.exe' },
 ];
+
+/**
+ * This convoluted function resolves the path to the manually downloaded fallback
+ * `sentry-cli` binary in a way that can't be analysed by @vercel/nft.
+ *
+ * Without this, the binary can be detected as an asset and included by bundlers
+ * that use @vercel/nft.
+ *
+ * @returns {string} The path to the sentry-cli binary
+ */
+function getFallbackBinaryPath() {
+  const parts = [];
+  parts.push(__dirname);
+  parts.push('..');
+  parts.push(`sentry-cli${process.platform === 'win32' ? '.exe' : ''}`);
+  return path.resolve(...parts);
+}
 
 function getDistributionForThisPlatform() {
   const arch = os.arch();
@@ -67,11 +86,25 @@ function getDistributionForThisPlatform() {
 }
 
 /**
- * This convoluted function resolves the path to the `sentry-cli` binary in a
- * way that can't be analysed by @vercel/nft.
+ * Throws an error with a message stating that Sentry CLI doesn't support the current platform.
  *
- * Without this, the binary can be detected as an asset and included by bundlers
- * that use @vercel/nft.
+ * @returns {never} nothing. It throws.
+ */
+function throwUnsupportedPlatformError() {
+  throw new Error(
+    `Unsupported operating system or architecture! Sentry CLI does not work on this architecture.
+
+Sentry CLI supports:
+- Darwin (macOS)
+- Linux and FreeBSD on x64, x86, ia32, arm64, and arm architectures
+- Windows x64, x86, and ia32 architectures`
+  );
+}
+
+/**
+ * Tries to find the installed Sentry CLI binary - either by looking into the relevant
+ * optional dependencies or by trying to resolve the fallback binary.
+ *
  * @returns {string} The path to the sentry-cli binary
  */
 function getBinaryPath() {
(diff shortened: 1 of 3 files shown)


adinauer (approved):
LGTM

js/helper.js
```diff
@@ -13,6 +15,23 @@ const BINARY_DISTRIBUTIONS = [
...
+ * that use @vercel/nft.
+ *
+ * @returns {string} The path to the sentry-cli binary
+ */
+function getFallbackBinaryPath() {
+  const parts = [];
+  parts.push(__dirname);
+  parts.push('..');
+  parts.push(`sentry-cli${process.platform === 'win32' ? '.exe' : ''}`);
+  return path.resolve(...parts);
```
anonrig, line 32:
```suggestion
  return path.resolve(__dirname, '..', `sentry-cli${process.platform === 'win32' ? '.exe' : ''}`);
```

js/helper.js
```diff
@@ -13,6 +15,23 @@ const BINARY_DISTRIBUTIONS = [
...
+/**
+ * This convoluted function resolves the path to the manually downloaded fallback
+ * `sentry-cli` binary in a way that can't be analysed by @vercel/nft.
+ *
+ * Without this, the binary can be detected as an asset and included by bundlers
+ * that use @vercel/nft.
+ *
+ * @returns {string} The path to the sentry-cli binary
+ */
+function getFallbackBinaryPath() {
```
anonrig, line 27: Does this need to be a function? Can we convert this to a variable?

js/helper.js
```diff
@@ -122,11 +154,7 @@ It seems like none of the "@sentry/cli" package's optional dependencies got inst
   return compatibleBinaryPath;
 }
 
-/**
- * Absolute path to the sentry-cli binary (platform dependent).
- * @type {string}
- */
-let binaryPath = getBinaryPath();
+let mockedBinaryPath = undefined;
```
anonrig, line 157: Same:

```suggestion
let mockedBinaryPath;
```

js/helper.js
```diff
@@ -222,7 +250,7 @@ function prepareCommand(command, schema, options) {
  * @returns {string}
  */
 function getPath() {
-  return binaryPath;
+  return mockedBinaryPath !== undefined ? mockedBinaryPath : getBinaryPath();
```
anonrig, line 253:
```suggestion
  return mockedBinaryPath || getBinaryPath();
```

package.json
```diff
@@ -39,6 +39,7 @@
     "@sentry/cli-win32-x64": "2.23.1"
   },
   "scripts": {
+    "postinstall": "node ./scripts/install.js",
```
anonrig, line 42: Since we have `usr/bin/env node` in the first line of the script, we don't need `node` here:

```suggestion
    "postinstall": "./scripts/install.js",
```

scripts/install.js
```diff
@@ -323,14 +317,30 @@ if (process.env.SENTRYCLI_SKIP_DOWNLOAD === '1') {
...
+}
+
+try {
+  require.resolve(`${distributionPackageName}/${distributionSubpath}`);
+  // If the `resolve` call succeeds it means a binary was installed successfully via optional dependencies so we can skip the manual postinstall download.
+  process.exit(0);
+} catch (e) {
+  // Optional dependencies likely didn't get installed - proceed with fallback downloading manually
+  logger.log(
+    `Sentry CLI binary installation via optional dependencies was unsuccessful. Downloading manually instead.`
```
anonrig, line 334: Should we at least log the error here?

scripts/install.js
```diff
@@ -323,14 +317,30 @@ if (process.env.SENTRYCLI_SKIP_DOWNLOAD === '1') {
...
+  );
+
+  downloadBinary()
+    .then(() => checkVersion())
+    .then(() => {
+      process.exit(0);
+    })
+    .catch((e) => {
+      console.error(e);
```
anonrig, line 343: Should this be `logger.error`? I think we also need `eslint-disable-next-line no-console` in here.

js/helper.js
lforst: As per the comment above this function is convoluted like this so that static code analyzers (like [Vercel's nft](https://github.com/vercel/nft), and pretty much only Vercel's nft) cannot detect the binary as a "dependency" (dependency here meaning a thing that needs to be included in an AWS lambda for the application to work), because generally speaking Sentry CLI is not a runtime dependency but one that is only required during builds.

js/helper.js
lforst: See other comment. This function is just this way so that Vercel's nft cannot resolve the binaries location.

package.json
lforst: I am not sure if the shebang will work on windows. Both SWC and esbuild have their postinstall scripts defined like this and it doesn't seem it's like that by accident so I will keep it.

js/helper.js
lforst: I am intentional with the undefined check here because of empty strings.

scripts/install.js
lforst: The error would just be Node's default "Cannot resolve module `@sentry/cli-whatever-architecture`" which is not helpful at all here - so likely not. During runtime (which is the actually important part because we cannot control what happens between installation and runtime - people do random shit like moving the platform specific binary from Windows to Debian in Docker) we will print a more specific error message with better instructions.

package.json
lforst: Actually, see https://github.com/getsentry/sentry-cli/pull/1151

austinphan-multiplier: @lforst It break in Amplify environment
```
npm ERR! path /root/.nvm/versions/node/v16.20.2/lib/node_modules/@sentry/cli
npm ERR! command failed
npm ERR! command sh -c -- node ./scripts/install.js
npm ERR! sh: node: command not found
```

Issue #1859, opened 2023-12-06
Improve error message if not full token is not copied (Checksum correct)


HazAT:
<img width="893" alt="image" src="https://github.com/getsentry/sentry-cli/assets/363802/577758ca-263e-4028-9048-a991514cdea5">

If you fail to copy the full token - we should improve the error message to tell users they fucked up copying the full token (because checksum check failed)

szokeasaurusrex:
We also need to make sure we validate the token on the client (or server) before sending

szokeasaurusrex:
@HazAT I proposed a simpler change in #1876 that would add the more specific error message when a 401 status code is received from the server. Adding the error message based on the HTTP status allows us to skip parsing and validating the auth token client-side, but of course, we might end up adding the message even when the auth token was correctly copied, but some other problem occurred (e.g. token was revoked). Since the error message I wrote only instructs users to make sure they copied the token correctly, I think it is okay to add the message even when the error has a different cause.

What do you think -- does #1876 accomplish what you wanted, or should we add the client-side validation for the auth token?

HazAT:
No, I would not do it like this: `401` represents another error and showing only this message for `401` could confuse users even more.
If the token is malformed can and should be validated by the client and has nothing to do with the return code.

szokeasaurusrex:
Good point @HazAT – let's implement the validation per the original plan in that case

szokeasaurusrex:
Reopening, since we needed to revert #1885 to fix #1888 temporarily, while we figure out what is causing #1888.

PR #1876, opened 2023-12-27
feat(api): More specific error message for 401 Sentry errors

szokeasaurusrex:
Add a line to the error message output when a 401 error occurs when communicating with Sentry to instruct users to verify that the auth token was copied correctly. An incorrectly copied auth token is a likely cause of a 401 error.

Fixes GH-1859


szokeasaurusrex:
Closed in favor of adding validation, per https://github.com/getsentry/sentry-cli/issues/1859#issuecomment-1872082427

Issue #1863, opened 2023-12-11
Source mapping is offset when using switch statements


richardsimko:
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Which SDK are you using?

@sentry/node

### SDK Version

7.86.0

### Link to Sentry event

https://fl101.sentry.io/issues/4715851450/events/4f6786772d4045d4b57a56cab225e50e/

### SDK Setup

N/A

### Steps to Reproduce

https://github.com/richardsimko/sentry-sourcemap-bug

1. `yarn install`
2. `yarn run tsc`
3. `yarn run sentry-cli sourcemaps inject ./build`
4. `yarn run sentry-cli sourcemaps resolve build/test.js.map  -l 6` (To simulate the lookup from one of the errors)

This happens in the live Sentry environment as well, the stack trace in the linked event are also off by 2 lines.

### Expected Result

Source line is line 7 in `test.ts`

### Actual Result

Source line is line 5 in `test.ts`

```
source map path: "build/test.js.map"
source map type: regular
lookup line: 5, column: 0:
  name: not found
  source file: "/test.ts"
  source line: 5
  source column: 13
  minified line: 4
  minified column: 15
  source code:
    console.log("test2");
    
    switch (process.env.TEST) {
      case "test":
        throw new Error("test");
      default:
        throw new Error("default");
```

richardsimko:
@lforst Thanks for transferring the issue. Although I'm not sure if it's a sentry-cli issue though since the problem appears in the stack traces we send to Sentry as well, I was just presuming the same logic is used in the CLI and that's the reason why the issue appears here as well.

lforst:
@richardsimko the SDK merely sends raw stack traces. If it is resolved wrongly there are 3 possibilities:

- The CLI `sourcemaps inject` command messed up the source map/generated file.
- The CLI `sourcemaps resolve` command is not working correctly.
- The Processing in Sentry's backend is wrong.
- `tsc` is generating faulty source maps.

Your repro also seems a bit off. If I try to reproduce, everything seems to resolve fine. Can you share in a bit more detail what exactly you would expect?

The only thing I noticed is that we may have an off-by-one in the cli output.

[...]

PR #1880, opened 2023-12-29
Improve `sourcemaps resolve` command output

szokeasaurusrex:
This PR makes two improvements to the `sourcemaps resolve` command output:
  - The PR changes the output line and column values from zero-indexed to one-indexed values. By switching to one-indexed values, the output line and column numbers use the same scale as the input values and follow the standard convention for line and column numbers in source code files. This change fixes the off-by-one error reported in #1863. 
  - The output text has been restructured and clarified to indicate that the sourcemap resolver locates the closest token to the line and column input that the user provides. With this change, it should be clear to users that the source and minified file line/column locations reported in the output are the locations of this nearest token that the sourcemap resolver locates. This change is specifically intended to help avoid confusion in cases where the nearest token may in a different location (perhaps even, on a different line) than the search location provided in the user's input.

# Example
To demonstrate this change, we include an example below to demonstrate how the command output has changed. We use the sourcemap from the "Steps to Reproduce" in https://github.com/getsentry/sentry-cli/issues/1863#issue-2035498391.

### Example command
```bash
$ sentry-cli sourcemaps resolve /path/to/example/sourcemap.js.map -l 6
```

### Old output
```
source map path: "[...]"
source map type: regular
lookup line: 5, column: 0:
  name: not found
  source file: "/test.ts"
  source line: 5
  source column: 13
  minified line: 4
  minified column: 15
  source code:
    console.log("test2");
    
    switch (process.env.TEST) {
      case "test":
        throw new Error("test");
      default:
        throw new Error("default");
```

### New output
```
source map path: "[...]"
source map type: regular

Searching for token nearest to line 6, column 1 in the minified file:

  Found the nearest token (unnamed) at line 5, column 16 in the minified file.

        - The same token is located at line 6, column 14 in source file /test.ts.


  Source code:
    console.log("test2");
    
    switch (process.env.TEST) {
      case "test":
        throw new Error("test");
      default:
        throw new Error("default");
```

Fixes GH-1863


Swatinem (approved):
🎉 1-based offsets is so much better

If you want to be even fancier, you could use so colors/format to bold the numbers or something, but I believe things are already much better.

PR #1881, opened 2023-12-30
Add "--environment" option to "monitors run" (supplement "-e" shorthand)

skalee:
There is `-e` option in `sentry-cli monitors run`, which allows to pass environment name. However, it was missing any non-shorthand variant. This pull request fixes this inconsistency by adding `--environment` as an alternative name for this option.

### Rationale

Some people prefer full option names for their explicitness, especially when writing scripts. All the other shorthand options for `monitors run`, i.e. `-s` and `-h`, have their non-shorthand variants, i.e. `--schedule` and `--help`, respectively.

### DIFF
--- a/src/commands/monitors/run.rs
+++ b/src/commands/monitors/run.rs
@@ -28,6 +28,7 @@ pub fn make_command(command: Command) -> Command {
         .arg(
             Arg::new("environment")
                 .short('e')
+                .long("environment")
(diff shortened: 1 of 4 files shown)


szokeasaurusrex (approved):
Looks good! I added a test for your changes and updated an existing test to reflect the new argument.

PR #1883, opened 2024-01-02
ref: Removed unneeded identity `map`

szokeasaurusrex:
Fixes Clippy error that was showing up in the linting CI action


lforst (approved):

lforst:
We should still pin clippy somehow so that CI doesn't go red out of the blue.

szokeasaurusrex:
> We should still pin clippy somehow so that CI doesn't go red out of the blue.

Please see #1884

PR #1885, opened 2024-01-03
feat(api): Validate Auth Tokens client-side

szokeasaurusrex:
Perform client-side validation of all Auth Tokens input to the Sentry CLI. To ensure future-compatibility, we only provide soft validation, meaning that we only print a warning message if we detect that the Auth Token is invalid – the CLI will still proceed with normal execution if the Auth Token is invalid. 

An example warning message generated by the CLI is copied below:
```
  WARN    2024-01-03 18:52:53.810662 +01:00 Unrecognized auth token format!
        Hint: Did you copy your token correctly?
```

Fixes GH-1859


src/config.rs
```diff
@@ -13,54 +13,27 @@ use lazy_static::lazy_static;
...
 use crate::constants::{CONFIG_RC_FILE_NAME, DEFAULT_RETRIES, DEFAULT_URL};
+use crate::utils::auth_token::AuthToken;
+use crate::utils::auth_token::AuthTokenPayload;
 use crate::utils::http::is_absolute_url;
 
 /// Represents the auth information
 #[derive(Debug, Clone)]
 pub enum Auth {
     Key(String),
+    #[deprecated = "Use Auth::AuthToken instead"]
```
Swatinem, line 28: Using `#[deprecated]` for completely internal things (we do not publish `sentry-cli` as a crate) is not really useful.

src/utils/auth_token/org_auth_token.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Constructs a new OrgAuthToken from a string. Returns an error if the string is not a valid org auth token.
+    fn construct_from_string(auth_string: String) -> Result<OrgAuthToken> {
+        if !auth_string.starts_with(ORG_AUTH_TOKEN_PREFIX) {
+            return Err(AuthTokenParseError);
+        }
+
+        let mut segment_split = auth_string.split('_');
+        segment_split.next(); // Skip the prefix
```
Swatinem, line 56: you should probably still validate it though

src/utils/auth_token/auth_token.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+impl<'a> From<&'a AuthToken> for &'a str {
+    /// Retrieves a reference to the auth token string from an AuthToken.
+    fn from(auth_token: &'a AuthToken) -> Self {
+        let ref inner = auth_token.0;
+        inner.into()
+    }
+}
```
Swatinem, line 38: This `From` impl is very weird, and you usually do these using a `as_str` fn.
On the other way around, you should probably have a `From<&str> for AuthToken` as well.

src/config.rs
szokeasaurusrex: > we do not publish `sentry-cli` as a crate

Did not realize that it wasn't published, so I thought this was a public API that we would need to remain backwards compatible with. In this case, I will just remove this enum variant, since it should never be getting created with my changes.

src/utils/auth_token/auth_token.rs
szokeasaurusrex: I think I'll just go with an `as_str` function

src/utils/auth_token/org_auth_token.rs
szokeasaurusrex: It is already validated on lines 51-53 – I will update the comment to indicate that we are skipping since the prefix is already validated to avoid future confusion

Swatinem (approved):

Issue #1888, opened 2024-01-05
Sentry-cli 2.24.0 login with auth token causes errors


PotatooSalad:
I used sentry-cli version 2.24.0 in bash command line.

I was deploying new release to sentry, with the same settings as always and after `sentry-cli` login --auth-token [...]` i got this error:
`thread 'main' panicked at 'Mismatch between definition and access of `auth_token`. Could not downcast to TypeId { t: 205527081075292425013923987422355844732 }, need to downcast to TypeId { t: 264673182028513726705440804281905241306 }
': C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\clap-4.1.6\src\parser\error.rs:30`

I tried with other tokens and I get the same message.

Yesterday everything was fine, issue started today after 2.24.0 release. I always use latest version to be up to date, but I guess I need to downgrade to 2.23.2 untill it's fixed.

szokeasaurusrex:
Thank you for reporting this error; it appears related to a recent change that was released in version 2.24.0. We will immediately investigate what is happening here.

PR #1893, opened 2024-01-08
fix(api): Fix panic that occurred when `sentry-cli login` called with `--auth-token`

szokeasaurusrex:
This PR fixes the bug in #1885 that led us to revert that commit. The bug, reported in #1888, caused the CLI to panic when the `sentry-cli login` command was called with the `--auth-token` argument.

We determined that the CLI panicked because the `auth_token` was being downcast to a `String` instead of to an `AuthToken` type when being read from the `matches` within the `login` command. This PR corrects the `login` command, so that the `auth_token` is correctly obtained from the `matches` as an `AuthToken`. We also checked whether the `auth_token` argument was being read as a `String` anywhere else in the code, and found that the `login` command was the only place the argument was being read as a `String`. Therefore, we would expect this PR to fix this class of bug completely.

Fixes GH-1859
Fixes GH-1888


src/commands/login.rs
```diff
@@ -31,7 +32,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         "{}/orgredirect/organizations/:orgslug/settings/auth-tokens/",
         config.get_base_url()?
     );
-    let predefined_token = matches.get_one::<String>("auth_token");
+    let predefined_token = matches.get_one::<AuthToken>("auth_token");
```
szokeasaurusrex, line 35: This is the change that fixes the error – the other changes to this file in this commit are to handle the new type

Swatinem (approved):

Issue #1895, opened 2024-01-16
Sentry CLI Xcode background uploads are not working


krystofwoldrich:
### Environment

What version of the CLI and what OS are you running?

macOS 14.1.2
Xcode 15.2 (15C500b)
Sentry CLI 2.21.4 – [2.25.0](https://github.com/getsentry/sentry-cli/releases/tag/2.25.0)

### Steps to Reproduce

Related:
- https://github.com/getsentry/sentry-react-native/issues/3456

This issue is a follow-up to the RN SDK issue, I reproduced the bug outside of RN projects.

Execute the following command in the run build script build phase.

```bash
sentry-cli debug-files upload "$DWARF_DSYM_FOLDER_PATH"
```

The last version where the background upload is working is 2.21.3.

- https://github.com/getsentry/sentry-cli/compare/2.21.3...2.21.4

### Expected Result

The command should finish uploading debug files in the background.

### Actual Result

The command stopped after going to the background without any error, no files were uploaded.

### Logs

When filing a bug report, please attach debug logs, either by using `--log-level=debug` flag or `SENTRY_LOG_LEVEL=debug` environment variable.

Before going to the background:
```bash
  WARN    2024-01-16 13:12:28.286868 +01:00 Failed to find file referenced by SENTRY_PROPERTIES (sentry.properties)
  DEBUG   2024-01-16 13:12:28.287177 +01:00 sentry-cli version: 2.21.4, platform: "darwin", architecture: "arm64"
  INFO    2024-01-16 13:12:28.287200 +01:00 sentry-cli was invoked with the following command line: "/Users/krystofwoldrich/repos/sentry-cli/sentry-cli-Darwin-arm64.2.21.4" "debug-files" "upload" "--log-level=debug" "/Users/krystofwoldrich/Library/Developer/Xcode/DerivedData/Sentry_CLI_Background_Test-csojtsktdkcocielvfjoqmdlcrbz/Build/Products/Debug-iphonesimulator"
  INFO    2024-01-16 13:12:28.287226 +01:00 Issuing a command for Organization: sentry-sdks Project: sentry-react-native
Continuing in background.
```

From background redirected stdout and stderr:
```bash
[30m[43m  DEBUG  [0m [2m2024-01-16 10:39:50.899044 +01:00[0m request GET https://sentry.io/api/0/organizations/sentry-sdks/chunk-upload/[2m[0m
[...]

PR #1896, opened 2024-01-16
Revert "chore: bump curl-sys dependency to 0.4.68 (#1815)"

krystofwoldrich:
This reverts commit 6b9b32b4481e79282f14dcfbb52f54f2e59b1d7d.

- This fixes https://github.com/getsentry/sentry-cli/issues/1895

- But it brings a bug back on Windows https://github.com/getsentry/sentry-cli/pull/1815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 "You know what they say. Fool me once, strike one, but fool me twice... strike three." — Michael Scott
 
+## Unreleased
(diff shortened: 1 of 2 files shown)


szokeasaurusrex (approved):
Looks good! Please open an issue referencing the original Windows problem after merging this PR.

#1897 should also solve the Windows problem, since it would get rid of `curl` completely

Issue #1913, opened 2024-01-23
Prefer `--url` over empty Org Auth Token `url`


szokeasaurusrex:
If an Org Auth Token's payload defines an **empty but not null** `url`, then we should prefer the url provided as a command line argument via `--url`. The current behavior is to raise an error that the two urls don't match, which makes little sense, given that an empty url in the org Auth Token likely indicates a misconfigured Sentry self-hosted instance, and most users would expect the `--url` to take precedence.

https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/465#issuecomment-1906453702

PR #1914, opened 2024-01-24
Prefer `--url` argument over empty auth token URL

szokeasaurusrex:
With this change, the CLI will treat an org auth token that specifies an empty URL the same way as an org auth token that is not specifying a URL at all.

We implemented this change in behavior by changing the type of the `url` in the `AuthTokenPayload` from an `Option<String>` to a `String`. In cases where the `url` previously would have been set to `None`, the `url` will now be set to an empty string. Uses of the `AuthTokenPayload`'s `url` field have been updated to treat empty strings the way they previously treated `None` values.

Fixes GH-1913

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -63,7 +63,7 @@ impl Config {
         let default_url = get_default_url(&ini);
         let token_url = token_embedded_data
             .as_ref()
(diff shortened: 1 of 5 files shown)


loewenheim (approved):
Looks good to me!

PR #1917, opened 2024-01-27
print the actual source map URL instead of {source_url} on error

jbg:


szokeasaurusrex (approved):
Looks good, thanks for the contribution!

Issue #1915, opened 2024-01-24
Sentry CLI Xcode build phase fails silently


lionel-alves:
### Environment

What version of the CLI and what OS are you running?

macOS 14.1
Xcode 15.2 (15C500b)
Sentry CLI [2.26.0](https://github.com/getsentry/sentry-cli/releases/tag/2.26.0)

### Steps to Reproduce

Related to https://github.com/getsentry/sentry-cli/issues/1895

Execute the following command in the run build script build phase.
```
sentry-cli debug-files upload --include-sources "$DWARF_DSYM_FOLDER_PATH"
```

I was able to reproduce this issue en 2.26.0 and 2.21.3 as well.

### Expected Result

The command should finish uploading debug files in the background.

### Actual Result

The command stopped after going to the background without any error, no files were uploaded.

### Logs

When filing a bug report, please attach debug logs, either by using `--log-level=debug` flag or `SENTRY_LOG_LEVEL=debug` environment variable.

Before going to the background:
```
  DEBUG   2024-01-24 22:15:21.475418 +01:00 sentry-cli version: 2.21.3, platform: "darwin", architecture: "arm64"
  INFO    2024-01-24 22:15:21.475577 +01:00 sentry-cli was invoked with the following command line: "sentry-cli" "debug-files" "upload" "--include-sources" "/Users/.../Library/Developer/Xcode/DerivedData/...-hczxkiymbkgecveoyxfadvexicxk/Build/Products/Debug-iphonesimulator"
  INFO    2024-01-24 22:15:21.475587 +01:00 Issuing a command for Organization: ... Project: ...
Continuing in background.
```

Not sure how to get the logs after that.

szokeasaurusrex:
Thank you for reporting! It would be extremely helpful to get the logs from the background, though. Perhaps @krystofwoldrich would know how you can obtain the logs, since he was able to include them on #1895

krystofwoldrich:
@lionel-alves Thank you for testing it with the older CLI version.

- We are currently not logging the file path with the redirected output https://github.com/getsentry/sentry-cli/pull/1920

But you should be able to find the file in `$TMPDIR`, after running the `Build Phase`, the file name is a UUID v4.

szokeasaurusrex:
@lionel-alves We just released version `2.27.0`, which includes the #1920 change that will print the log output file path when the CLI goes into the background.

Please share the background logs with us, so we can investigate your issue.

lionel-alves:
[...]

PR #1920, opened 2024-01-29
feat(xcode): Print redirected output file path when going to the background

krystofwoldrich:
Enabled users to share background logs with us.

Help debug the following issues:
- https://github.com/getsentry/sentry-cli/issues/1895
- https://github.com/getsentry/sentry-cli/issues/1915

### DIFF
--- a/src/utils/xcode.rs
+++ b/src/utils/xcode.rs
@@ -403,9 +403,12 @@ impl<'a> MayDetach<'a> {
             return Ok(false);
         }
 
-        println!("Continuing in background.");
-        show_notification("Sentry", &format!("{} starting", self.task_name))?;
         let output_file = TempFile::create()?;
+
+        println!("Continuing in background.");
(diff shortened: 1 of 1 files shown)


src/utils/xcode.rs
```diff
@@ -403,9 +403,12 @@ impl<'a> MayDetach<'a> {
             return Ok(false);
         }
 
+        let output_file = TempFile::create()?;
+
         println!("Continuing in background.");
+        println!("Output is redirected to {}", output_file.path().display());
```
szokeasaurusrex, line 409: Perhaps it would make more sense to log this as a debug message, so we only print this line if the user specified the debug log level?

krystofwoldrich: I understand but I think there is so little information before going to the background that knowing the path is essential we should always log it.

szokeasaurusrex (approved):

PR #1934, opened 2024-02-09
ci: Lint without `actions-rs`

szokeasaurusrex:
We should lint without `actions-rs`. `actions-rs` is unmaintained, and we can call the necessary Cargo commands directly. `cargo clippy` occasionally fails when run with `actions-rs` even though it does not fail when run locally. Running the Cargo commands directly appears to fix the problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,28 +19,15 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # pin@v1
-        with:
-          toolchain: stable
-          profile: minimal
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,28 +19,15 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # pin@v1
```
Swatinem, line 22: you should still use `rustup` to ensure you have the latest Rust, as the github runners might have a random version pre-installed

Swatinem (approved):
lgtm, while you are at it, you can update all the rest of the actions as well, including the `checkout` action and my `rust-cache` which is at `v2` for quite a while already

PR #1935, opened 2024-02-09
ci: Updates to lint GH action

szokeasaurusrex:
This PR implements @Swatinem's suggestions from #1934. #1934 was auto-merged before I could implement the suggestions there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,12 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Rust Toolchain
-        run: rustup toolchain install stable --profile minimal --component clippy rustfmt
+        run: rustup toolchain install stable --profile minimal --component clippy --component rustfmt --no-self-update
 
       - uses: swatinem/rust-cache@v2
-        with:
-          key: ${{ github.job }}
 
       - name: Run Rustfmt
         run: cargo fmt --all -- --check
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -17,9 +17,12 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - uses: swatinem/rust-cache@81d053bdb0871dcd3f10763c8cc60d0adc41762b # pin@v1
+      - name: Install Rust Toolchain
+        run: rustup toolchain install stable --profile minimal --component clippy rustfmt
```
szokeasaurusrex, line 23: Does this line look correct? I referenced [this line](https://github.com/getsentry/relay/blob/c5c4b0561bb174739d87c1b9e69ac73197bc2321/.github/workflows/ci.yml#L39) from Relay and the old config changed in #1934

Swatinem: there is also `--no-self-update`, but I believe that is most relevant for Windows.

You can take a look at https://github.com/getsentry/ophio/blob/c8b08063562f67306b6cc22a36322806fe8e588c/.github/workflows/ci.yml#L19-L25, which is the "most recent" version of setting all this up.

Also, `checkout` is already at `v4` as well :-)

.github/workflows/ci.yml
```diff
@@ -17,9 +17,12 @@ jobs:
...
+
+      - uses: swatinem/rust-cache@v2
         with:
           key: ${{ github.job }}
```
Swatinem, line 27: you also don’t need the `with: key: …` anymore with my `v2` action.

Swatinem (approved):

Issue #1936, opened 2024-02-09
Could not parse Info.plist file  - caused by: Serde("invalid type: integer `32924`, expected a string")


kenyipp:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.18.0

**`react-native` version:** 0.70.13

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  // other options
});
```
---
**I have the following issue:**

Since yesterday, I've been encountering the following error message when attempting to build the iOS app:

```
❌  error: Could not parse Info.plist file
    Run script build phase '[CP-User] Config codegen' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'react-native-config' from project 'Pods')
    Run script build phase '[CP-User] Generate app.config for prebuilt Constants.manifest' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'EXConstants' from project 'Pods')
Last lines of the Xcode's build log:
    export __IS_NOT_SIMULATOR_simulator\=NO
    export arch\=undefined_arch
    export variant\=normal
    /bin/sh -c /Users/[REDACTED]/Library/Developer/Xcode/DerivedData/Mylo-fcequrikxyqjreaosttkhtfkpnoi/Build/Intermediates.noindex/ArchiveIntermediates/Mylo/IntermediateBuildFilesPath/Mylo.build/Release-iphoneos/Mylo.build/Script-A31ED4BC5358436EB4C1FDEE.sh
  INFO    2024-02-09 00:33:23.878117 +00:00 Loaded file referenced by SENTRY_PROPERTIES (sentry.properties)
error: Could not parse Info.plist file
[...]

PR #1940, opened 2024-02-12
fix(xcode): Only parse Plist when required during RN source maps upload

krystofwoldrich:
- this fix enables a workaround for https://github.com/getsentry/sentry-cli/issues/1936

If users use debug IDs and auto-release reading is disabled, we don't have to read and part Info.plist.

The same applies when users use ENV to set release and dist

### DIFF
--- a/src/commands/react_native/xcode.rs
+++ b/src/commands/react_native/xcode.rs
@@ -201,12 +201,6 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         return Ok(());
     }
 
-    info!("Parsing Info.plist");
-    let plist = match InfoPlist::discover_from_env()? {
-        Some(plist) => plist,
-        None => bail!("Could not find info.plist"),
-    };
-    info!("Parse result from Info.plist: {:?}", &plist);
     let report_file = TempFile::create()?;
     let node = find_node();
     info!("Using node interpreter '{}'", &node);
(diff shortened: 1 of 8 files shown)


src/commands/react_native/xcode.rs
```diff
@@ -371,21 +365,43 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+            }
+
+            if let Ok(_dist) = dist_from_env {
+              info!("Using dist from `SENTRY_DIST` env");
+              dist = Some(_dist);
+            }
+            if let Ok(_release) = release_from_env {
+              info!("Using release from `SENTRY_RELEASE` env");
+              release_name = Some(_release);
+            }
```
szokeasaurusrex, line 396: I find this logic somewhat difficult to follow – I would prefer, if possible, to set the `dist` and `release_name` in one place, ideally keeping them immutable. Perhaps this could be achieved with a `match` statement?

szokeasaurusrex: Also, please check the indentation (the linter/formatter might be checking it already, but it appears the code is inconsistently combining both 2- and 4-space indentations)

krystofwoldrich: Thank you, yes I agree, I've used match, and now it's much simpler!

src/commands/react_native/xcode.rs
```diff
@@ -371,21 +365,44 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                            let dist_string = plist.build().to_string();
+                            let release_string = format!("{}@{}+{}", plist.bundle_id(), plist.version(), dist_string);
+                            info!("Parse result from Info.plist: {:?}", &plist);
+                            (Some(dist_string), Some(release_string))
+                        },
+                        _ => {
+                            bail!("Info.plist was not found or an parsing error occurred");
+                        }
+                    }
+                }
```
szokeasaurusrex, line 396:
```suggestion
                (Err(_), Err(_)) => {
                    // Neither environment variable is present, attempt to parse Info.plist
                    info!("Parsing Info.plist");
                    match InfoPlist::discover_from_env() {
                        Ok(Some(plist)) => {
                            // Successfully discovered and parsed Info.plist
                            let dist_string = plist.build().to_string();
                            let release_string = format!("{}@{}+{}", plist.bundle_id(), plist.version(), dist_string);
                            info!("Parse result from Info.plist: {:?}", &plist);
                            (Some(dist_string), Some(release_string))
                        },
                        _ => {
                            bail!("Info.plist was not found or an parsing error occurred");
                        }
                    }
                }
                _ => (dist_from_env.ok(), release_from_env.ok()),
```

We can combine the first three cases into a single case by using the [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) function, which converts `Result<T, E>` into `Option<T>`. To do this, however, the order needs to be rearranged.

szokeasaurusrex (changes requested):
Thank you for switching to a `match` statement; the logic is much clearer to me know, and I was able to identify a way we can simplify the `match` statement even further. Once you address the suggestion, I am ready to approve.

szokeasaurusrex (approved):
Looks good, thanks!

PR #1943, opened 2024-02-13
fix: support windows on arm via x64 binary

MarshallOfSound:
Ensures #1858 works even after a refactor


szokeasaurusrex (approved):
Thanks for the contribution!

PR #1962, opened 2024-03-01
fix: Merge "Scripts" and "Minified Scripts" under sourcemaps upload report

szokeasaurusrex:
The distinction between "Scripts" and "Minified Scripts" is arbitrary. #1958 already eliminates the distinction for `sourcemaps inject`, and since the same code is also used to distinguish between files to output as "Scripts" and "Minified Scripts," merging #1958 already will cause the two sections to merge in the `sourcemaps upload` output.

However, with #1958, the two sections are merged as "Minified Scripts." If this PR is applied along with #1958 (which is my intention), it will cause the two sections to be merged as "Scripts." However, this PR is also completely independent of #1958 – applied alone, it will have the same effect of merging the "Minified Scripts" and "Scripts" sections of the sourcemaps upload report output under "Scripts."

### DIFF
--- a/tests/integration/_cases/react_native/xcode-upload-source-maps-release_and_dist_from_env.trycmd
+++ b/tests/integration/_cases/react_native/xcode-upload-source-maps-release_and_dist_from_env.trycmd
@@ -19,8 +19,7 @@ Processing react-native sourcemaps for Sentry upload.
 
 Source Map Upload Report
   Scripts
-    ~/react-native-xcode-bundle.js.map
-  Minified Scripts
+    ~/react-native-xcode-bundle.js.map (no sourcemap ref)
     ~/react-native-xcode-bundle.js (sourcemap at react-native-xcode-bundle.map)
 
 ```
--- a/tests/integration/_cases/sourcemaps/sourcemaps-upload-some-debugids.trycmd
+++ b/tests/integration/_cases/sourcemaps/sourcemaps-upload-some-debugids.trycmd
@@ -25,8 +25,7 @@ $ sentry-cli sourcemaps upload tests/integration/_fixtures/upload_some_debugids
 
 Source Map Upload Report
   Scripts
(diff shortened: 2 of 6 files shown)


tests/integration/_cases/sourcemaps/sourcemaps-upload-some-debugids.trycmd
```diff
@@ -25,8 +25,7 @@ $ sentry-cli sourcemaps upload tests/integration/_fixtures/upload_some_debugids
 
 Source Map Upload Report
   Scripts
-    ~/server/chunks/flight-server-css-manifest.js
-  Minified Scripts
+    ~/server/chunks/flight-server-css-manifest.js (no sourcemap ref)
```
lforst, line 28: One day we can maybe reword `no sourcemap ref` to something more clear. Eg `no source map reference` or `no source map found.`

lforst (approved):

tests/integration/_cases/react_native/xcode-upload-source-maps-release_and_dist_from_env.trycmd
```diff
@@ -19,8 +19,7 @@ Processing react-native sourcemaps for Sentry upload.
 
 Source Map Upload Report
   Scripts
-    ~/react-native-xcode-bundle.js.map
-  Minified Scripts
+    ~/react-native-xcode-bundle.js.map (no sourcemap ref)
```
szokeasaurusrex, line 22: It is probably a bug that this line is under "Scripts." I already have a separate issue (#1963) to investigate

tests/integration/_cases/sourcemaps/sourcemaps-upload-some-debugids.trycmd
szokeasaurusrex: Good idea, I already opened an issue: #1964

loewenheim (approved):

PR #1969, opened 2024-03-05
feat(crons): Fully drop auth-token based check in support

evanpurkhiser:
We brought this back in GH-1634, since unfortunately at the time we had many customers still using the legacy check-in endpoints.

We've just disabled the legacy endpoints as of [0] and will be removing them completely with [1]. This means the token-auth based version of the monitors run command will no longer work.

Unfortunatly we did claim that we would remove these *only in the next major version*

> Token auth is deprecated for cron monitor checkins and will be removed in the next major version.
> Please use DSN auth.

This was unfortuantely probably a lie since removing these endpoints is something we commited to at the beginning of Q1, and it's unlikely we're going to release a new major just for this

[0]: https://github.com/getsentry/sentry-options-automator/pull/851
[1]: https://github.com/getsentry/sentry/pull/66172

### DIFF
--- a/tests/integration/_cases/monitors/monitors-run-help.trycmd
+++ b/tests/integration/_cases/monitors/monitors-run-help.trycmd
@@ -18,8 +18,6 @@ Options:
   -s, --schedule <schedule>
           Configure the cron monitor with the given schedule (crontab format). Enclose the schedule
           in quotes to ensure your command line environment parses the argument correctly.
-      --auth-token <AUTH_TOKEN>
-          Use the given Sentry auth token.
       --check-in-margin <checkin_margin>
           The allowed margin of minutes after the expected check-in time that the monitor will not
           be considered missed for. Requires --schedule.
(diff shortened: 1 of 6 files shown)


wedamija (approved):

tests/integration/_cases/monitors/monitors-run-help.trycmd
```diff
@@ -15,28 +15,28 @@ Options:
       --header <KEY:VALUE>
           Custom headers that should be attached to all requests
           in key:value format.
-      --auth-token <AUTH_TOKEN>
-          Use the given Sentry auth token.
   -s, --schedule <schedule>
           Configure the cron monitor with the given schedule (crontab format). Enclose the schedule
           in quotes to ensure your command line environment parses the argument correctly.
+      --auth-token <AUTH_TOKEN>
+          Use the given Sentry auth token.
```
szokeasaurusrex, line 22: We should not be including the `--auth-token` in the output for `sentry-cli monitors run --help` if auth token authentication is no longer supported

szokeasaurusrex: I am going to add a commit to tweak this

PR #1983, opened 2024-03-14
fix(hc): Adds support for DSYM config driven uploads

GabeVillalobos:
Adds support for the new `/projects/<organization_slug>/files/dsyms/config/` endpoint, which supplies a config object containing a region-specific upload URL.

This change will alleviate issues we've seen with CLI uploads in the past that go through our integration proxy due to referencing the base `https://sentry.io` url.

Because the new endpoint will only be available in SaaS for a bit, and may take some time for self-hosted users to roll it out, I've made the upload process ignore any 404s for config requests while falling back to the old base URL behavior.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1280,6 +1280,22 @@ impl<'a> AuthenticatedApi<'a> {
         Ok(state.missing)
     }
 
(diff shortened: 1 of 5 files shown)


GabeVillalobos:
Closing this in favor of #67496 which takes a different approach to this.

PR #1984, opened 2024-03-19
fix(hc): Routes DSym upload requests directly to the region based if possible

GabeVillalobos:
Updates the `dif_upload` logic to pull the region URL data from the organization auth token specified by the user. If no region URL is found, the method falls back to the previous implementation using a relative path.

This is an alternate solution to the one specified in PR #1983.


GabeVillalobos:
Closing in favor of #1997

PR #1988, opened 2024-03-25
ref: Fixed new Clippy errors

szokeasaurusrex:
A new Rust version introduced some new Clippy lints, which are now blocking PRs from merging since our code was violating the new lints. This PR fixes those lints.

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -72,6 +72,7 @@ impl TempFile {
             .read(true)
             .write(true)
             .create(true)
+            .truncate(false)
             .open(&self.path)?;
 
         f.rewind().ok();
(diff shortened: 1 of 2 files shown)


src/utils/fs.rs
```diff
@@ -72,6 +72,7 @@ impl TempFile {
             .read(true)
             .write(true)
             .create(true)
+            .truncate(false)
```
szokeasaurusrex, line 75: This line [keeps the old behavior](https://rust-lang.github.io/rust-clippy/master/index.html#/suspicious_open_options) and satisfies Clippy. I would have thought that `truncate(true)` would make more sense, but it causes our tests to fail.

loewenheim (approved):

src/utils/fs.rs
szokeasaurusrex: We will investigate further in #1989

PR #1994, opened 2024-03-25
build: `make lint` only with default features

szokeasaurusrex:
Companion PR for #1993, please see that PR for context!


Swatinem (approved):
this is a bikesheddy topic: I’m not really a fan of makefiles :-)

loewenheim (approved):

Issue #1990, opened 2024-03-25
Route API requests to the organization's region


szokeasaurusrex:
API requests, specific to an organization and requiring authentication, should be routed directly to the organization's region. How we do this will depend on the authentication mechanism:
  - With org auth tokens, we extract the region URL encoded in the token
  - With other auth methods, we will query a new endpoint to obtain the region URL. Should this API request fail, we emit a warning, and upload to the US region for backwards compatibility.
  - These changes shall only affect sentry.io users. Self-hosted users should continue to have the same experience. 

This issue only addresses the `files/dsyms` and `releases/{}/files` endpoints, which are critical to migrate to the region URLs, since they block GA of hybrid cloud. We will address the other endpoints in a separate issue, since they are lower-priority.

szokeasaurusrex:
Implemented by #1997, #1999, and #2002

PR #1997, opened 2024-03-26
feat(api): `files/dsyms/` hits org auth token's region URL

szokeasaurusrex:
When using org auth token authentication, we now direct any API requests to the `projects/{}/{}/files/dsyms/` endpoint directly to the region URL encoded in the token. We only do this when the URL is configured to `https://sentry.io`. For other authentication methods, we default to the old behavior, and log an info message to inform the user that the default URL was used for the upload.

This PR implements part of what is required by #1990. The next step will be to use the region URL when a user auth token is used.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -194,9 +194,9 @@ pub struct AuthenticatedApi<'a> {
     api: &'a Api,
 }
 
-pub struct RegionSpecificApi<'a, 'b> {
+pub struct RegionSpecificApi<'a> {
     api: &'a AuthenticatedApi<'a>,
-    org: &'b str,
+    org: &'a str,
     region_url: Option<&'a str>,
 }
 
@@ -1667,7 +1667,7 @@ impl<'a> AuthenticatedApi<'a> {
         }
     }
 
-    pub fn region_specific<'b>(&'a self, org: &'b str) -> RegionSpecificApi<'a, 'b> {
+    pub fn region_specific(&'a self, org: &'a str) -> RegionSpecificApi<'a> {
         let base_url = self.api.config.get_base_url();
         if base_url.is_err()
             || base_url.expect("base_url should not be error") != DEFAULT_URL.trim_end_matches('/')
@@ -1712,7 +1712,7 @@ impl<'a> AuthenticatedApi<'a> {
     }
 }
 
-impl<'a, 'b> RegionSpecificApi<'a, 'b> {
+impl<'a> RegionSpecificApi<'a> {
     fn request(&self, method: Method, url: &str) -> ApiResult<ApiRequest> {
         self.api.api.request(method, url, self.region_url)
     }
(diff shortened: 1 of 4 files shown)


src/api.rs
```diff
@@ -1685,6 +1666,72 @@ impl<'a> AuthenticatedApi<'a> {
             resp.convert()
         }
     }
+
+    pub fn region_specific<'b>(&'a self, org: &'b str) -> RegionSpecificApi<'a, 'b> {
+        let base_url = self.api.config.get_base_url();
+        if base_url.is_err()
+            || base_url.expect("base_url should not be error") != DEFAULT_URL.trim_end_matches('/')
```
GabeVillalobos, line 1673: Is this any chance this error is surfaced to the user? If so, should we rephrase the expect message to be something more along the lines of "Invalid Base URL provided"?

src/api.rs
```diff
@@ -1685,6 +1666,72 @@ impl<'a> AuthenticatedApi<'a> {
...
+                    api: self,
+                    org,
+                    region_url: None,
+                }
+            }
+        }
+    }
+}
+
+impl<'a, 'b> RegionSpecificApi<'a, 'b> {
```
GabeVillalobos, line 1715: Interesting, we can't use the same lifetime for both of the return types in the Result tuple? (Sorry, my Rust isn't the strongest 😅 )

src/api.rs
```diff
@@ -1685,6 +1666,72 @@ impl<'a> AuthenticatedApi<'a> {
...
+    fn request(&self, method: Method, url: &str) -> ApiResult<ApiRequest> {
```
GabeVillalobos, line 1716: I wonder if we can eventually extract out these methods as a general API trait, providing a regionless implementation as a fallback? It might be a ton of extra plumbing though for the fallback cases 😬

GabeVillalobos (approved):
Changes look good!

src/api.rs
szokeasaurusrex: This error should never occur because the `base_url.is_err()` check before will short circuit. `expect` is equivalent to `unwrap` just allows an error message, but I prefer to use `expect` mostly just to document why it is okay to unwrap the value.

src/api.rs
szokeasaurusrex: I did not try, but I assumed the lifetimes could be different because the org string could live shorter than the AuthenticatedApi

markstory (commented):
Looks good to me, but I have low context on this repo.

src/api.rs
```diff
@@ -194,6 +194,12 @@ pub struct AuthenticatedApi<'a> {
     api: &'a Api,
 }
 
+pub struct RegionSpecificApi<'a, 'b> {
```
Swatinem, line 197: its a bit weird that this has two lifetimes. I believe you can just get along with a single one.

Swatinem (approved):

src/api.rs
szokeasaurusrex: Okay, I will try it out!

PR #1999, opened 2024-03-26
feat(api): Region specific API with user token

szokeasaurusrex:
Extends #1997 to also work with user auth tokens.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -194,9 +194,9 @@ pub struct AuthenticatedApi<'a> {
     api: &'a Api,
 }
 
-pub struct RegionSpecificApi<'a, 'b> {
+pub struct RegionSpecificApi<'a> {
     api: &'a AuthenticatedApi<'a>,
-    org: &'b str,
+    org: &'a str,
     region_url: Option<Box<str>>,
 }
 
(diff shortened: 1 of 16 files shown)


src/api.rs
```diff
@@ -197,7 +197,7 @@ pub struct AuthenticatedApi<'a> {
 pub struct RegionSpecificApi<'a, 'b> {
     api: &'a AuthenticatedApi<'a>,
     org: &'b str,
```
GabeVillalobos, line 199: I'm starting to wonder if it'd be better to just clone these strings to avoid potential lifetime issues in the future? I suppose this only matters if there's a situation where the API outlives the org or region_url references which seems unlikely.

GabeVillalobos (approved):

src/api.rs
szokeasaurusrex: Right now, we simply create the region specific API on the fly when we need it to make a request. We don't store it around, so it will always be shorter-lived than both the `api` and the `org` references

Swatinem (approved):

src/api.rs
```diff
@@ -1680,41 +1686,46 @@ impl<'a> AuthenticatedApi<'a> {
...
-                RegionSpecificApi {
-                    api: self,
-                    org,
-                    region_url: Some(region_url),
+            Auth::Token(token) => match token.payload() {
+                Some(payload) => Some(payload.region_url.clone().into()),
+                None => {
+                    let region_url = self.get_region_url(org);
+                    if let Err(err) = &region_url {
+                        log::warn!("Failed to get region URL due to following error: {err}");
```
szokeasaurusrex, line 1700: Perhaps, we should consider erroring here? I made this a warning for backwards compatibility, but since we only reach this line if we have verified the user is using `sentry.io`, we don't need to remain backwards compatible with self-hosted instances here, so an error might be okay. I could imagine that maybe some customers don't want us sending their data to the US region if we fail to retrieve the region URL.

szokeasaurusrex: We can probably make this change in a separate PR, however

PR #2006, opened 2024-03-28
Added default count for large commits

ankitghosh:
We faced an issue where for some reason our commit history was huge around 45K, sentry-cli set-commits command was in a hung state for this use case. 
When I dig deep I found initial-depth flag (default_count) was not being used to limit the commits `get_commits_from_git` function. Default limit set is 20. 
So to fix that i have added a filter to take only `default_count` number of commits.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -471,6 +471,7 @@ pub fn get_commits_from_git<'a>(
                 .filter_map(move |id: Result<git2::Oid, git2::Error>| {
                     repo.find_commit(id.ok()?).ok()
                 })
+                .take(default_count)
                 .collect();
 
             // If there is a previous commit but cannot find it in git history


ankitghosh:
@loewenheim please can you take a look at this PR.

szokeasaurusrex:
Hi @ankitghosh, I have taken over from @loewenheim as the primary Sentry CLI maintainer.

While I appreciate the PR, it needs significant changes before we can consider merging it. `default_count` specifies the number of commits to include when the previous release's SHA cannot be found, and the `--ignore-missing` command line argument is provided. It is meant to only be used for this specific purpose. The default behavior is, and should continue to be, that we search for the previous commit with no limit to how far back we go. Implementing a new default limit would be a breaking change, since users might have releases with more than the default number of commits, and implementing a default limit would prevent them from being able to make such releases.

We can consider a different solution where we have a separate limit specifically for how far back we search for the commit. However, this limit would need to be disabled by default to avoid introducing a breaking change.

Since we need to rethink how we make this change entirely, I will close this PR for now. Please [open a new issue](https://github.com/getsentry/sentry-cli/issues/new?assignees=&labels=&projects=&template=BUG_REPORT.md)[^*] referencing this PR, and we can continue our discussion there!

[^*]: For future reference, we typically triage issues more quickly than PRs, so I would always recommend opening an issue when you have a problem, so we can assist ASAP! You can always also open a PR and link the PR to the issue.

ankitghosh:
Created a new issue https://github.com/getsentry/sentry-cli/issues/2026

Issue #2008, opened 2024-04-02
Refactor `src/api.rs`


szokeasaurusrex:
We should split the code into multiple sub-modules to improve code organization, and simplify the code along the way if possible. We will split the work into two phases.

# Phase 1 – current phase

In this phase, we focus on splitting off code that we can relatively easily move out of the main src/api/mod.rs file into separate submodules, with little to no modification. We can consider simplifying any overly complex split out code if it is easy to do so, but the main focus should just be to reduce the line count in the main src/api/mod.rs file by moving things that don't need to be there out of the file. 

```[tasklist]
### Subtasks
- [x] Split helper types (e.g. error types and pagination logic) into separate submodules
- [ ] https://github.com/getsentry/sentry-cli/issues/2077
- [ ] https://github.com/getsentry/sentry-cli/issues/2079
- [ ] https://github.com/getsentry/sentry-cli/issues/2078
```

### Why should we implement the proposed changes in this phase?

The main benefit will be to make development on the `src/api` portion of the Sentry CLI codebase easier. `src/api/mod.rs` is over 2500 lines long, and the file is doing many different things. Thus, it is difficult for developers to find where they need to make changes when fixing a bug/implementing a new feature.

In the sub-issues above, I have identified several areas of the file's code that have very clear distinct responsibilities, and can therefore relatively easily be split into separate files. Separating items into separate files makes it easier for devs to find the code they need to change without being overwhelmed, and hopefully sets a precedent that future new functionality is split into its own file rather than making a very long file even longer. And it is easier to write tests when things are separated out.

### What are the risks of implementing the proposed changes?

The risk is that we could break something if we make a mistake when moving code around. 

We can mitigate this risk by manually testing our changes before publishing them. We can also mitigate the risk by limiting changes during this refactor to simply moving code around – if we also attempt to simplify code, we greatly increase the risk of making a mistake.

# Phase 2 – under consideration

After completing phase 1, we will have greatly improved the `api` module by moving logic that can live in a submodule out of the main module. This should reduce the main module's original size by about half – this is a great improvement, but with ~1500 lines, the file is still a bit long.

Phase 2 is, however, less simple that Phase 1, because we can no longer gain much by splitting the remaining `Api`, `AuthenticatedApi`, and `RegionSpecificApi` structs into separate files – `AuthenticatedApi` itself is ~1000 lines. We likely need to change the architecture of how this part of our codebase works, so we can move things into separate submodules.

[...]

PR #2014, opened 2024-04-03
ref(api): Simplify pagination code

szokeasaurusrex:
Simplifies pagination logic by removing result that never results in error, and introducing new `next_pagination_cursor` convenience method. Tests added in #2015.

ref #2008

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -3,6 +3,8 @@
 //! to the GitHub API to figure out if there are new releases of the
 //! sentry-cli tool.
 
+mod pagination;
+
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
@@ -11,7 +13,6 @@ use std::fs::{create_dir_all, File};
 use std::io::{self, Read, Write};
 use std::path::Path;
 use std::rc::Rc;
-use std::str::FromStr;
 use std::sync::Arc;
 use std::{env, fmt};
 
@@ -41,11 +42,13 @@ use uuid::Uuid;
 use crate::config::{Auth, Config};
 use crate::constants::{ARCH, DEFAULT_URL, EXT, PLATFORM, RELEASE_REGISTRY_LATEST_URL, VERSION};
 use crate::utils::file_upload::UploadContext;
-use crate::utils::http::{self, is_absolute_url, parse_link_header};
+use crate::utils::http::{self, is_absolute_url};
 use crate::utils::progress::ProgressBar;
 use crate::utils::retry::{get_default_backoff, DurationAsMilliseconds};
 use crate::utils::sourcemaps::get_sourcemap_reference_from_headers;
 use crate::utils::ui::{capitalize_string, make_byte_progress_bar};
+
+use self::pagination::Pagination;
 
 // Based on https://docs.rs/percent-encoding/1.0.1/src/percent_encoding/lib.rs.html#104
 // WHATWG Spec: https://url.spec.whatwg.org/#percent-encoded-bytes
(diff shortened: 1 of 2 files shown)


src/api/pagination.rs
```diff
@@ -1,42 +1,35 @@
...
-                cursor: (*item.get("cursor").unwrap_or(&"")).to_string(),
-            });
-        }
-
-        Ok(rv)
+impl<'p> From<&'p str> for Pagination<'p> {
+    fn from(value: &'p str) -> Self {
+        http::parse_link_header(value)
+            .iter()
+            .rev() // Reversing is necessary for backwards compatibility with a previous implementation
```
szokeasaurusrex, line 26: I reversed the iterator since, in the case where there are multiple link headers with `"rel"`, we will end up using the last one with the old implementation (since we overwrite all previous values we may have found). Do you think this reversal is necessary, or can we be sure that there will only be one `"rel"` header? If we aren't sure, let's just keep the `rev`; the comment should prevent people from being too confused.

src/api/pagination.rs
```diff
@@ -1,42 +1,15 @@
...
-                cursor: (*item.get("cursor").unwrap_or(&"")).to_string(),
-            });
-        }
-
-        Ok(rv)
-    }
+pub(super) fn next_cursor(value: &str) -> Option<&str> {
+    http::parse_link_header(value)
+        .iter()
+        .rev() // Reversing is necessary for backwards compatibility with a previous implementation
```
szokeasaurusrex, line 6: I reversed the iterator since, in the case where there are multiple link headers with "rel", we will end up using the last one with the old implementation (since we overwrite all previous values we may have found). Do you think this reversal is necessary, or can we be sure that there will only be one "rel" header? If we aren't sure, let's just keep the rev; the comment should prevent people from being too confused.

src/api/mod.rs
```diff
@@ -675,9 +674,9 @@ impl<'a> AuthenticatedApi<'a> {
                 }
             }
 
-            let pagination = resp.pagination();
+            let next = resp.next_pagination_cursor();
             rv.extend(resp.convert::<Vec<Artifact>>()?);
-            if let Some(next) = pagination.into_next_cursor() {
+            if let Some(next) = next {
```
Swatinem, line 679: Is there a reason the `extend` is sandwitched between getting the `next` and matching it?
It might be simpler to read having the call and the the match on the same line.

src/api/mod.rs
```diff
@@ -1472,13 +1471,13 @@ impl<'a> AuthenticatedApi<'a> {
             if resp.status() == 404 {
                 break;
             } else {
-                let pagination = resp.pagination();
-                rv.extend(resp.convert::<Vec<Repo>>()?);
-                if let Some(next) = pagination.into_next_cursor() {
+                let next = resp.next_pagination_cursor();
+                if let Some(next) = next {
```
Swatinem, line 1475:
```suggestion
                if let Some(next) = resp.next_pagination_cursor() {
```

src/api/mod.rs
```diff
@@ -2111,11 +2110,10 @@ impl ApiResponse {
...
     }
 
-    /// Returns the pagination info
-    pub fn pagination(&self) -> Pagination {
+    fn next_pagination_cursor(&self) -> Option<String> {
         self.get_header("link")
-            .and_then(|x| x.parse().ok())
-            .unwrap_or_default()
+            .and_then(pagination::next_cursor)
+            .map(String::from)
```
Swatinem, line 2116: I believe converting to a `String` is because of borrowing? Possibly caused by the sandwitching I mentioned above. So one more reason to figure out if we really need that?

Swatinem (approved):

src/api/mod.rs
szokeasaurusrex: Yes, resp.convert consumes ownership of resp

src/api/mod.rs
szokeasaurusrex: Yes, exactly. But the `resp.convert` is not the only problem; also, we are storing the next cursor in a variable that would outlive `resp` even if `resp.convert` would not take ownership of `resp`. I already spent too much time trying to figure out how to avoid a string creating here; maybe later on in the refactoring we can try to get rid of it.

Issue #1968, opened 2024-03-05
More tests in Sentry CLI


szokeasaurusrex:
We should add more tests to the Sentry CLI, to ensure we are thoroughly testing common usage scenarios and edge cases. We especially need to test `sourcemaps inject` more thoroughly, but all areas of the CLI could use more testing.

We might also consider testing the CLI in different ways. Currently, most of the tests are integration tests that check the exact output of a command. Perhaps, we could add more unit tests to test narrower portions of the code, and to make it easier to have parameterized test cases where we run similar tests with slightly different parameters.

It will likely make sense to split this issue into several sub-issues.

szokeasaurusrex:
This issue is too vague to be helpful

PR #2015, opened 2024-04-03
test(api): Add tests for pagination submodule

szokeasaurusrex:
Depends on #2014

ref #1968

### DIFF
--- a/src/api/pagination.rs
+++ b/src/api/pagination.rs
@@ -1,0 +1,42 @@
+use std::str::FromStr;
+
+use crate::utils::http;
+
+#[derive(Debug, Clone)]
+pub struct Link {
+    results: bool,
+    cursor: String,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct Pagination {
+    next: Option<Link>,
+}
+
+impl Pagination {
+    pub fn into_next_cursor(self) -> Option<String> {
+        self.next
+            .and_then(|x| if x.results { Some(x.cursor) } else { None })
+    }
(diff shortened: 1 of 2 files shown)


src/api/pagination.rs
```diff
@@ -13,3 +13,34 @@ pub(super) fn next_cursor(value: &str) -> Option<&str> {
...
+            <https://sentry.io/api/0/organizations/sentry/releases/?&cursor=100:1:0>; \
+            rel=\"next\"; results=\"true\"; cursor=\"100:1:0\"",
+        );
+        assert_eq!(result, Some("100:1:0"));
+    }
+
+    #[test]
+    fn test_pagination_without_next() {
+        let result = next_cursor(
+            "<https://sentry.io/api/0/organizations/sentry/releases/?&cursor=100:-1:1>; \
```
Swatinem, line 41: using a raw string like `r#""#` would be nice because then you won’t have to escape the quotes. Though not sure how that plays with the line continuations.

Swatinem (approved):

src/api/pagination.rs
szokeasaurusrex: The line continuations would appear as actual newline characters in the string, since the raw string does not allow us to escape anything (including newlines). If we do a raw string, everything needs to be on a single line to ensure the string has the same value. 

I think it's more readable to break the string across multiple lines than to get rid of the escaped quotes but have everything on a single line. So, if it's cool with you, let's keep it as is.

PR #2022, opened 2024-04-09
test(api): Add tests for Sentry error

szokeasaurusrex:
ref #1968

### DIFF
--- a/src/api/errors/sentry_error.rs
+++ b/src/api/errors/sentry_error.rs
@@ -34,3 +34,72 @@ impl fmt::Display for SentryError {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    fn error_from_status_and_detail_helper(status: u32, detail: Option<&str>) -> SentryError {
+        SentryError {
+            status,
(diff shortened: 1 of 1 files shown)


src/api/errors/sentry_error.rs
```diff
@@ -34,3 +34,72 @@ impl fmt::Display for SentryError {
...
+    #[test]
+    fn test_display_no_detail_with_extra() {
+        let error = SentryError {
+            status: 400,
+            detail: None,
+            extra: Some(serde_json::json!("extra info")),
+        };
+        assert_eq!(
+            format!("{}", error),
+            "sentry reported an error: bad request (http status: 400)\n  String(\"extra info\")"
```
Swatinem, line 102:
```suggestion
            format!("{error}"),
            "sentry reported an error: bad request (http status: 400)\n  String(\"extra info\")"
```

Its weird that you are using `Debug` formatting for the `serde_json::Value`. Why not just format that as JSON instead?

Swatinem (approved):
tbh, these tests add very little value

szokeasaurusrex:
Okay, let's just revert. I wasn't super confident that we need these tests, either

PR #2023, opened 2024-04-09
ref(api): ApiError submodule

szokeasaurusrex:
ref #2008

### DIFF
--- a/src/api/errors/api_error.rs
+++ b/src/api/errors/api_error.rs
@@ -1,0 +1,89 @@
+use std::fmt;
+
+#[derive(Debug, thiserror::Error)]
+pub struct ApiError {
+    inner: ApiErrorKind,
+    #[source]
+    source: Option<anyhow::Error>,
+}
+
+/// Represents API errors.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, thiserror::Error)]
+pub(in crate::api) enum ApiErrorKind {
+    #[error("could not serialize value as JSON")]
+    CannotSerializeAsJson,
+    #[error("could not parse JSON response")]
+    BadJson,
+    #[error("not a JSON response")]
+    NotJson,
+    #[error("request failed because API URL was incorrectly formatted")]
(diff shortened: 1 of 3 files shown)


src/api/mod.rs
```diff
@@ -1923,7 +1923,7 @@ impl ApiResponse {
     }
 
     /// Like convert but produces resource not found errors.
-    pub fn convert_rnf<T: DeserializeOwned>(self, res_err: ApiErrorKind) -> ApiResult<T> {
+    fn convert_rnf<T: DeserializeOwned>(self, res_err: ApiErrorKind) -> ApiResult<T> {
```
szokeasaurusrex, line 1926: Privatizing this function allows us to restrict `ApiErrorKind` to `pub(in crate::api)`

src/api/errors/api_error.rs
```diff
@@ -0,0 +1,89 @@
...
+    where
+        E: Into<anyhow::Error>,
+    {
+        ApiError {
+            inner: kind,
+            source: Some(source.into()),
+        }
+    }
+
+    pub(in crate::api) fn kind(&self) -> ApiErrorKind {
```
Swatinem, line 60: interesting, first time I see this syntax. Is that the same as `pub(super)`? Hm, actually that would be equivalent to `crate::api::errors` I believe.

Swatinem (approved):

src/api/errors/api_error.rs
szokeasaurusrex: No, actually `pub(super)` is syntactical sugar for `pub(in super)`. `pub(in xyz)` restricts visibility to module `xyz` and `xyz`'s submodules ([see docs](https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself)).

In this case, we need to use `pub(in crate::api)` rather than `pub(super)`, since this function gets called in the `src/api/mod.rs`.

Issue #1924, opened 2024-02-01
Update file permissions error message and docs


szokeasaurusrex:
We want to prevent confusion in cases like this: https://github.com/getsentry/sentry-cli/issues/1902#issuecomment-1906088229

szokeasaurusrex:
## Solution brainstorm

1. Find the line of code where we attempt to load `.sentryclirc`
2. If we have an error opening the file, check whether we are running in Xcode (I believe we do this somewhere else in the code, so we should use the same logic here)
3. If we are in Xcode, add an additional line to the error message that informs the user that they need to specify `.sentryclirc` as an input file to the Xcode run script to allow it through the sandbox.
4. Update the documentation (repo `getsentry/docs`) to explain that `.sentryclirc` needs to be listed as an input file. A screenshot, like the one linked in https://github.com/getsentry/sentry-cli/issues/1902#issuecomment-1906088229, might be helpful.

PR #2028, opened 2024-04-11
fix(xcode): Improve Xcode error msg when config load fails

elramen:
When a user sets up a project in xcode, e.g., by using the wizard, the .sentryclirc config is created in the project folder. However, when this file isn't included in the Input Files of the Xcode script for uploading debug files, an error is thrown  in Xcode which isn't very clear. This commit improves that error message by specifying what the user needs to do.

Fixes GH-1924

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -20,7 +20,6 @@ use crate::constants::{CONFIG_RC_FILE_NAME, DEFAULT_RETRIES, DEFAULT_URL};
 use crate::utils::auth_token::AuthToken;
 use crate::utils::auth_token::AuthTokenPayload;
 use crate::utils::http::is_absolute_url;
-use crate::utils::xcode;
 
 /// Represents the auth information
 #[derive(Debug, Clone)]
@@ -579,7 +578,7 @@ fn load_global_config_file() -> Result<(PathBuf, Ini)> {
 fn get_failed_cli_config_load_msg(file_desc: &str) -> String {
     let msg = format!("Failed to load {file_desc}.");
     #[cfg(target_os = "macos")]
-    if xcode::launched_from_xcode() {
+    if crate::utils::xcode::launched_from_xcode() {
         return msg + " Hint: Please ensure that ${SRCROOT}/.sentryclirc is added to the Input Files of this Xcode Build Phases script.";
     }
     msg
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -575,6 +576,15 @@ fn load_global_config_file() -> Result<(PathBuf, Ini)> {
     }
 }
 
+fn get_failed_cli_config_load_msg(file_desc: &str) -> String {
```
szokeasaurusrex, line 578:
```suggestion
fn failed_config_load_message(file_desc: &str) -> String {
```

Small nitpick; I would remove the words `get` and `cli` to make the name more concise, but expand `msg` to `message`, so we use full words ideally rather than arbitrary shortenings.

Perhaps, you could also consider adding a documentation comment to state that this is the error message we print when we fail to load the CLI configuration.

src/config.rs
```diff
@@ -575,6 +576,15 @@ fn load_global_config_file() -> Result<(PathBuf, Ini)> {
...
+    let msg = format!("Failed to load {file_desc}.");
+    #[cfg(target_os = "macos")]
+    if xcode::launched_from_xcode() {
+        return msg + " Hint: Please ensure that ${SRCROOT}/.sentryclirc is added to the Input Files of this Xcode Build Phases script.";
```
szokeasaurusrex, line 582: Also nitpick: I slightly preferred your previous implementation with `push_str` here.

szokeasaurusrex (commented):
Two small nitpicks – you can choose whether or not to implement the changes – this code looks pretty good though! Great first PR!

src/config.rs
elramen: @szokeasaurusrex Clippy would give an error for having msg as mutable

szokeasaurusrex (commented):
Code looks good!

One last thing before merging, however: please make sure the PR title and comment follow the [Sentry commit naming conventions](https://develop.sentry.dev/commit-messages/). I specifically would recommend changing the title to "fix(xcode)" (lowercase). Please also make sure to reference the issue that this PR addresses at the end of the PR description, like so:

> Fixes GH-_<issue number here>_

szokeasaurusrex (approved):

Issue #2033, opened 2024-04-19
`debug-files check` should search directory recursively


szokeasaurusrex:
While investigating #2032, I discovered that when passed a path to a directory, `debug-files upload` recursively searches the directory for debug files. However, `debug-files check` treats the directory as if it were the debug file, and so the command fails with an "unsupported file" error, since a directory is clearly not a debug file that we would support.

Instead, we should recursively search the directory for debug files in the `debug-files check` command, to match the `upload` command's behavior. Ideally, we should reuse the recursive searching logic from the upload command.

Or, at a minimum, we should improve the error message when we `check` a directory. Something like the following could do:

> Error: `debug-files check` can only check individual debug files, but the provided path leads to a directory. Please a pass path to an individual file, instead.

PR #2034, opened 2024-04-19
Improve error message when we pass a directory to `debug-files check`

szokeasaurusrex:
See #2033.

Currently, the `debug-files check` command only supports checking individual files. However, the error message we emit when passing a directory is a very vague `error: Unsupported file`. This change improves the error message. We should still consider #2033's suggestion of supporting recursive search through directories for this command.

### DIFF
--- a/src/commands/debug_files/check.rs
+++ b/src/commands/debug_files/check.rs
@@ -43,6 +43,14 @@ pub fn make_command(command: Command) -> Command {
 pub fn execute(matches: &ArgMatches) -> Result<()> {
     let path = Path::new(matches.get_one::<String>("path").unwrap());
 
+    if path.is_dir() {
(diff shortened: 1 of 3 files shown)


szokeasaurusrex:
Thanks for the review, @elramen! @Swatinem and @loewenheim, I would also appreciate a review from one of you, since @elramen is new to CLI code reviews.

Swatinem (approved):
code looks good.
this is a bit surprising though, other commands treat the path argument as a glob pattern and will recurse into subdirectories and even zip files in some cases.

szokeasaurusrex:
@Swatinem yeah, that's what I thought too. I would like to implement similar behavior here, this is a quick fix until I get around to that!

PR #2044, opened 2024-04-26
ref(commands): Add clap derive parser

elramen:
Future commands (and existing ones by means of refactoring) can use the new derive parser to make the code easier to read, write, and modify, e.g., by keeping argument declarations and reads in sync.

### DIFF
--- a/src/commands/derive_parser.rs
+++ b/src/commands/derive_parser.rs
@@ -17,7 +17,7 @@ pub(super) struct SentryCLI {
 
     #[arg(global=true, ignore_case=true, value_parser=["trace", "debug", "info", "warn", "error"])]
     #[arg(long, help = "Set the log output verbosity")]
-    pub(super) log_level: Option<&'static str>,
+    pub(super) log_level: Option<String>,
 
     #[arg(global=true, action=SetTrue, visible_alias="silent", long)]
     #[arg(help = "Do not print any output while preserving correct exit code. \


src/utils/mod.rs
```diff
@@ -5,6 +5,7 @@ pub mod args;
 pub mod auth_token;
 pub mod chunks;
 pub mod cordova;
+pub mod derive_parser;
```
szokeasaurusrex, line 8: Regarding the naming/location of this module: to clarify, this is going to be shared between all commands using the new derive API, i.e. it is not specific to metrics?

I would also suggest that we consider moving this to somewhere inside `src/commands` since the logic is related to parsing commands, rather than being a general utility.

src/utils/value_parsers.rs
```diff
@@ -0,0 +1,9 @@
+use std::error::Error;
+
+/// Parse key:value pair from string, used as a value_parser for Clap arguments
+pub(crate) fn kv_parser(s: &str) -> Result<(String, String), Box<dyn Error + Send + Sync>> {
```
szokeasaurusrex, line 4: How are we parsing the existing headers argument? I assume we already have some code there that we could reuse here.

szokeasaurusrex (changes requested):
Nice PR! This is just a first pass review, I will likely have some more feedback depending on your answers to some of the questions I left. Looks nice overall though 😄

.vscode/settings.json
```diff
@@ -8,6 +8,6 @@
     "eslint.autoFixOnSave": true,
     "editor.tabSize": 4,
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
+        "source.fixAll.eslint": "explicit"
     }
-}
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: Thank you for attempting – this is not good though right now. It looks like you committed the deleting of the `.vscode/settings.json`, which might delete the local copy for folks when they pull these changes. We also should not need to modify the `.gitignore` for this repo. Let's troubleshoot this together in person on Monday 🙂

src/utils/derive_parser.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub auth_token: Option<AuthToken>,
+
+    #[arg(global=true, ignore_case=true, long, value_parser=["trace", "debug", "info", "warn", "error"], help="Set the log output verbosity")]
+    pub log_level: Option<String>,
+
+    #[arg(global=true, action=SetTrue, visible_alias="silent", long, help="Do not print any output while preserving correct exit code. This flag is currently implemented only for selected subcommands")]
+    pub quiet: bool,
+
+    #[arg(global=true, action=SetTrue, long, hide=true, help="Always return 0 exit code")]
+    pub allow_failure: bool,
```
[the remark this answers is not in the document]
szokeasaurusrex: Let's just keep as `pub` then. `pub(crate)` and `pub` are equivalent for the `sentry-cli`, since we only publish it as a binary program, not as a library. With either specifier, the item is available universally in the crate, but cannot be accessed from outside.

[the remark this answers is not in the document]
szokeasaurusrex: There would only be a difference between the two if this was a library – then, `pub(crate)` would mean the library can access this code, but external users of the library would be blocked. In this case `pub` would allow external library users to access the code

src/commands/derive_parser.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::utils::value_parsers::kv_parser;
+use clap::{command, value_parser, ArgAction::SetTrue, Parser, Subcommand};
+
+#[derive(Parser)]
+pub(super) struct SentryCLI {
+    #[command(subcommand)]
+    pub(super) command: SentryCLICommand,
+
+    #[arg(global=true, long="header", value_name="KEY:VALUE", value_parser=kv_parser, help="Custom headers that should be attached to all requests{n}in key:value format")]
+    pub(super) headers: Vec<(String, String)>,
```
szokeasaurusrex, line 11: I wonder whether we can use `Vec<&'static str, &'static str>` here, since I would expect the CLI arguments to outlive the program. But maybe not, idk

src/commands/derive_parser.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    #[arg(global=true, long, value_parser=value_parser!(AuthToken), help="Use the given Sentry auth token")]
+    pub(super) auth_token: Option<AuthToken>,
+
+    #[arg(global=true, ignore_case=true, long, value_parser=["trace", "debug", "info", "warn", "error"], help="Set the log output verbosity")]
+    pub(super) log_level: Option<String>,
```
szokeasaurusrex, line 17: Same here, would be cool if we can use `Option<&'static str>`. Will need some code that calls this to check if possible

src/commands/derive_parser.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    #[arg(global=true, action=SetTrue, visible_alias="silent", long, help="Do not print any output while preserving correct exit code. This flag is currently implemented only for selected subcommands")]
```
szokeasaurusrex, line 19: Nitpick: might be a good idea to split this line into multiple lines (same applies to other long `#[...]` directives in this file)

src/commands/derive_parser.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub(super) quiet: bool,
+
+    #[arg(global=true, action=SetTrue, long, hide=true, help="Always return 0 exit code")]
+    pub(super) allow_failure: bool,
+}
+
+#[derive(Subcommand)]
+pub(super) enum SentryCLICommand {}
```
szokeasaurusrex, line 27: What does this empty struct do? Do we even need it? If yes, please add a code comment so we remember the reason in the future 🙂

src/utils/value_parsers.rs
```diff
@@ -0,0 +1,9 @@
+use std::error::Error;
+
+/// Parse key:value pair from string, used as a value_parser for Clap arguments
+pub fn kv_parser(s: &str) -> Result<(String, String), Box<dyn Error + Send + Sync>> {
```
szokeasaurusrex, line 4: Following offline discussion, this should return a `Result<(String, String), CustomErrorType>`

Issue #2001, opened 2024-03-26
Add ability to send metrics via sentry-cli


AbhiPrasad:
Using `sentry-cli` is a nice way to send metrics in bash scripts and similar.

We already have support for [sending events](https://docs.sentry.io/product/cli/send-event/), so this seems like a nice fit.

Ideal API:

```bash
sentry-cli send-metric gauge -n cpu_usage -v 34 -t os:MacOS -u percent
```

`send-metric`: Command to send metric. Requires argument of `increment`, `distribution`, `set`, or `gauge` (matches with our sdk public API).

Options:
`-n`: Name of metric (required)
`-v`: Metric value (required)
`-u`: unit (optional)
`-t`: tag (optional, can provide multiple)

maybe we can also somehow capture code location?

szokeasaurusrex:
@jan-auer Is metrics rate limiting going to be relevant for the CLI, given that each send-metrics call will originate from a separate CLI call with the proposed API? If so, I think we will need to implement this change by somehow saving the backoff information in a temp file

AbhiPrasad:
Nice work @elramen! Looking forward to this 😄

PR #2045, opened 2024-04-29
feat(commands): add send-metric command

elramen:
Add CLI command that can emit metrics.

Fixes GH-2001


src/utils/metrics/payload.rs
```diff
@@ -0,0 +1,77 @@
...
+        .with_name(&metric_data.key)?
+        .with_unit(metric_data.unit)?
+        .with_value(metric_data.value)?
+        .with_type(metric_data.metric_type)?
+        .with_tags(metric_data.tags)?
+        .with_timestamp()
+    }
+}
+
+impl MetricPayload {
```
szokeasaurusrex, line 29: We can maybe combine this under MetricData

src/utils/metrics/metric_data.rs
```diff
@@ -0,0 +1,35 @@
+use crate::commands::send_metric::subcommands::{
+    CommonMetricArgs, FloatValueMetricArgs, IncrementArgs, SendMetricSubcommand, SetArgs,
+};
+
+pub(super) struct MetricData {
+    pub(super) key: String,
+    pub(super) value: f64,
```
szokeasaurusrex, line 7: We should make this an enum to separately handle ints and floats

src/utils/metrics/metric_data.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub(super) metric_type: char,
```
szokeasaurusrex, line 8: I would suggest making an enum with all the different metric types, and storing the enum here. Then, at serialization, we would convert to a char

src/utils/metrics/metric_data.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub(super) unit: Option<String>,
```
szokeasaurusrex, line 9: Are there a fixed set of allowed units? If so, may make sense to store as enum

src/utils/metrics/tags.rs
```diff
@@ -0,0 +1,76 @@
+use crate::{config::Config, utils::releases};
+use itertools::Itertools;
+use regex::Regex;
+use std::collections::HashMap;
+
+pub(super) struct NormalizedTags {
```
szokeasaurusrex, line 6: I think I would just call this struct something like `MetricsTags`, then have a function defined on the struct which normalizes all the tags

src/utils/value_parsers.rs
```diff
@@ -10,3 +10,11 @@ pub fn kv_parser(s: &str) -> Result<(String, String)> {
         s[pos + 1..].parse().expect("infallible"),
     ))
 }
+
+/// Parse the value argument for a set metric. Floats are floored and strings are hashed.
+pub fn set_value_parser(s: &str) -> Result<f64> {
```
szokeasaurusrex, line 15: Looks like this function can never return an error. If so, we should either return `f64` directly, or if we need a Result, we should return `Result<f64, Infallible>` to indicate that the function always succeeds. 

However, if we move to the enum for the metric value, maybe we will not need this function at all.

szokeasaurusrex (commented):
Initial quick-pass review (did not look at tests yet). Once this feedback is addressed, I'll do another more detailed review

Issue #2046, opened 2024-04-30
sentry-cli exits with an error, even when --no-upload flag is provided


markushi:
### CLI Version

2.31.0

### Operating System and Architecture

- [X] macOS (arm64)
- [ ] macOS (x86_64)
- [ ] Linux (i686)
- [ ] Linux (x86_64)
- [ ] Linux (armv7)
- [ ] Linux (aarch64)
- [ ] Windows (i686)
- [ ] Windows (x86_64)

### Operating System Version

Mac OS 14.3.1

### CLI Command

sentry-cli --log-level=trace upload-proguard --no-upload

### Exact Reproduction Steps

1. install cli
2. ensure no org / project / auth-token is setup
3. execute `sentry-cli --log-level=trace upload-proguard --no-upload`
```shell
sentry-cli --log-level=trace upload-proguard --no-upload
  INFO    2024-04-30 11:04:18.033176 +02:00 Loaded config from <readacted>/sentry-cli/.sentryclirc
  DEBUG   2024-04-30 11:04:18.033333 +02:00 sentry-cli version: 2.31.0, platform: "darwin", architecture: "arm64"
  INFO    2024-04-30 11:04:18.033360 +02:00 sentry-cli was invoked with the following command line: "sentry-cli" "--log-level=trace" "upload-proguard" "--no-upload"
error: Auth token is required for this request. Please run `sentry-cli login` and try again!
````

### Expected Results

Since the `--no-upload` flag was provided, no web requests should be made and a dry run is performed.

### Actual Results

CLI fails, breaking CI.

### Logs

Linked issue: https://github.com/getsentry/sentry-android-gradle-plugin/issues/697

PR #2047, opened 2024-04-30
Fix ensure authenticated API is only requested when no_upload flag is not present

markushi:
Fixes https://github.com/getsentry/sentry-cli/issues/2046


szokeasaurusrex (approved):
Took me a moment to understand what you did with the test code, but I think this looks good!

PR #2048, opened 2024-04-30
fix(commands): Remove misleading help messages

elramen:
Remove "Check debug output for transmission errors by passing --log-level=debug" from send-envelope's and send-event's help message as this functionality does not seem to exist, e.g., turning off wifi does not give the user any error info. I believe this is a consequence of the actual transmission being handled in sentry-core which does not propagate errors back to sentry-cli. 

Fixes GH-2049

### DIFF
--- a/src/commands/send_envelope.rs
+++ b/src/commands/send_envelope.rs
@@ -17,8 +17,7 @@ pub fn make_command(command: Command) -> Command {
             "Send a stored envelope to Sentry.{n}{n}\
              This command will validate and attempt to send an envelope to Sentry. \
              Due to network errors, rate limits or sampling the envelope is not guaranteed to \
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
@elramen, I am not convinced that it makes sense to make this change. Although it is possible that `--log-level=debug` does not provide helpful output when there is no network available, there should be output available for other error cases (e.g. HTTP 4xx or 5xx errors), and we typically ask users for `--log-level` output whenever they experience a bug to help diagnose the problem

szokeasaurusrex:
I think the bigger problem would be if we don't have a helpful error message when the network is unavailable. What happens when you run this command with WiFi disabled (what is the log output)?

szokeasaurusrex:
Following offline discussion, we decided to close this issue. Instead, we will work on getting error messages to be logged

Issue #2050, opened 2024-04-30
Add missing env vars for release name detection


elramen:
Looking through the SDK docs, it seems that environment variable SENTRY_RELEASE is commonly used to configure a release name. There is also an environment variable for Google App Engine used by the Python SDK. Is there a reason why the CLI does not use these? If not, they should be added to ```detect_release_name()```

PR #2051, opened 2024-04-30
fix(commands): Add missing env vars for release name detection

elramen:
Check env vars SENTRY_RELEASE and GAE_DEPLOYMENT_ID when detecting release name, as done by SDKs.

Fixes GH-2050

### DIFF
--- a/src/utils/releases.rs
+++ b/src/utils/releases.rs
@@ -85,7 +85,14 @@ pub fn infer_gradle_release_name(path: Option<PathBuf>) -> Result<Option<String>
 
 /// Detects the release name for the current working directory.
 pub fn detect_release_name() -> Result<String> {
-    // cordova release detection first.
+    // try SENTRY_RELEASE environment variable
+    if let Ok(release) = env::var("SENTRY_RELEASE") {
+        if !release.is_empty() {
+            return Ok(release);
+        }
+    }
+
+    // try cordova release detection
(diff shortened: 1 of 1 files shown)


src/utils/releases.rs
```diff
@@ -125,6 +132,13 @@ pub fn detect_release_name() -> Result<String> {
         }
     }
 
+    // try Google App Engine: https://cloud.google.com/appengine/docs/standard/python3/runtime#environment_variables
+    if let Ok(release) = env::var("GAE_DEPLOYMENT_ID") {
+        if !release.is_empty() {
+            return Ok(release);
+        }
+    }
+
```
szokeasaurusrex, line 141: Let's move this one to the end (just before the `match` statement)

src/utils/releases.rs
```diff
@@ -85,7 +85,14 @@ pub fn infer_gradle_release_name(path: Option<PathBuf>) -> Result<Option<String>
 
 /// Detects the release name for the current working directory.
 pub fn detect_release_name() -> Result<String> {
-    // cordova release detection first.
+    // try SENTRY_RELEASE environment variable
```
sl0thentr0py, line 88: if the comment is almost the same as the code below it, just leave out the comment

sl0thentr0py (approved):

PR #2059, opened 2024-05-02
feat(auth): Allow global config to be located in XDG directory

elramen:
If the global config file does not exist in the home directory, look in the following paths:
- Linux: $XDG_CONFIG_HOME/sentry/sentrycli.ini  and $HOME/.config/sentry/sentrycli.ini
- Mac: $HOME/Library/Application Support/sentry/sentrycli.ini
- Windows: {FOLDERID_RoamingAppData}

This enables users to keep their home dir clean, adhering to standards such as XDG on linux.

Fixes GH-1521

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,6 +14,7 @@ use log::{debug, info, set_max_level, warn};
 use parking_lot::Mutex;
 use sentry::types::Dsn;
 
+use crate::constants::CONFIG_INI_FILE_PATH;
 use crate::constants::DEFAULT_MAX_DIF_ITEM_SIZE;
 use crate::constants::DEFAULT_MAX_DIF_UPLOAD_SIZE;
 use crate::constants::{CONFIG_RC_FILE_NAME, DEFAULT_RETRIES, DEFAULT_URL};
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -527,10 +527,13 @@ impl Config {
...
 fn find_global_config_file() -> Result<PathBuf> {
     dirs::home_dir()
-        .ok_or_else(|| format_err!("Could not find home dir"))
-        .map(|mut path| {
-            path.push(CONFIG_RC_FILE_NAME);
-            path
+        .map(|p| p.join(CONFIG_RC_FILE_NAME))
+        .filter(|p| p.exists())
+        .or(dirs::config_dir()
+            .map(|p| p.join("sentry/sentrycli.ini"))
```
szokeasaurusrex, line 533: We should extract `"sentry/sentrycli.ini"` into a constant, like `CONFIG_RC_FILE_NAME`

src/config.rs
```diff
@@ -527,10 +527,13 @@ impl Config {
...
+            .filter(|p| p.exists()))
+        .ok_or_else(|| {
+            format_err!("Could not find config file. Please run `sentry-cli login` and try again!")
```
szokeasaurusrex, line 536:
```suggestion
            format_err!("Could not find config file.")
```

Why add the note about running `sentry-cli login` here? I am unsure whether this is always necessary – auth token is not needed for all CLI requests. Even if we should be adding this, this change should be made in a separate PR, since it is unrelated to what we want to do in this change.

sl0thentr0py (approved):

PR #2063, opened 2024-05-14
feat(metrics): Add send-metric command

elramen:
Add CLI command `send-metric` for emitting metrics to Sentry.
Add new command-parser that uses clap's [Derive API](https://docs.rs/clap/latest/clap/_derive/index.html). Future commands should use this as the Derive API makes it:
-Easier to read, write, and modify arguments and commands.
-Easier to keep the argument declaration and reading of argument in sync.
-Easier to reuse shared arguments.

Fixes GH-2001

### DIFF
--- a/src/utils/value_parsers.rs
+++ b/src/utils/value_parsers.rs
@@ -2,11 +2,7 @@ use anyhow::{anyhow, Result};
 
 /// Parse key:value pair from string, used as a value_parser for Clap arguments
 pub fn kv_parser(s: &str) -> Result<(String, String)> {
-    let pos = s
-        .find(':')
-        .ok_or_else(|| anyhow!(format!("`{s}` is missing a `:`")))?;
-    Ok((
-        s[..pos].parse().expect("infallible"),
-        s[pos + 1..].parse().expect("infallible"),
-    ))
+    s.split_once(':')
+        .map(|(k, v)| (k.into(), v.into()))
+        .ok_or_else(|| anyhow!("`{s}` is missing a `:`"))
 }


package.json
```diff
@@ -30,13 +30,13 @@
...
-    "@sentry/cli-win32-i686": "2.31.2",
-    "@sentry/cli-win32-x64": "2.31.2"
+    "@sentry/cli-darwin": "2.31.0",
+    "@sentry/cli-linux-arm": "2.31.0",
+    "@sentry/cli-linux-arm64": "2.31.0",
+    "@sentry/cli-linux-i686": "2.31.0",
+    "@sentry/cli-linux-x64": "2.31.0",
+    "@sentry/cli-win32-i686": "2.31.0",
+    "@sentry/cli-win32-x64": "2.31.0"
   },
```
szokeasaurusrex, line 40: Can you please rebase on master? We should not be reducing the version number here (or changing this file in any way)

src/commands/send_metric/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[command(about = "Send a value to a set metric")]
+    Set(SetArgs),
+}
+
+pub(super) fn make_command(command: Command) -> Command {
+    SendMetricSubcommand::augment_subcommands(command)
+}
+
+pub(super) fn execute(_: &ArgMatches) -> Result<()> {
+    let SentryCLICommand::SendMetric(SendMetricArgs { subcommand }) = SentryCLI::parse().command;
```
szokeasaurusrex, line 43: When we add more subcommands, this line won't compile, right? If so, maybe we should add a comment about how to fix this line when that happens

src/commands/send_metric/set.rs
```diff
@@ -0,0 +1,26 @@
...
+    utils::metrics::{
+        arg_parsers, normalized_payload::NormalizedPayload, types::MetricType, values::MetricValue,
+    },
+};
+use anyhow::Result;
+use clap::{command, Args};
+use sentry::protocol::EnvelopeItem;
+
+#[derive(Args)]
+pub(super) struct SetArgs {
```
szokeasaurusrex, line 13:
```suggestion
pub(super) struct SetMetricArgs {
```

I think this name is slightly clearer

src/utils/metrics/arg_parsers.rs
```diff
@@ -0,0 +1,27 @@
...
+
+// Parse the value argument for a set metric.
+pub fn set_value_parser(s: &str) -> Result<i64> {
+    match s.parse::<i64>() {
+        Ok(res) => Ok(res),
+        Err(_) => match s.parse::<f64>() {
+            Ok(_) => Err(anyhow!(format!("floats are not supported for set metrics"))),
+            Err(_) => Ok(crc32fast::hash(s.as_bytes()) as i64),
+        },
+    }
```
szokeasaurusrex, line 11: I think two things:
  - We should just treat floating point values as strings
  - It might be a good idea to force users to specify the type of data they are sending. Otherwise, a user who sends a string containing only digits will have that string treated as an int

szokeasaurusrex (commented):
I have only reviewed some of the files. Will continue reviewing the rest later

szokeasaurusrex:
Please also remember to rename this PR to "feat(metrics)"

src/utils/metrics/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod normalized_key;
+pub mod normalized_payload;
+mod normalized_tags;
+mod normalized_unit;
+pub mod types;
+pub mod values;
```
szokeasaurusrex, line 6: Please separate pub and private modules

src/utils/metrics/normalized_key.rs
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    #[test]
+    fn test_from() {
+        let expected = "aA1_-.____________";
+
+        let actual =
+            NormalizedKey::from(&MetricKey("aA1_-./+ö{😀\n\t\r\\| ,".to_string())).to_string();
+
+        assert_eq!(expected, actual);
```
szokeasaurusrex, line 39: Maybe more tests to verify edge cases here?

src/utils/metrics/normalized_payload.rs
```diff
@@ -0,0 +1,80 @@
...
+        let mut data = Vec::new();
+        write!(data, "{}", self.key)?;
+        write!(data, "@{}", self.unit)?;
+        write!(data, ":{}", self.value)?;
+        write!(data, "|{}", self.metric_type)?;
+        write!(data, "|#{}", self.tags)?;
+        let timestamp = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
+        write!(data, "|T{timestamp}")?;
+        Ok(data)
+    }
```
szokeasaurusrex, line 45: Could we perhaps do this inside a single `format!` statement, and then convert to bytes? It would really reinforce the fact that the ordering matters in that case. Would also allow us to avoid needing a mutable `data` variable

src/utils/metrics/normalized_payload.rs
```diff
@@ -0,0 +1,80 @@
...
+        commands::send_metric::common_args::{CommonMetricArgs, MetricKey},
+        config::Config,
+        utils::metrics::{
+            normalized_payload::NormalizedPayload, types::MetricType, values::MetricValue,
+        },
+    };
+    use regex::Regex;
+
+    #[test]
+    fn test_to_bytes() {
```
szokeasaurusrex, line 60: If we are just testing `to_bytes`, you could manually build the normalized payload

src/utils/metrics/normalized_payload.rs
```diff
@@ -0,0 +1,80 @@
...
+        Config::from_cli_config().unwrap().bind_to_process();
```
szokeasaurusrex, line 61: Why does this test depend on the Config being there? I don't think it should, otherwise it is not really a unit test

src/utils/metrics/normalized_payload.rs
```diff
@@ -0,0 +1,80 @@
...
+        .unwrap();
+
+        let bytes =
+            NormalizedPayload::from_cli_args(&common_args, MetricValue::Int(1), MetricType::Set)
+                .to_bytes()
+                .unwrap();
+        let actual = String::from_utf8_lossy(&bytes);
+
+        assert!(expected.is_match(&actual));
+    }
```
szokeasaurusrex, line 79: Let's also test separately the `from_cli_args` function

src/utils/metrics/normalized_tags.rs
```diff
@@ -0,0 +1,163 @@
+use crate::{config::Config, utils::releases};
+use itertools::Itertools;
+use regex::Regex;
+use std::{borrow::Cow, cmp::min, collections::HashMap};
+
+pub(super) struct NormalizedTags<'a> {
+    tags: HashMap<Cow<'a, str>, String>,
+}
+
+impl<'a> From<&'a Vec<(String, String)>> for NormalizedTags<'a> {
```
szokeasaurusrex, line 10: If I recall correctly, @Swatinem once told me that typically we use `From` to convert from an owned type. So, instead this should be 

```suggestion
impl<'a> From<Vec<(String, String)>> for NormalizedTags<'a> {
```

and the rest of the code should be changed correspondingly.

In any case, we should never take a reference to a `Vec`; instead, we should use a slice in such cases

src/utils/metrics/normalized_tags.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl std::fmt::Display for NormalizedTags<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let res = self
+            .tags
+            .iter()
+            .map(|(k, v)| format!("{}:{}", k, v))
+            .sorted()
+            .join(",");
+        write!(f, "{res}")
```
szokeasaurusrex, line 64: why not do this in one line?

szokeasaurusrex (commented):
Another partial review – I think I checked everything besides the integration tests. Mostly small changes/nitpicks, but I did find a bug related to string indexing which will need to be fixed

src/api/envelopes_api.rs
```diff
@@ -0,0 +1,45 @@
...
+impl EnvelopesApi {
+    pub fn try_new() -> ApiResult<EnvelopesApi> {
+        let api = Api::current();
+        api.config
+            .get_dsn()
+            .map(|dsn| EnvelopesApi { api, dsn })
+            .map_err(|_| ApiErrorKind::DsnMissing.into())
+    }
+
+    pub fn send_item(&self, item: EnvelopeItem) -> ApiResult<ApiResponse> {
```
iambriccardo, line 24: nit: Just a personal preference, I would rather prefer to have a method to build the `Envelope` given a set of `EnvelopeItem`(s) and then send it directly.

iambriccardo (commented):
LGTM, I will leave the approval to Daniel since I don't have all the CLI context I would need to properly review this PR.

szokeasaurusrex (commented):
I added some comments, and requested some changes, primarily related to adding clearer user-facing documentation. This time, I looked through all the files (including integration tests) – some of the changes I requested are in the files changed for the integration tests.

Overall, this PR seems to be coming together really well, and the code looks overall very good!

src/utils/value_parsers.rs
```diff
@@ -0,0 +1,12 @@
+use anyhow::{anyhow, Result};
+
+/// Parse key:value pair from string, used as a value_parser for Clap arguments
+pub fn kv_parser(s: &str) -> Result<(String, String)> {
+    let pos = s
+        .find(':')
+        .ok_or_else(|| anyhow!(format!("`{s}` is missing a `:`")))?;
+    Ok((
+        s[..pos].parse().expect("infallible"),
```
sl0thentr0py, line 9: I don't know idiomatic rust but isn't there some simple split function to do this

src/commands/derive_parser.rs
```diff
@@ -0,0 +1,36 @@
+use crate::utils::auth_token::AuthToken;
+use crate::utils::value_parsers::kv_parser;
+use clap::{command, value_parser, ArgAction::SetTrue, Parser, Subcommand};
+
+use super::send_metric::SendMetricArgs;
+
+#[derive(Parser)]
+pub(super) struct SentryCLI {
```
sl0thentr0py, line 8: can you write in the PR description (and thus in the final commit message) that this is a new way of writing future CLI commands (using the derives) since we were using a different pattern before?

sl0thentr0py (approved):
just a couple of comments, rest looks good

PR #2076, opened 2024-05-31
meta: Delete `stale.yml` workflow

szokeasaurusrex:
We have deleted this workflow in other repos (e.g. https://github.com/getsentry/sentry-python/pull/2245), and we should probably delete it here as well, by the same logic.


szokeasaurusrex:
Before merging this PR, we need to make sure that the new stalebot logic also applies to the `sentry-cli` repo. I found that the list of repos is defined [here](https://github.com/getsentry/eng-pipes/blob/3236a12d691c4bf26721b3aad95cc247daad3813/src/webhooks/pubsub/stalebot.ts#L12) as a concatenation of `SENTRY_MONOREPOS` and `SENTRY_SDK_REPOS`, but since I was unable to find where these constants are defined, I am unsure whether `sentry-cli` is included. If it is not included, maybe it should be

hubertdeng123:
Yep, `sentry-cli` is being monitored by stalebot in eng-pipes: https://github.com/getsentry/eng-pipes/blob/950e5365bba331bee1cbdf063ac2ef78ecd11856/github-orgs.yml#L55

This should be safe to do

hubertdeng123 (approved):

PR #2083, opened 2024-06-05
ref(api): Send events using EnvelopesApi instead of sentry-core

elramen:
Use sentry-cli's envelopes client, `EnvelopesApi`, instead of the sentry-core `Client`'s `capture_event`method. This change enables debug logging and non-zero exit codes for transmission failures, like the other commands.

Fixes GH-2060


szokeasaurusrex:
@elramen Should this PR be closed in favor of #2081?

src/commands/send_event.rs
```diff
@@ -159,14 +159,20 @@ pub fn make_command(command: Command) -> Command {
         )
 }
 
-fn send_raw_event(event: Event<'static>, dsn: Dsn) -> Uuid {
-    debug!("{:?}", event);
-    with_sentry_client(dsn, |c| c.capture_event(event, None))
+pub(super) fn send_raw_event(event: Event<'static>) -> Result<Uuid> {
```
Swatinem, line 162: since this is not taking a `Dsn` anymore, it just picks that up from the environment?

Swatinem (approved):
lgtm, though I wonder where the DSN is configured?

Issue #2077, opened 2024-06-03
API refactor: Split data types into separate submodules


szokeasaurusrex:
Data types are types which exist primarily for serializing/deserializing request/response data. Examples include `User` and `ReleaseStatus`.

szokeasaurusrex:
Or, perhaps it makes more sense to move this into `sentry-rust`

stephanie-anderson:
Maybe get @Swatinem or @loewenheim's opinion on that

Swatinem:
Also discussed with @elramen recently that IMO we shouldn’t bloat the SDK with types and definitions that are not directly useful for the SDK itself.

szokeasaurusrex:
@Swatinem I was considering that we might implement this change as a new feature under `sentry-types`, so as not to increase binary size when not needed.

However, you have a fair point – the code would be, at least initially, only used in the CLI, so probably makes more sense to keep the code in the CLI. We can always reevaluate later

szokeasaurusrex:
See comment on #2008

PR #2086, opened 2024-06-07
ref(api): Move chunk data types to submodule

szokeasaurusrex:
- Create a new `data_types` submodule under the `api` module.
 - Crate a `chunking` submodule under the `data_types` module.
 - Move all data types related to chunking from `src/api/mod.rs` to the `chunking` submodule, splitting their definitions further into submodules under `chunking`.

This PR only moves existing code, adding `mod` and `use` statements and adjusting visibility as needed. No other modifications were made.

Ref #2077

### DIFF
--- a/src/api/data_types/chunking/compression.rs
+++ b/src/api/data_types/chunking/compression.rs
@@ -1,0 +1,48 @@
+use std::fmt;
+
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, Default)]
+pub enum ChunkCompression {
+    /// No compression should be applied
+    #[default]
+    Uncompressed = 0,
+    /// GZIP compression (including header)
+    Gzip = 10,
+    /// Brotli compression
+    Brotli = 20,
+}
+
+impl ChunkCompression {
+    pub(in crate::api) fn field_name(self) -> &'static str {
+        match self {
+            ChunkCompression::Uncompressed => "file",
+            ChunkCompression::Gzip => "file_gzip",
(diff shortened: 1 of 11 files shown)


src/api/data_types/chunking/compression.rs
```diff
@@ -0,0 +1,48 @@
+use std::fmt;
+
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, Default)]
+pub enum ChunkCompression {
```
Swatinem, line 6: driveby question: how much effort would it be to add `zstd` compression to the whole workflow?

Swatinem (approved):

src/api/data_types/chunking/compression.rs
szokeasaurusrex: I am not sure – have not worked too extensively with compression in the CLI yet.

What would be the benefit of adding `zstd`? Do we have any docs written about it?

Swatinem: no, we don’t have any docs on that.
and `zstd` is pretty much the best choice if you want to trade compression ratio for resource usage.

szokeasaurusrex: Ok, we should probably investigate how much effort this would be, will open an issue

szokeasaurusrex:
@loewenheim Fair enough, I guess it comes down to personal preference a bit.

My reasoning is that each of these new modules has now a distinct responsibility; all code in a single module is closely related. When files get long, I have difficulty keeping track of everything in the file. It is much easier for me to work with many small files where I can clearly wrap my head around each one

PR #2099, opened 2024-07-15
build: Bump `curl` to `0.4.46` in `Cargo.toml`

szokeasaurusrex:
`curl` had already been updated to `0.4.44` in `Cargo.lock`; this commit updates the dependency in `Cargo.toml` to match.

ref #2097


Swatinem (approved):
this does not really have any effect, as the `Cargo.toml` only defines the minimum version, and the lockfile already has a more uptodate version.

PR #2101, opened 2024-07-16
docs(id-support): Document that -p and -o arguments accept slugs and IDs

iamrajjoshi:
With the API changes [we made to support ids as well as slugs](https://vanguard.getsentry.net/p/clyoo0m4a000u12nb53fkxanl), we get the added bonus that our CLI can also support ids.

Turns out the CLI also uses a couple endpoints which passed project slug in the body parameter, so we also had to update them:
https://github.com/getsentry/sentry/pull/74232
https://github.com/getsentry/sentry/pull/74371

With these prs, all the endpoints the CLI uses have ID support.

@szokeasaurusrex  and I had a conversation about how we would like to roll this change out in an [earlier closed pr](https://github.com/getsentry/sentry-cli/pull/2098), and decided that once we support ids in all the CLI commands, we can change all the help strings at once, which is what i am doing here.

### DIFF
--- a/src/api/errors/api_error.rs
+++ b/src/api/errors/api_error.rs
@@ -24,7 +24,7 @@ pub(in crate::api) enum ApiErrorKind {
     OrganizationNotFound,
     #[error("resource not found")]
     ResourceNotFound,
-    #[error("Project not found. Please check that you entered the project and organization slugs correctly.")]
+    #[error("Project not found. Please check that you entered the project and organization ids or slugs correctly.")]
     ProjectNotFound,
     #[error("release not found")]
     ReleaseNotFound,
--- a/src/api/errors/api_error.rs
+++ b/src/api/errors/api_error.rs
@@ -24,7 +24,7 @@ pub(in crate::api) enum ApiErrorKind {
     OrganizationNotFound,
     #[error("resource not found")]
     ResourceNotFound,
-    #[error("Project not found. Please check that you entered the project and organization ids or slugs correctly.")]
+    #[error("Project not found. Please check that you entered the project and organization IDs or slugs correctly.")]
     ProjectNotFound,
     #[error("release not found")]
     ReleaseNotFound,
--- a/src/config.rs
+++ b/src/config.rs
@@ -339,7 +339,7 @@ impl Config {
                 .get_from(Some("defaults"), "org")
                 .map(str::to_owned)
                 .ok_or_else(|| {
-                    format_err!("An organization id or slug is required (provide with --org)")
+                    format_err!("An organization ID or slug is required (provide with --org)")
                 }),
             (None, Some(cli_org)) => Ok(cli_org),
             (Some(token_org), None) => Ok(token_org.to_string()),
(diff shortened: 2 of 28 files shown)


src/api/errors/api_error.rs
```diff
@@ -24,7 +24,7 @@ pub(in crate::api) enum ApiErrorKind {
     OrganizationNotFound,
     #[error("resource not found")]
     ResourceNotFound,
-    #[error("Project not found. Please check that you entered the project and organization slugs correctly.")]
+    #[error("Project not found. Please check that you entered the project and organization ids or slugs correctly.")]
```
szokeasaurusrex, line 27:
```suggestion
    #[error("Project not found. Please check that you entered the project and organization IDs or slugs correctly.")]
```

src/config.rs
```diff
@@ -339,7 +339,7 @@ impl Config {
                 .get_from(Some("defaults"), "org")
                 .map(str::to_owned)
                 .ok_or_else(|| {
-                    format_err!("An organization slug is required (provide with --org)")
+                    format_err!("An organization id or slug is required (provide with --org)")
```
szokeasaurusrex, line 342:
```suggestion
                    format_err!("An organization ID or slug is required (provide with --org)")
```

Please update anywhere else you have used lowercase `id`.

szokeasaurusrex (commented):
Looks good, thanks for making these changes!

Before merging, please update "id" so that it is consistently capitalized, like "ID". Then, I will approve this PR!

src/config.rs
```diff
@@ -407,7 +407,7 @@ impl Config {
                     .get_from(Some("defaults"), "project")
                     .map(str::to_owned)
             })
-            .ok_or_else(|| format_err!("A project slug is required (provide with --project)"))
+            .ok_or_else(|| format_err!("A project id or slug is required (provide with --project)"))
```
szokeasaurusrex, line 410:
```suggestion
            .ok_or_else(|| format_err!("A project ID or slug is required (provide with --project)"))
```
Missed one

src/utils/args.rs
```diff
@@ -4,7 +4,7 @@ use clap::{Arg, ArgAction, Command};
 
 fn validate_org(v: &str) -> Result<String, String> {
     if v.contains('/') || v == "." || v == ".." || v.contains(' ') {
-        Err("Invalid value for organization. Use the URL slug and not the name!".to_string())
+        Err("Invalid value for organization. Use the URL slug or the ID and not the name!".to_string())
```
szokeasaurusrex, line 7:
```suggestion
        Err("Invalid value for organization. Use the URL slug or the ID, not the name!".to_string())
```

Think this would read more clearly

src/utils/args.rs
```diff
@@ -19,7 +19,7 @@ pub fn validate_project(v: &str) -> Result<String, String> {
         || v.contains('\t')
         || v.contains('\r')
     {
-        Err("Invalid value for project. Use the URL slug and not the name!".to_string())
+        Err("Invalid value for project. Use the URL slug or the ID and not the name!".to_string())
```
szokeasaurusrex, line 22:
```suggestion
        Err("Invalid value for project. Use the URL slug or the ID, not the name!".to_string())
```

szokeasaurusrex (approved):
One more suggestion, otherwise this looks good. Much simpler than your previous PR – thanks for getting the rest of the endpoints updated!

PR #2118, opened 2024-08-02
fix: Improve token redaction in CLI arg logging

szokeasaurusrex:
#2115 aimed to redact auth tokens when logging the arguments to the CLI. Although that change addressed some cases where auth tokens were passed as a CLI argument, not all cases were addressed. For example, the following was redacted properly with #2115:

```sh
sentry-cli --auth-token this-gets-redacted --log-level=info info
```

But, the following was not:

```sh
sentry-cli --auth-token=this-does-not-get-redacted --log-level=info info
```

The difference is that in the second example, the auth token is passed with `--auth-token=token` rather than separated by whitespace `--auth-token token`.

This change improves the redacting so that auth tokens passed like `--auth-token=token` are also redacted. The change also redacts any non-whitespace-containing substrings starting with `sntrys_` or `sntryu_` (prefixes that all auth tokens generated in the latest version of Sentry should start with), so that if an auth token appears where it is not expected, we redact it. For example, the following would be redacted with this change:

```sh
sentry-cli --auth=sntrys_my-token-passed-as-non-existing-auth-argument --log-level=info info
```

Note that as in #2115, this change is only relevant in the case where the log level is set to `info` or `debug` (the default is `warn`) – command line arguments are logged at the `info` level.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1,19 +1,18 @@
 //! This module implements the root command of the CLI tool.
 
-use std::io;
-use std::process;
-use std::{env, iter};
-
 use anyhow::{bail, Result};
 use clap::{value_parser, Arg, ArgAction, ArgMatches, Command};
 use clap_complete::{generate, Generator, Shell};
 use log::{debug, info, set_logger, set_max_level, LevelFilter};
+use std::borrow::Cow;
+use std::io;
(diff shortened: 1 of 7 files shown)


loewenheim (approved):
Very thorough!

Issue #2104, opened 2024-07-18
Always prefer URL from org auth token


szokeasaurusrex:
Currently, if the user manually provides a URL (e.g. via `--url`) which is different from the URL encoded in the org auth token they are using, the CLI exits with an error stating that the provided URLs do not match. This provides a poor user experience (see #2089, for example).

The URL encoded in the auth token is likely the URL that the user actually wishes to use, so instead of erroring when there is a URL mismatch, we should simply raise a warning (or maybe just an info message) whenever a user manually provides a URL in addition to an org auth token, which encodes a URL, and we should proceed with using the URL in the org auth token.

This would make it clear to users that setting a URL is unnecessary when using an org auth token, and would generally provide a better user experience.

PR #2122, opened 2024-08-06
feat: Prefer org auth token URL over manually provided URL

szokeasaurusrex:
Previously, we would error when the auth token's embedded URL differed from the manually provided URL. Now, we instead always use the token URL and log a warning.

Closes [#2104](https://github.com/getsentry/sentry-cli/issues/2104)

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -120,7 +120,7 @@ fn configure_args(config: &mut Config, matches: &ArgMatches) -> Result<()> {
     }
 
     if let Some(url) = matches.get_one::<String>("url") {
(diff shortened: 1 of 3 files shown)


loewenheim (approved):

jamesbeith:
Having upgraded from `2.33.1` to `2.34.1` I'm getting the following warning.

> WARN    2024-09-04 08:50:02.128067 +10:00 Using https://sentry.io (embedded in token) rather than manually-configured URL https://sentry.io/. To use https://sentry.io/, please provide an auth token for this URL.

Does that seem right, only the URLs are the same in the message?

PR #2132, opened 2024-08-27
fix(debugIds): Always instantiate global `Error` class in debugId injection snippet

Lms24:
This PR makes a small change to the JavaScript snippet that's injected into user files to set the debugId.

Previously, we'd instantiate `new Error` in the script, which usually implicitly instantiates `globalThis.Error`. However, if users or frameworks would create their own `Error` class in the very file where the debugId snippet is instantiated, `new Error` would of course take the user-created class rather than the global one. To fix this, this PR now explicitly instantiates the `globalThis.Error` class.

fixes: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/593


lforst (approved):

szokeasaurusrex (commented):
No objections to merging

PR #2162, opened 2024-09-23
feat: Use `globalThis` for code injection

timfish:
Since v8 of the JavaScript SDKs we dropped Node v12 support which means all the platforms we support have `globalThis`. 

This saves ~90 bytes every time debug IDs are injected.

- SDKs read from `globalThis` - https://github.com/getsentry/sentry-javascript/pull/11351
- Next.js SDK injects `globalThis` - https://github.com/getsentry/sentry-javascript/pull/13132
- Bundler plugins inject `globalThis` - https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/610

@lforst [points out](https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/610#issuecomment-2367988050) that this would require a major version bump of both the CLI and the bundler plugins so these bumps would likely need to occur at the same time and it might not be a priority to merge this straight away!


JakeShirley:
I just re-implemented this in #2271 before seeing this.  I'd love this change, as this would allow us to use Sentry in QuickJS contexts.  Without this QuickJS cannot use the debug injection (`global`, `window`, and `self` do not exist in ES2023 which is what QuickJS supports).

PR #2216, opened 2024-11-05
ref(utils): Remove `allow(dead_code)` from update utils

szokeasaurusrex:
Instead, mark all code that is not needed for the `managed` build with `cfg(not(feature = "managed"))`.

### DIFF
--- a/src/utils/update.rs
+++ b/src/utils/update.rs
@@ -1,12 +1,14 @@
-#![allow(dead_code)] // a ton of functions/fields might be unused based on the `managed` feature
-
+#[cfg(not(feature = "managed"))]
 use std::env;
 use std::fs;
 use std::io;
 use std::io::Write;
+#[cfg(not(feature = "managed"))]
 use std::path::Path;
 
-use anyhow::{bail, format_err, Result};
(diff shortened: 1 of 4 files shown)


src/utils/update.rs
```diff
@@ -175,10 +184,12 @@ pub fn get_latest_sentrycli_release() -> Result<SentryCliUpdateInfo> {
     })
 }
 
+#[cfg(not(feature = "managed"))]
 pub fn can_update_sentrycli() -> bool {
```
szokeasaurusrex, line 188: I wonder whether this function is even necessary. Ideally, we would distribute to package managers the `managed` build, so that this function would always return true.

src/utils/update.rs
```diff
@@ -52,6 +57,7 @@ fn rename_exe(exe: &Path, downloaded_path: &Path, elevate: bool) -> Result<()> {
     Ok(())
 }
 
+#[cfg(not(feature = "managed"))]
 #[cfg(not(windows))]
```
Swatinem, line 61: This might make sense to use boolean combinators

Swatinem (approved):

PR #2240, opened 2024-11-07
test(react_native): Bubble up `target_os`

szokeasaurusrex:
We can just mark the entire module as macos-only


loewenheim:
Why is any of that MacOS-specific in the first place?

szokeasaurusrex:
@loewenheim looks like all the tests are Xcode-specific, meaning they are only relevant to macOS

loewenheim (approved):

Issue #2223, opened 2024-11-07
Bump `mockito` to `1.x`


szokeasaurusrex:
Mockito 1.x contains features which we need for #2194, but it also breaks the API we use at the moment. Let's bump to the newer Mockito version to take advantage of the new features

PR #2246, opened 2024-11-14
ref(test): Refactor `EndpointOptions`

szokeasaurusrex:
Instead of using this struct to store options for the endpoint, which then manually need to be checked in the `mock_endpoint` method, convert the struct into a builder which constructs the mock with the provided configuration.

This will make `EndpointOptions` more flexible, allowing us to expand it with new features, and this also reduces branching.

Keeping `EndpointOptions` around, rather than simply calling `mockito` methods directly, will make it easier to handle any breaking changes to the `mockito` library, such as the ones that will occur when implementing #2223

### DIFF
--- a/tests/integration/test_utils/mocking.rs
+++ b/tests/integration/test_utils/mocking.rs
@@ -1,74 +1,56 @@
 use mockito::{Matcher, Mock};
 
+/// Options for creating a mock endpoint.
 pub struct EndpointOptions {
-    method: String,
-    endpoint: String,
-    status: usize,
-    response_body: Option<String>,
-    response_file: Option<String>,
-    matcher: Option<Matcher>,
-    header_matcher: Option<(&'static str, Matcher)>,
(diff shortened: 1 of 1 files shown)


tests/integration/test_utils/mocking.rs
```diff
@@ -1,74 +1,68 @@
...
+/// Options for creating a mock endpoint.
 pub struct EndpointOptions {
-    method: String,
-    endpoint: String,
-    status: usize,
-    response_body: Option<String>,
-    response_file: Option<String>,
-    matcher: Option<Matcher>,
-    header_matcher: Option<(&'static str, Matcher)>,
+    builder: Box<dyn FnOnce() -> Mock>,
```
Dav1dde, line 5: Do you need this closure here?  `builder: Mock` seems like it would be enough.

Dav1dde (approved):

PR #2263, opened 2024-11-20
ref(proguard): Delete pointless code

szokeasaurusrex:
We have some code which opens all the proguard files just to compute the SHA1 sum, which is just debug-logged and then added into a vector, which is never read from, and which is not needed for any lifetime purposes. We definitely can get rid of the vector, the debug log is also probably okay to get rid of because this information can easily be obtained using the `sha1sum` command


loewenheim (approved):
Nice find

PR #2271, opened 2024-11-26
fix: Add support for globalThis (ES2020) when injecting debug IDs to JavaScript

JakeShirley:
This PR adds an additional global object to look for when injecting debug IDs into JS files.

We are using the [QuickJS](https://github.com/bellard/quickjs) JavaScript engine which conforms to the ES2023 standard, and thus does not have `global`, `self`, or `window`, but it does have `globalThis`.

### DIFF
--- a/src/utils/sourcemaps/inject.rs
+++ b/src/utils/sourcemaps/inject.rs
@@ -14,7 +14,7 @@ use magic_string::{GenerateDecodedMapOptions, MagicString};
 use sentry::types::DebugId;
 use sourcemap::SourceMap;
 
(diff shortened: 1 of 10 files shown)


szokeasaurusrex:
@JakeShirley Thanks for the PR. However, it looks like this might be a duplicate of #2162. Could you please double check, and close your PR in favor of #2162 if that is the case?

gid-sentry:
Adding a comment here that this PR would be easier to push through compared to #2162 because it doesn't seem to be a breaking change. (unlike #2162) - globalThis is appended here instead of replacing the existing implementation.

PR #2280, opened 2024-11-27
test(chunk-upload): Test chunk upload with many chunks

szokeasaurusrex:
Add a test that verifies that chunk uploading works as expected for an upload which requires many (500+) chunks, split accross multiple (10) requests. This test also verifies that the chunk uploading code works for smaller chunks, since the way we achieve simulating a many-chunk upload without requiring a very large file to upload is by reducing the chunk size (in this case to 2 KiB).

ref #2194


szokeasaurusrex:
This PR is a duplicate of #2275 – I think I meant to push different changes here. Gonna find them then reopen

PR #2289, opened 2024-12-02
ref(chunks): Rename `Named` trait to `Assemblable`

szokeasaurusrex:
Rename the `Named` trait to `Assemblable`, in preparation for adding another method to get the Debug ID. We will need to be able to get the DebugID anywhere where we prepare a call to the assemble endpoint, and this is also the only place we need to have a name. So, it makes sense to have a single trait for both.

### DIFF
--- a/src/utils/chunks/mod.rs
+++ b/src/utils/chunks/mod.rs
@@ -7,7 +7,7 @@
 
 mod types;
 
-pub use types::{Chunked, MissingObjectsInfo, Named};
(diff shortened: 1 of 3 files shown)


loewenheim (approved):
Not a huge fan of the word "Assemblable", but can't really think of anything better either

szokeasaurusrex:
> Not a huge fan of the word "Assemblable", but can't really think of anything better either

I agree with you on both points 😅

PR #2301, opened 2024-12-05
feat(debugid): Update debug ID snippet to write on `globalThis` when available

lforst:
Similar to https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/636 we want to write debug IDs to `globalThis` for runtimes that do not support the other global objects.


szokeasaurusrex (approved):
@lforst This is not a breaking change, correct? 

If it is not, then I think we can merge this (although maybe someone more familiar with JS should also review)

lforst:
Correct, this is not a breaking change. Tagging @Lms24 for a review.

Lms24 (approved):

Issue #2297, opened 2024-12-04
Don't restrict chunk size to powers of two


szokeasaurusrex:
The Sentry CLI errors on chunk upload when the chunk size is not a power of two:

https://github.com/getsentry/sentry-cli/blob/87444d59cd35c01ea2ecbb9836e900f64434924c/src/utils/fs.rs#L175-L177

The reason for adding this check, according to @jan-auer, was that the storage we were using for uploaded chunks only supports chunk sizes that are powers of two. However, since the chunk size is provided by the server (it is not user-configurable), any such restriction can (and should) be enforced by the server. The Sentry CLI should use whatever chunk size the server requests, and should not perform any validation on the chunk size it receives (except perhaps, checking that the chunk size is nonzero, since a zero chunk size makes no sense).

With this check in place, if the server ever were to change to request a chunk size that is not a power of two, Sentry CLI users would no longer be able to perform chunk uploads since the Sentry CLI would error, and they would have no way to work around the error.

After removing the restriction in Sentry CLI, we should document in the Sentry codebase which Sentry CLI versions still have the power-of-two chunk size restriction. Anyone changing the chunk size in Sentry would then be aware which Sentry CLI versions would break if the chunk size is changed to a non-power-of-two value.

szokeasaurusrex:
Although this has no impact right now, having this check increases risk that someone unaware of this CLI-side restriction makes a change to the chunk size in Sentry, which breaks the CLI. Given that this is a low-effort change, we should prioritize implementation

PR #2311, opened 2024-12-12
test: Add tests for `get_sha1_checksums`

szokeasaurusrex:
In preparation for implementing #2297


szokeasaurusrex:
Makes more sense just to add this stuff in #2312

PR #2314, opened 2024-12-16
ref(chunks): Make `ChunkOptions` a `struct`

szokeasaurusrex:
Previously, `ChunkOptions` was implemented as a trait, but we realized it makes more sense to have it be a struct, instead. Now, as a struct, we also have `ChunkOptions` store the `ChunkServerOptions`, which will eventually allow us to simplify some API's, since we can take `ChunkOptions` only, rather than `DIFUpload` and `ChunkServerOptions` for methods such as `upload_difs_chunked`.


loewenheim (approved):
Much better IMO.

PR #2316, opened 2024-12-16
ref(chunks): Extract upload logic from `upload_difs_chunked`

szokeasaurusrex:
The `upload_difs_chunked` function contains logic specific to uploading DIFs, in addition to logic general to chunked uploads.

This change extracts the general chunked upload logic, so it can be reused for other types of chunked uploads.

Successor of #2303


src/utils/chunks/upload.rs
```diff
@@ -0,0 +1,332 @@
...
+    // Return only successful uploads
+    Ok((
+        successes.into_iter().filter_map(|(_, r)| r.dif).collect(),
+        has_errors,
+    ))
+}
+
+/// Renders the given detail string to the command line. If the `detail` is
+/// either missing or empty, the optional fallback will be used.
+fn render_detail(detail: &Option<String>, fallback: Option<&str>) {
```
szokeasaurusrex, line 315: This `&Option<String>` was copied from existing code

szokeasaurusrex:
Note to self (reason for marking as draft): Before merging, rename variables and modify comments as needed to make any references to DIFs generic to any uploaded object

loewenheim (approved):

PR #2324, opened 2024-12-19
build: Make backtraces useable in release builds

### LINKED ISSUE #2200: Remove `panic = abort` in `Cargo.toml`
Setting `panic` to `abort` disables any panic handlers in Rust. While having this does likely reduce binary size, the downside is that panics immediately crash Sentry CLI, and they cannot be handled.

If we add Sentry to the CLI (#1821), Sentry would not be able to capture panics due to this option.

szokeasaurusrex:
Having `panic = "abort"` and `strip = "symbols"` makes backtraces generated with `RUST_BACKTRACE=1` completely useless in release builds.

By changing these options to `panic = "unwind"` and `strip = "debuginfo"` makes backtraces usable in release builds. This will help us debug issues we are unable to reproduce, since we can ask users to set `RUST_BACKTRACE=1` and provide the logs with this option.

This change does somewhat significantly increase release build binary size (from 7.9 MB to 13.9 MB on my system). This is likely acceptable if it improves our ability to debug problems. If folks complain, we can consider distributing a separate "stripped" binary for users sensitive to large binary size.

Closes #2200

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
+[build]
+rustflags = ["-C", "force-unwind-tables"] # Make backtraces work.
+
 [target.'cfg(all(windows, target_env = "msvc"))']
 rustflags = ["-C", "target-feature=+crt-static"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ winapi = "0.3.9"
 [profile.release]
 opt-level = "z"
 panic = "abort"
-strip = true      # Strip symbols from the binary.
-codegen-units = 1 # Parallel compilation prevents some optimizations.
+strip = "debuginfo" # Only strip debuginfo (not symbols) to keep backtraces useful.
+codegen-units = 1   # Parallel compilation prevents some optimizations.
 # We do not enable link-time optimizations (lto) because they cause the
 # CLI to timeout when run in Xcode Cloud.


Swatinem (approved):
this looks reasonable.

Ideally, instead of enabling unwinding directly, another alternative would be to add `-C force-unwind-tables` to the `RUSTFLAGS` (see https://swatinem.de/blog/unwind-tables/#rust)

Another option to consider would be using `debug = "line-tables-only"`, which would give you a full-fidelity backtrace with inline frames, etc, but I bet that would inflate the binary size even more.

szokeasaurusrex:
> Ideally, instead of enabling unwinding directly, another alternative would be to add -C force-unwind-tables to the RUSTFLAGS (see https://swatinem.de/blog/unwind-tables/#rust)

@Swatinem So, if I understand correctly, with this option we would still have `panic = "abort"`? Would we expect the binary size to be smaller with this option?

> Another option to consider would be using debug = "line-tables-only", which would give you a full-fidelity backtrace with inline frames, etc, but I bet that would inflate the binary size even more.

I think let's just try with the current change first, I think it may already be helpful enough for what I need it for (mostly, I just would like to be able to find where errors originating from libraries are coming from. Usually if the error is emitted in Sentry CLI, I can find the error by searching for the error string).

Swatinem:
> with this option we would still have panic = "abort"? Would we expect the binary size to be smaller with this option?

Yes, that codegen option adds the necessary unwind info to the binary so that backtraces work, but it avoids generating all the code to safely unwind in Rust, like destructors, `catch_unwind`, etc.

szokeasaurusrex:
Alright, I went with your suggestion to set the `-C force-unwind-tables` instead of `panic = "unwind"`. With this change, the release binary size (on my system) only increased to 12.2 MB, instead of 13.9 MB.

PR #2331, opened 2025-01-02
feat(dif): Fail `debug-files upload` when file is too big

### LINKED ISSUE #2313: Error when debug file too big to upload
There is a file size limit for debug files that are uploaded to Sentry (c.a. 2 GB). When a file exceeds this limit, we log a warning to inform users that the file was skipped:

https://github.com/getsentry/sentry-cli/blob/1d5d59341c4b15067a80745fd6ee8be04d1d581e/src/utils/dif_upload.rs#L2049-L2054

This warning is easy to miss. So, instead of just warning users, we should error and exit with a nonzero status. Our error message can link users to the [Symbol Server](https://docs.sentry.io/platforms/unreal/data-management/debug-files/symbol-servers/) docs, and also inform them about the ability to set `SENTRY_ALLOW_FAILURE` to unblock build.

szokeasaurusrex:
Previously, we only logged an easy-to-miss warning when we had to skip uploading a debug file because it was too big. Now, we instead fail the entire upload loudly, to ensure users do not miss this important information.

Depends on:
  - #2330 
  - #2329 

Resolves #2313

### DIFF
--- a/tests/integration/debug_files/upload.rs
+++ b/tests/integration/debug_files/upload.rs
@@ -731,3 +731,19 @@ fn chunk_upload_multiple_files_small_chunks_only_some() {
         "Uploaded chunks differ from the expected chunks"
     );
 }
+
+#[test]
+fn test_dif_too_big() {
+    TestManager::new()
+        .mock_endpoint(
+            MockEndpointBuilder::new("GET", "/api/0/organizations/wat-org/chunk-upload/")
+                .with_response_file("debug_files/get-chunk-upload-small-max-size.json"),
+        )
+        .assert_cmd(vec![
+            "debug-files",
+            "upload",
+            "tests/integration/_fixtures/SrcGenSampleApp.pdb",
+        ])
+        .with_default_token()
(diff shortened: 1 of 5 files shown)


tests/integration/debug_files/upload.rs
```diff
@@ -731,3 +731,16 @@ fn chunk_upload_multiple_files_small_chunks_only_some() {
...
+
+
+#[test]
+fn test_dif_too_big() {
+    TestManager::new()
+        .mock_endpoint(
+            MockEndpointBuilder::new("GET", "/api/0/organizations/wat-org/chunk-upload/")
+                .with_response_file("debug_files/get-chunk-upload-small-max-size.json"),
+        )
+        .assert_cmd(vec!["debug-files", "upload", "tests/integration/_fixtures/SrcGenSampleApp.pdb"])
```
szokeasaurusrex, line 743: This file is about 11 KiB, exceeding the maximum size of 2 KiB

PR #2337, opened 2025-01-08
ref(api): Replace custom deserializer with derive

szokeasaurusrex:
This custom deserializer can be easily replaced with `derive(Deserialize)`.

### DIFF
--- a/src/api/data_types/chunking/compression.rs
+++ b/src/api/data_types/chunking/compression.rs
@@ -1,16 +1,18 @@
 use std::fmt;
 
-use serde::{Deserialize, Deserializer};
+use serde::Deserialize;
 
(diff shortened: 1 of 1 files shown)


src/api/data_types/chunking/compression.rs
```diff
@@ -1,16 +1,18 @@
...
-    /// No compression should be applied
-    #[default]
-    Uncompressed = 0,
     /// GZIP compression (including header)
     Gzip = 10,
     /// Brotli compression
     Brotli = 20,
+    /// No compression should be applied
+    #[default]
+    #[serde(other)]
```
szokeasaurusrex, line 14: `#[serde(other)]` has to come last. I don't think that reordering this enum should have any effect, since we already manually set the enum values, and this PR keeps those the same

PR #2343, opened 2025-01-14
ref(sourcemaps): Reduce sourcemap upload memory usage

### LINKED ISSUE #2342: Prevent unnecessary memory allocation in sourcemaps uploading
The following line creates a copy of all of the sourcemaps being uploaded with the `sourcemaps upload` command:

https://github.com/getsentry/sentry-cli/blob/739bc00573c9e024cb7e1b199a476a6e385623bf/src/utils/file_upload.rs#L189

If the files being uploaded are large, this will consume a lot of memory.

It should be possible to eliminate this memory allocation.

szokeasaurusrex:
Prevent an unnecessary copy of all sourcemap files being uploaded from being stored in memory, by using an `Arc` for the `contents` field of `SourceFile`.

Depends on:
  - #2340
  - #2341

Fixes #2342

### DIFF
--- a/src/commands/debug_files/bundle_jvm.rs
+++ b/src/commands/debug_files/bundle_jvm.rs
@@ -12,6 +12,7 @@ use std::collections::BTreeMap;
 use std::fs;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 use symbolic::debuginfo::sourcebundle::SourceFileType;
 
 pub fn make_command(command: Command) -> Command {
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
The memory is saved in the clone on this line (also linked in #2342):

https://github.com/getsentry/sentry-cli/blob/739bc00573c9e024cb7e1b199a476a6e385623bf/src/utils/file_upload.rs#L189

By using an `Arc`, we prevent creating another copy of the sourcemaps we are uploading.

loewenheim (approved):

PR #2378, opened 2025-02-10
build: `yarn upgrade`

szokeasaurusrex:
Bump our out of date JS dependencies


lforst:
This is slightly scary. I'd rather we bump individual deps instead of blanket bumping all transitive ones 🤔

Issue #1503, opened 2023-03-06
publish docker image to aws ecr public gallery


benconnito:
docker hub heavily rates limits anonymous users. using the sentry cli via a docker image pulling from docker hub often hits these rate limits in ephemeral ci systems (codepipeline+codebuild). can we get the sentry cli docker image published to ecr public gallery (https://gallery.ecr.aws) as a verified account?

benconnito:
it certainly wouldnt hurt. im sure those who use it on gcp would benefit from that and depending on gcr rate limiting it might suffice. anything other than docker hub would help for sure.

asottile-sentry:
just to disambiguate -- kamil is suggesting g**h**cr -- github's container registry (which would be pretty trivial to set up here -- compared to setting up an aws account for ecr)

benconnito:
the main idea is to avoid docker hubs rate limiting for anonymous users. and have official/verified images on other registries (gcr,ecr). there are images of sentry cli on ecr public but they are not official - ie https://gallery.ecr.aws/?searchTerm=getsentry

szokeasaurusrex:
As of #2393, we now publish Docker images to GHCR, which should have less strict rate limits than Docker Hub. We don't plan to publish to AWS ECR public gallery for now, since the GHCR image should be sufficient for most users.

PR #2393, opened 2025-02-19
ci(docker): Release Docker image to GHCR

szokeasaurusrex:
We already build and push our Docker images to GHCR, tagged with commit hash, during the Release Build action. This PR adds two new actions which will tag these images with the version number and with the "latest" once the GitHub release is created. We have two separate actions for each, since we only want to add the "latest" tag for releases, not for preleases, but the version tag should be created for both.

Ref #1503

### DIFF
--- a/.github/workflows/release-ghcr-latest-tag.yml
+++ b/.github/workflows/release-ghcr-latest-tag.yml
@@ -1,0 +1,22 @@
+name: Release GHCR Latest Image
+
+on:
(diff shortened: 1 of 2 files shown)


asottile-sentry (approved):
could probably combine the two but this works!  also iirc there's a way to retag without pushing and pulling through `docker manifest something something` but this is probably "good enough" and works as is

PR #2397, opened 2025-02-19
Fix bundle file count report

### LINKED ISSUE #2136: Skipped files still included in source maps upload report and bundled file count
## Reproduction

```sh
sentry-cli sourcemaps upload example/ --log-level=info
```

where the `example/` directory contains at least one non-UTF encoded sourcemap file.

## Expected behavior

The non-UTF-8 file is neither included in the count of files bundled for upload nor in the source map upload report, as it is skipped.

## Actual result

We skip the UTF-8 file, but it is included in both the bundled file count and in the source map upload report

a-hariti:
some files fail to be added to the source bundle for different errors,

but the report shows the count of the original number of file that would to be added if no errors occurred.

closes #2136


tests/integration/sourcemaps/upload.rs
```diff
@@ -183,3 +183,11 @@ fn command_sourcemaps_upload_complex_extension() {
         .register_trycmd_test("sourcemaps/sourcemaps-upload-complex-extension.trycmd")
         .with_default_token();
 }
+
+#[test]
+fn command_sourcemaps_upload_skip_invalid_utf8() {
+    TestManager::new()
+        .mock_common_upload_endpoints(ServerBehavior::Modern, Default::default())
+        .register_trycmd_test("sourcemaps/sourcemaps-with-invalid-utf8.trycmd")
+        .with_default_token();
```
szokeasaurusrex, line 192: Did you check whether the `mock_common_upload_endpoints` is actually needed here? If not, please check quickly before merging, and remove if it's unnecessary (there's a good chance it will be needed, though).

If you don't need the `mock_common_upload_endpoints`, you likely won't need the `with_default_token()`, either.

szokeasaurusrex (approved):
Looks good, see comment prior to merge!

PR #2401, opened 2025-02-20
build(macos): Sign macOS binaries

### LINKED ISSUE #1882: sentry-cli for macOS doesn't seem to be signed properly
### Environment

macOS Sonoma 14.0, sentry-cli 2.23.1
cli executable downloaded directly from GitHub releases page [sentry-cli-Darwin-arm64](https://github.com/getsentry/sentry-cli/releases/download/2.23.1/sentry-cli-Darwin-arm64)

### Steps to Reproduce

What you did?
Downloaded the cli, renamed the file to `sentry-cli`, then `chmod 755 sentry-cli`, and tried to run it from the terminal.

### Expected Result

What you thought would happen?

Cli to run with no issues, or only with a warning that it has been downloaded from the internet, and it has passed anti-virus checks.

### Actual Result

What actually happened?

image

### Logs

N/A

szokeasaurusrex:
Closes #1882

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,7 +147,7 @@ jobs:
           input_path: sentry-cli-Darwin-${{ matrix.arch }}
           p12_file: ${{ env.APPLE_CERT_PATH }}
           p12_password: ${{ secrets.APPLE_CERT_PASSWORD }}
-          sign_args: --code-signature-flags runtime
+          sign_args: --for-notarization
 
       - name: Zip signed binary
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,6 +147,7 @@ jobs:
           input_path: sentry-cli-Darwin-${{ matrix.arch }}
           p12_file: ${{ env.APPLE_CERT_PATH }}
           p12_password: ${{ secrets.APPLE_CERT_PASSWORD }}
+          sign_args: '--code-signature-flags runtime'
 
       - name: Zip signed binary
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,10 +133,13 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: unsigned-bin-macos-${{ matrix.arch }}
-          path: sentry-cli-Darwin-${{ matrix.arch }}
 
-      - name: Move binary to root
-        run: mv sentry-cli-Darwin-${{ matrix.arch }}/sentry-cli-Darwin-${{ matrix.arch }} sentry-cli-Darwin-${{ matrix.arch }}
+      - name: Debug
+        run: |
+          echo "Current directory:"
+          pwd
+          echo "Files in current directory:"
+          ls -Rla .
 
       - name: Sign binary
         uses: indygreg/apple-code-sign-action@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,11 @@ jobs:
       APPLE_API_KEY_PATH: /tmp/apple_key.json
 
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: sentry-cli
+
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -91,18 +91,83 @@ jobs:
...
+      APPLE_API_KEY_PATH: /tmp/apple_key.json
+
+    steps:
+      - name: Decode Apple signing certificate and API key
+        env:
+          APPLE_CERT_DATA: ${{ secrets.APPLE_CERT_DATA }}
+          APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
+        run: |
+          echo $APPLE_CERT_DATA | base64 --decode > ${{ env.APPLE_CERT_PATH }}
+          echo $APPLE_API_KEY | base64 --decode > ${{ env.APPLE_API_KEY_PATH }}
```
BYK, line 130: Safer
```suggestion
          echo "$APPLE_CERT_DATA" | base64 --decode > ${{ env.APPLE_CERT_PATH }}
          echo "$APPLE_API_KEY" | base64 --decode > ${{ env.APPLE_API_KEY_PATH }}
```

BYK (approved):
I think this looks pretty great!

It might be worth adding a smoke test after signing and notarization that just runs the binary and makes sure it doesn't exit with a non-zero code.

.github/workflows/build.yml
```diff
@@ -91,18 +91,91 @@ jobs:
...
+      APPLE_API_KEY_PATH: /tmp/apple_key.json
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install `rcodesign`
+        run: |
+          curl -L https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign%2F0.29.0/apple-codesign-0.29.0-x86_64-unknown-linux-musl.tar.gz | tar -xz --strip-components=1
+          mv rcodesign /usr/local/bin/rcodesign
```
mdtro, line 130: Recommend adding a step here that verifies the expected SHA256 checksum.

```
https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign%2F0.29.0/apple-codesign-0.29.0-aarch64-unknown-linux-musl.tar.gz.sha256
```

```
4af92c87ddf52f5f2d1258a3b4e56c7dcb8f1b2468df744976c5f139e031961f
```

BYK: I think this is a good idea as I forgot releases can be mutable.

BYK: Added to Spotlight here: https://github.com/getsentry/spotlight/pull/722/files

BYK: Note that the SHA256 @mdtro shared is for arm64 so different in my PR.

mdtro: Oh woops. That's my mistake. 😅

BYK: No worries, I made the same mistake even when I thought I was looking _very_ closely :D

.github/workflows/build.yml
```diff
@@ -91,18 +91,88 @@ jobs:
...
+    needs: [macos, macos_universal]
+    name: Sign & Notarize macOS Binary (${{ matrix.arch }})
+    runs-on: ubuntu-24.04
+
+    env:
+      APPLE_CERT_PATH: /tmp/certs.p12
+      APPLE_API_KEY_PATH: /tmp/apple_key.json
+
+    steps:
+      - name: Checkout repository
```
BYK, line 124: ~~Don't think you need a check out as you're not using anything from the repo?~~

I see that we use entitlements from the repo. Not sure how long this step takes but if it's longer than a few seconds we can upload entitlements file along with the artifacts and skip this step.

BYK (approved):
LGTM!

.github/workflows/build.yml
szokeasaurusrex: I don't think it takes that long – the waiting for notarization I think takes quite a bit longer

PR #2418, opened 2025-03-18
ci(docker): Fix GHCR releases so they are multiarch

szokeasaurusrex:
Our current release process for GHCR makes the images tagged with the version number and "latest" only be single-arch images, rather than keeping them as multi-arch images.

This change _should_ fix this behavior.


tobias-wilfert (approved):
This should work, you can manually test it on a different trigger if you want see if the `buildx` works  (see: https://github.com/getsentry/relay/pull/4532).

PR #2421, opened 2025-03-20
build: update `zip` dependency

szokeasaurusrex:


loewenheim (approved):
Updating zip had a big (positive) performance impact in Symbolicator.

PR #2424, opened 2025-03-20
ci: Stop using `actions-rs`

szokeasaurusrex:
`actions-rs` is no longer maintained. We can instead just call Rust commands directly.

This change also removes explicit installs of Rust toolchains, since GitHub Runners ship with the needed Rust tools installed. Instead, we only use `rustup target add` when needed for cross-compilation to other targets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,18 +61,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


szokeasaurusrex (commented):
I also pushed this branch to `release/test`, so you can see the results of the "Release Build" (defined in `build.yml`) also on this commit

loewenheim (approved):

PR #2427, opened 2025-03-21
ci: Update and securely pin all actions

szokeasaurusrex:
Many of the actions we are using were pinned to old versions. This PR updates all actions to the latest avaialable versions, and securely pins them using their full hashes, which is a [recommended security practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), since version tags are mutable.

A future PR will add a Dependabot configuration which will periodically check for new actions releases, and open PRs to update the actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


szokeasaurusrex (commented):
Will also push this to a `release/` branch to test the Release Build workflow

loewenheim (approved):

Issue #1426, opened 2023-01-04
Build for Windows on ARM


mattjohnsonpint:
Since [Windows on ARM](https://learn.microsoft.com/windows/arm/overview) is now a thing, and we already build releases for ARM64 for Linux and MacOS, it would probably be worth building an ARM64 release for Windows as well.

## Implementation notes

- Seems like we need to get rid of our dependency on `winapi` (including the indirect dependency through `username`). This crate does not compile on Windows on ARM. It is no longer maintained. Instead, there is now an official [windows-rs](https://github.com/microsoft/windows-rs) crate from Microsoft

wojtekmaj:
In the light of multiple ARM PCs introduced to the market recently, like Surface Pro 9, I'd like to give this issue a bump. This issue is only going to become more relevant now.

szokeasaurusrex:
Hey @wojtekmaj, thank you for reaching out. I have increased the priority of this issue in our backlog; however, since our team is quite busy with other projects, it could still take us some time to get around to this.

Galkon:
Bump, we would love to be able to use this at Highlight AI.

szokeasaurusrex:
Hi all, thank you for your patience with this one.

We just released version [2.42.5 of Sentry CLI](https://github.com/getsentry/sentry-cli/releases/tag/2.42.5). This release includes a build for Windows on ARM, which you can download from the link provided.

PR #2431, opened 2025-03-21
build: Remove direct `winapi` dependency

szokeasaurusrex:
We can use the `windows-sys` crate (published by Microsoft and actively maintained) instead.

Ref #1426


szokeasaurusrex:
We still need to remove indirect dependencies on this crate to unblock #1426 fully

loewenheim (approved):

PR #2432, opened 2025-03-21
build: Replace `username` with `whoami`

szokeasaurusrex:
`username` is completely unmaintained, and it depends on `winapi`, preventing us from supporting Windows on ARM. Use `whoami` instead to get the curent user.

Ref #1426

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,16 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


szokeasaurusrex (commented):
I tested this change manually to ensure `whoami` gets the correct username when using the `sentry-cli send-event` command. Example event here: https://sentry-sdks.sentry.io/issues/6438150058/events/a04c3b53975748abbaff4692a1bb977a/.

szokeasaurusrex:
There are still some remaining transitive dependencies on `winapi`, but this PR might actually be sufficient to unblock https://github.com/getsentry/sentry-cli/issues/1426, since the remaining `winapi` transitive dependencies are on a newer version of `winapi`.

loewenheim (approved):

PR #2448, opened 2025-03-27
ref(sourcemaps): Remove `SourceMapReference` struct

szokeasaurusrex:
Following #2447, we no longer ever set the `SourceMapReference` struct's `original_url` field to any `Some(_)` value. This essentially makes the struct unnecessary, since we can just pass the URL values around as `String`s.

Depends on:
  - #2447


loewenheim (approved):

szokeasaurusrex:
#2447 was changed between this PR being opened and that PR being merged; now we in fact still need the `SourceMapReference` struct, so I will close this PR

Issue #2451, opened 2025-04-07
Support multiple SENTRY_DOTENV_PATH


Kinqdos:
Hey there I would love the see support for multiple SENTRY_DOTENV_PATH files. We have a setup with `.env` and `.env.local`. I would like to put the SENTRY_URL in the public `.env` file and the SENTRY_TOKEN in the local `.env.local`

Somthing like `SENTRY_DOTENV_PATH=.env,.env.local` would be great.

szokeasaurusrex:
Hi @Kinqdos, thank you for the suggestion, I can put this issue on our backlog.

It might take some time for us to implement this, however, since we have other priorities. PRs are always welcome if you would like to see this feature implemented more quickly!

Kinqdos:
Hi @szokeasaurusrex thank you for the response. I don't have much expierence with rust, but I will give it a try. What do you think about a comma as a separator for the files?

szokeasaurusrex:
@Kinqdos I just realized this change should be pretty simple, I can do it for you

Kinqdos:
@szokeasaurusrex Just created a pr, if you didn't finished yet, you can also use mine ;)

szokeasaurusrex:
@Kinqdos replied on the PR, let's continue the conversation over there

szokeasaurusrex:
Fixed in #2454

PR #2454, opened 2025-04-08
Support multiple files in SENTRY_DOTENV_PATH

Kinqdos:
Added support for multiple files in SENTRY_DOTENV_PATH variable. Multiple files are separated with a comma and override each other.

See #2451

### DIFF
--- a/src/utils/system.rs
+++ b/src/utils/system.rs
@@ -110,37 +110,35 @@ pub fn load_dotenv() -> DotenvResult<()> {
         return Ok(());
     }
 
-    match env::var("SENTRY_DOTENV_PATH") {
-        Ok(path) => load_dotenv_from_path(&*path),
-        Err(_) => dotenvy::dotenv().map(|_| ()),
+    let mut paths: Vec<String> = Vec::new();
+
+    if let Ok(path) = env::var("SENTRY_DOTENV_PATH") {
+        paths.push(path)
     }
-    .map_or_else(
-        |error| {
-            // We only propagate errors if the .env file was found and failed to load.
-            if error.not_found() {
-                Ok(())
-            } else {
-                Err(error)
-            }
-        },
-        |_| Ok(()),
-    )
-}
 
-/// Loads all env files from the given path (use comma as separator)
-fn load_dotenv_from_path(path: &str) -> DotenvResult<()> {
-    let files = path.split(",").map(|file| file.trim());
+    if let Ok (path) = env::var("SENTRY_DOTENV_PATHS") {
+        paths.extend(path.split(",").map(|file| file.trim().to_string()));
+    }
 
-    for file in files {
-        if let Err(err) = dotenvy::from_path_override(file) {
-            // We only propagate errors if the .env file was found and failed to load.
-            if !err.not_found() {
-                return Err(err);
-            }
+    // Fallback to default dotenv
+    if paths.is_empty() {
+        check_dotenv_result(dotenvy::dotenv().map(|_| ()))?;
+    } else {
+        for path in paths {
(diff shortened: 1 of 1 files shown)


szokeasaurusrex (commented):
@Kinqdos thanks for the contribution!

While I think the comma-separation makes sense, changing `SENTRY_DOTENV_PATH` to accept comma-separated paths would be a breaking change, since file names on UNIX can contain commas. For example, if someone currently sets `SENTRY_DOTENV_PATH="path/with,comma/.env"`, your proposed change would cause this to be read as two separate paths (`path/with` and `comma/.env`) instead of as the single path `path/with,comma/.env`. 

I think we may need to introduce a separate environment variable, such as `SENTRY_DOTENV_PATHS` (plural) which can take multiple files separated by commas. That way, if someone has an edge-case setup with dotenv path containing a comma, their setup won't break, but it is still possible to support your use case via the new environment variable

Kinqdos:
@szokeasaurusrex Thank you for the feedback, I will make a new version ready. The plural variant should be preferred over the singular variant if both is set right?

szokeasaurusrex:
@Kinqdos i think either order of precedence is fine, please pick whatever is easiest for you to implement

Kinqdos:
@szokeasaurusrex I pushed the new version, looking forward for your feedback.

szokeasaurusrex:
@Kinqdos I will try to take a look at some point this week. Unfortunately I currently have higher priority items on my list

Kinqdos:
Hey @szokeasaurusrex just a friendly reminder, it would be great if you can take a look :)

src/utils/system.rs
```diff
@@ -110,21 +110,35 @@ pub fn load_dotenv() -> DotenvResult<()> {
...
-        Err(_) => dotenvy::dotenv().map(|_| ()),
+    let mut paths: Vec<String> = Vec::new();
+
+    if let Ok(path) = env::var("SENTRY_DOTENV_PATH") {
+        paths.push(path)
+    }
+
+    if let Ok (path) = env::var("SENTRY_DOTENV_PATHS") {
+        paths.extend(path.split(",").map(|file| file.trim().to_string()));
+    }
```
szokeasaurusrex, line 121: I would use an `else if let` for the second `if` statement, so that if both environment variables are provided, we only consider `SENTRY_DOTENV_PATH`. Doing this will also allow you to simplify the code, likely removing the need for `paths` to be mutable.

szokeasaurusrex:
Hey @Kinqdos, thanks for your patience here. I left a comment; please also check the linter errors. I can also take over the PR from here if you would prefer

Kinqdos:
If it speeds up the merge you can take over now.

src/utils/system.rs
Kinqdos: I thought it would be nicer to consider both variables. But I would insist on this

src/utils/system.rs
```diff
@@ -110,21 +109,29 @@ pub fn load_dotenv() -> DotenvResult<()> {
...
+        &paths
+            .split(",")
+            .map(|path| path.trim())
+            .filter(|path| !path.is_empty())
+            .map(|path| path.to_string())
+            .collect::<Vec<_>>()
+    } else {
+        // Fallback to default dotenv
+        dotenvy::dotenv()
+            .map_or_else(|e| if e.not_found() { Ok(()) } else { Err(e) }, |_| Ok(()))
```
loewenheim, line 124: Tbh I think this `map_or_else` is really hard to understand, but it was already that way before. This isn't a blocker IMO.

loewenheim (approved):

Kinqdos:
@szokeasaurusrex Is this ready to merge now?

PR #2480, opened 2025-04-29
WIP: Upload sourcemaps to multiple projects

### LINKED ISSUE #2408: Support uploading sourcemaps to multiple projects
Currently, it is only possible to upload sourcemaps to a single project at a time. However, the Sentry HTTP API which we use to upload sourcemaps supports associating the uploaded sourcemaps with multiple projects, and so, we would like to enable this functionality in Sentry CLI.

### Solution brainstorm

It is already possible to provide multiple projects to a Sentry CLI command by passing the `-p` or `--project` argument multiple times, like so:

```sh
sentry-cli sourcemaps upload -p proj1 -p proj2 ...
```

However, the `sentry-cli sourcemaps upload` command only takes the first project provided (see `Config::get_org_and_project` and `Config::get_project` methods), the rest of the projects passed are ignored.

Instead, let's use `Config::get_projects` (note: plural) to get all of the projects passed. Then, modify the `UploadContext` struct to store multiple projects, instead of just a single project, and pass all of these projects in the HTTP request we make to Sentry.

a-hariti:
Closes #2408

### DIFF
--- a/src/commands/files/upload.rs
+++ b/src/commands/files/upload.rs
@@ -124,6 +124,7 @@ pub fn make_command(command: Command) -> Command {
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
+    println!("✅ executing ...");
     let config = Config::current();
     let release = config.get_release_with_legacy_fallback(matches)?;
     let org = config.get_org(matches)?;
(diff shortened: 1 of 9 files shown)


src/commands/files/upload.rs
```diff
@@ -124,6 +124,7 @@ pub fn make_command(command: Command) -> Command {
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
+    println!("✅ executing ...");
```
a-hariti, line 127: 💀

Issue #2036, opened 2024-04-23
Improved issue creation flow


szokeasaurusrex:
We should improve the issue creation flow in this repository to ensure that people who open issues provide all the information we need to triage and investigate their problems upfront.

These improvements can be implemented by switching from using issue templates to issue forms, and by hiding the "blank issue" link. The specific changes are broken down below:

### Tasks

- [ ] [https://github.com/getsentry/sentry-cli/issues/2037](https://github.com/getsentry/sentry-cli/issues/2037)
- [ ] [https://github.com/getsentry/sentry-cli/issues/2038](https://github.com/getsentry/sentry-cli/issues/2038)
- [ ] Hide "blank issue" link

PR #2493, opened 2025-05-07
chore: update issue templates

stephanie-anderson:
closes https://github.com/getsentry/sentry-cli/issues/2038 (added the same form as for other repos)
closes https://github.com/getsentry/sentry-cli/issues/2036


.github/ISSUE_TEMPLATE/maintainer-blank.yml
```diff
@@ -0,0 +1,11 @@
+name: Blank Issue
```
szokeasaurusrex, line 1: I'm not sure we need this, given that maintainers can raise blank issues via Linear now

.github/ISSUE_TEMPLATE/01-feature.yml
```diff
@@ -0,0 +1,29 @@
...
+description: Suggest an idea for improving Sentry CLI
+labels: ["CLI", "Feature"]
+body:
+  - type: textarea
+    id: problem
+    attributes:
+      label: Problem Statement
+      description: What problem could CLI solve that it currently doesn't?
+      placeholder: |-
+        I want to make whirled peas, but Sentry doesn't blend.
```
szokeasaurusrex, line 11: I think this "whirled peas" thing is a bit weird, but I guess we use it in our other repos so it's okay

szokeasaurusrex (approved):

.github/ISSUE_TEMPLATE/01-feature.yml
stephanie-anderson: yeah it's exactly the same, if we change it at one point, we can change it everywhere simultaneously

.github/ISSUE_TEMPLATE/maintainer-blank.yml
stephanie-anderson: Screenshot 2025-05-07 at 14 36 14


This way, it shows up like this for users

szokeasaurusrex: yeah, that's what I am saying though, is if the blank issue template is anyways reserved for maintainers, and maintainers can make the issues through Linear, we don't have to expose the blank issue template at all in GitHub.

But on second thought, it probably doesn't hurt to keep it in GitHub and might even be useful sometimes (e.g. because you cannot link other GitHub things on Linear as easily as when creating the issue on GitHub)

So let's keep it!

PR #2512, opened 2025-05-20
feat: Mark all `files` subcommands as deprecated.

Swatinem:
The commands are now hidden from help output (making them less likely to be discovered and used), and they are marked with a deprecation notice.

When invoking the commands, a deprecation notice will be printed as well.

Also the previously deprecated `sourcemaps explain` command (which was replaced by an in-app sourcemaps debugger)  which already prints a deprecation notice is also being marked as hidden to be less discoverable.

Ref #2513

### DIFF
--- a/src/commands/files/mod.rs
+++ b/src/commands/files/mod.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use clap::{ArgMatches, Command};
+use console::style;
 
 use crate::utils::args::ArgExt;
 
@@ -25,7 +26,8 @@ pub fn make_command(mut command: Command) -> Command {
     }
 
     command = command
-        .about("Manage release artifacts.")
+        .about("[DEPRECATED] Manage release artifacts.")
+        .hide(true)
         .subcommand_required(true)
         .arg_required_else_help(true)
         .org_arg()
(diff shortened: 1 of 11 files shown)


src/commands/files/mod.rs
```diff
@@ -42,6 +44,9 @@ pub fn make_command(mut command: Command) -> Command {
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
+    println!("{}", style("⚠ DEPRECATION NOTICE: This functionality will be removed in a future version of `sentry-cli`. \
+        Use the `sourcemaps` command instead.").yellow());
```
szokeasaurusrex, line 48: We should probably use `eprintln!` for this:

```suggestion
    eprintln!("{}", style("⚠ DEPRECATION NOTICE: This functionality will be removed in a future version of `sentry-cli`. \
        Use the `sourcemaps` command instead.").yellow());
```

We will also need to update [this docs section](https://docs.sentry.io/cli/releases/#managing-release-artifacts) (including subsections) to reflect this change; that page makes it seem like the `sourcemaps` commands are a "special case" of the `files` commands

tests/integration/_cases/help/help.trycmd
szokeasaurusrex, line 1: you also need to update the corresponding Windows test, so that the tests pass

szokeasaurusrex (commented):
two small comments, otherwise looks good! I'm gonna open an issue to fully remove these commands with the next major release

src/commands/files/mod.rs
Swatinem: Done.
Thanks for the reminder: https://github.com/getsentry/sentry-docs/pull/13767

PR #2525, opened 2025-05-27
Setup initial mobile-app command scaffold

rbro112:
In preparation for adding mobile application upload support to Sentry, this creates the basic scaffolding of a new `mobile-app` command with a single `upload` subcommand. The implementation of this subcommand will come in subsequent PRs.

Currently, the `mobile-app` command is set to hidden to ensure minimal exposure if a release is cut.

### DIFF
--- a/src/commands/mobile_app/mod.rs
+++ b/src/commands/mobile_app/mod.rs
@@ -21,12 +21,11 @@ pub fn make_command(mut command: Command) -> Command {
     }
 
     command = command
-        .about("(Experimental) Manage mobile apps.")
+        .about("[EXPERIMENTAL] Manage mobile apps.")
         .subcommand_required(true)
         .arg_required_else_help(true)
         .org_arg()
         .project_arg(true)
-        .release_arg()
         // TODO: Remove this when ready for release
         .hide(true);
     each_subcommand!(add_subcommand);
--- a/src/commands/mobile_app/mod.rs
+++ b/src/commands/mobile_app/mod.rs
@@ -1,0 +1,48 @@
+use anyhow::Result;
+use clap::{ArgMatches, Command};
+
+use crate::utils::args::ArgExt;
+
+pub mod upload;
+
+macro_rules! each_subcommand {
+    ($mac:ident) => {
+        $mac!(upload);
+    };
+}
+
+pub fn make_command(mut command: Command) -> Command {
+    macro_rules! add_subcommand {
+        ($name:ident) => {{
+            command = command.subcommand(crate::commands::mobile_app::$name::make_command(
+                Command::new(stringify!($name).replace('_', "-")),
+            ));
+        }};
+    }
+
+    command = command
+        .about("(Experimental) Manage mobile apps.")
+        .subcommand_required(true)
+        .arg_required_else_help(true)
+        .org_arg()
+        .project_arg(true)
+        .release_arg()
+        // TODO: Remove this when ready for release
+        .hide(true);
+    each_subcommand!(add_subcommand);
+    command
+}
+
+pub fn execute(matches: &ArgMatches) -> Result<()> {
+    macro_rules! execute_subcommand {
+        ($name:ident) => {{
+            if let Some(sub_matches) =
+                matches.subcommand_matches(&stringify!($name).replace('_', "-"))
+            {
+                return crate::commands::mobile_app::$name::execute(&sub_matches);
+            }
+        }};
+    }
+    each_subcommand!(execute_subcommand);
+    unreachable!();
+}
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -18,6 +18,7 @@ pub fn make_command(command: Command) -> Command {
         )
 }
 
+#[allow(clippy::unnecessary_wraps)]
 pub fn execute(_matches: &ArgMatches) -> Result<()> {
     println!("Uploading mobile app files to a project is not yet implemented.");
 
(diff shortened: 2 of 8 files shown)


src/commands/mobile_app/upload.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub fn make_command(command: Command) -> Command {
+    command
+        .about("Upload mobile app files to a project.")
+        .org_arg()
+        .project_arg(false)
+        .arg(
+            Arg::new("paths")
+                .value_name("PATH")
+                .help("The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive and XCFramework.")
```
trevor-e, line 15: I believe for iOS it's only XCArchive for now.

trevor-e (approved):

src/commands/mobile_app/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        }};
+    }
+
+    command = command
+        .about("(Experimental) Manage mobile apps.")
+        .subcommand_required(true)
+        .arg_required_else_help(true)
+        .org_arg()
+        .project_arg(true)
+        .release_arg()
```
szokeasaurusrex, line 29: [question] is setting all of these going to be relevant for mobile app uploads?

src/commands/mobile_app/upload.rs
```diff
@@ -0,0 +1,26 @@
...
+                .help("The path to the mobile app files to upload. Supported files include Apk, Aab or XCArchive.")
+                .num_args(1..)
+                .action(ArgAction::Append),
+        )
+}
+
+#[allow(clippy::unnecessary_wraps)]
+pub fn execute(_matches: &ArgMatches) -> Result<()> {
+    println!("Uploading mobile app files to a project is not yet implemented.");
+
```
szokeasaurusrex, line 24: [nit] I would use `eprintln!` here so that this message goes to `stderr` rather than `stdout`.

```suggestion
    eprintln!("Uploading mobile app files to a project is not yet implemented.");
```

src/commands/mobile_app/upload.rs
```diff
@@ -0,0 +1,26 @@
...
+        .arg(
+            Arg::new("paths")
+                .value_name("PATH")
+                .help("The path to the mobile app files to upload. Supported files include Apk, Aab or XCArchive.")
+                .num_args(1..)
+                .action(ArgAction::Append),
+        )
+}
+
+#[allow(clippy::unnecessary_wraps)]
```
szokeasaurusrex, line 21: Please use `expect` rather than `allow`.

The difference is that the linter will complain if you use an `expect` where it is not needed, whereas there is nothing to indicate if you have an unnecessary `allow`. So, once this function starts returning an error `clippy` will remind you to remove this if you use `expect`.

```suggestion
#[expect(clippy::unnecessary_wraps)]
```

src/commands/mobile_app/upload.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
+use clap::ArgAction;
+use clap::{Arg, ArgMatches, Command};
+
+use crate::utils::args::ArgExt;
+
+pub fn make_command(command: Command) -> Command {
+    command
+        .about("Upload mobile app files to a project.")
```
szokeasaurusrex, line 9: Please also mark this as experimental; the parent command's about text, which already includes an experimental tag, does not show up when you run `sentry-cli mobile-app upload --help`, so we should include it here as well.

```suggestion
        .about("[EXPERIMENTAL] Upload mobile app files to a project.")
```

src/commands/mobile_app/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    macro_rules! add_subcommand {
+        ($name:ident) => {{
+            command = command.subcommand(crate::commands::mobile_app::$name::make_command(
+                Command::new(stringify!($name).replace('_', "-")),
+            ));
+        }};
+    }
+
+    command = command
+        .about("(Experimental) Manage mobile apps.")
```
szokeasaurusrex, line 24: [nit] this is the style we typically use for marking commands. We don't have any other experimental commands at the moment, but this style aligns with how we mark commands as deprecated using `[DEPRECATED]`

```suggestion
        .about("[EXPERIMENTAL] Manage mobile apps.")
```

szokeasaurusrex (approved):
Hey, looks really good! I have left a couple of comments I'd like you to address before merging (including some nitpicks which you can feel free to ignore).

I would also recommend adding some basic tests to check that the command exists and also to assert the `--help` texts of both the parent command and the sub-command. For a good simple example of this, check out the tests for the `login` command. [tests/integration/login.rs](https://github.com/getsentry/sentry-cli/blob/b9d50ff102e4b081bc0e9c0c57fcf04b63dcae1c/tests/integration/login.rs) and [tests/integration/_cases/login](https://github.com/getsentry/sentry-cli/blob/8be89c62f5f683f4417a57a5a3b8bed7b31ffc6e/tests/integration/_cases/login) are where the tests are defined, and we use the [trycmd](https://docs.rs/trycmd/latest/trycmd/) crate for these kinds of tests.

You can add the tests in this PR or in a separate PR

src/commands/mobile_app/mod.rs
rbro112: Project and org will be used. Release is TBD, but tentatively yes as that's the first spot we'll be fitting in. I'll go ahead and remove release and add it later should we need it.

src/commands/mobile_app/upload.rs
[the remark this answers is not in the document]
rbro112: It is not. Generally, uploading an XCArchive won't be possible without being on macOS from my understanding (as the file won't be available on other OSes), but IMO there's no need to lock this down to have an OS check.

PR #2531, opened 2025-05-28
Add basic file validation to mobile-app upload command

rbro112:
Adds basic validation to ensure the paths provided by the `[PATHS]` argument are an APK, AAB (zip) or XCArchive directory.

### DIFF
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -1,8 +1,14 @@
+use std::fs::File;
+use std::io::{BufReader, Seek, SeekFrom};
+use std::path::Path;
+
+use anyhow::anyhow;
 use anyhow::Result;
 use clap::ArgAction;
 use clap::{Arg, ArgMatches, Command};
 
 use crate::utils::args::ArgExt;
+use crate::utils::mobile_app::{is_aab_file, is_apk_file, is_xcarchive_directory, is_zip_file};
 
 pub fn make_command(command: Command) -> Command {
     command
@@ -19,8 +25,60 @@ pub fn make_command(command: Command) -> Command {
 }
 
 #[expect(clippy::unnecessary_wraps)]
-pub fn execute(_matches: &ArgMatches) -> Result<()> {
-    eprintln!("Uploading mobile app files to a project is not yet implemented.");
+pub fn execute(matches: &ArgMatches) -> Result<()> {
+    let path_strings = matches.get_many::<String>("paths").unwrap();
 
+    let mut paths: Vec<&Path> = Vec::new();
+    for path_string in path_strings {
+        let path: &Path = path_string.as_ref();
+
+        if !path.exists() {
+            return Err(anyhow!("File does not exist: {}", path.display()));
+        }
+
+        if !path.is_file() {
+            return Err(anyhow!("Path is not a file: {}", path.display()));
+        }
+
+        validate_mobile_app_file(path)?;
+        paths.push(path);
+    }
+
+    for path in paths {
+        println!("Uploading mobile app file: {}", path.display());
+        // TODO: Upload the file to the chunked uploads API
+    }
+
+    eprintln!("Uploading mobile app files to a project is not yet implemented.");
     Ok(())
+}
+
+/// Validate the file is a valid mobile app file
+pub fn validate_mobile_app_file(path: &Path) -> Result<()> {
+    let file = File::open(path)?;
+    let mut reader = BufReader::new(file);
+
+    // Check for APK/AAB (both are ZIP files with specific structure)
+    if is_zip_file(&mut reader)? {
+        reader.seek(SeekFrom::Start(0))?;
+
+        if is_aab_file(&mut reader)? {
+            return Ok(());
+        }
+
+        reader.seek(SeekFrom::Start(0))?;
+        if is_apk_file(&mut reader)? {
+            return Ok(());
+        }
+    }
+
+    // Check for XCArchive (directory structure)
+    if path.is_dir() && is_xcarchive_directory(path)? {
+        return Ok(());
+    }
+
+    Err(anyhow!(
+        "File is not a recognized mobile app format (APK, AAB, or XCArchive): {}",
+        path.display()
+    ))
 }
--- a/src/utils/mobile_app/validation.rs
+++ b/src/utils/mobile_app/validation.rs
@@ -1,0 +1,57 @@
+use std::io::{Read, Seek, SeekFrom};
+use std::path::Path;
+
+use anyhow::Result;
+
+pub fn is_zip_file<R: Read + Seek>(reader: &mut R) -> Result<bool> {
+    let mut magic = [0u8; 4];
+    reader.seek(SeekFrom::Start(0))?;
+
+    if reader.read_exact(&mut magic).is_err() {
+        return Ok(false);
+    }
+
+    // https://en.wikipedia.org/wiki/List_of_file_signatures
+    const ZIP_MAGIC: [u8; 4] = [0x50, 0x4B, 0x03, 0x04];
+    const ZIP_MAGIC_EMPTY: [u8; 4] = [0x50, 0x4B, 0x05, 0x06];
+    const ZIP_MAGIC_SPANNED: [u8; 4] = [0x50, 0x4B, 0x07, 0x08];
+    Ok(magic == ZIP_MAGIC || magic == ZIP_MAGIC_EMPTY || magic == ZIP_MAGIC_SPANNED)
+}
+
+pub fn is_apk_file<R: Read + Seek>(reader: &mut R) -> Result<bool> {
+    reader.seek(SeekFrom::Start(0))?;
+
+    let mut archive = zip::ZipArchive::new(reader)?;
+
+    // APK files must contain AndroidManifest.xml at the root of the zip file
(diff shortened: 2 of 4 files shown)


src/commands/mobile_app/upload.rs
```diff
@@ -18,9 +24,59 @@ pub fn make_command(command: Command) -> Command {
...
+
+    let mut paths: Vec<&Path> = Vec::new();
+    for path_string in path_strings {
+        let path: &Path = path_string.as_ref();
+
+        if !path.exists() {
+            return Err(anyhow!("File does not exist: {}", path.display()));
+        }
 
+        if !path.is_file() {
```
trevor-e, line 38: Technically `.xcarchive` is a directory, not sure if it would fail this check?

src/commands/mobile_app/upload.rs
```diff
@@ -18,9 +24,59 @@ pub fn make_command(command: Command) -> Command {
...
+        }
+
+        reader.seek(SeekFrom::Start(0))?;
+        if is_apk_file(&mut reader)? {
+            return Ok(());
+        }
+    }
+
+    // Check for XCArchive (directory)
+    if path.is_dir() && is_xcarchive_directory(path)? {
```
trevor-e, line 74: Since down here you check if it's a dir.

trevor-e (approved):

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
+use std::io::{Read, Seek, SeekFrom};
+use std::path::Path;
+
+use anyhow::Result;
+
+pub fn is_zip_file<R: Read + Seek>(reader: &mut R) -> Result<bool> {
```
szokeasaurusrex: [nit] I find the following syntax more readable, especially for funcitons which have mutiple generics and/or trait bounds

```suggestion
pub fn is_zip_file<R>(reader: &mut R) -> Result<bool>
where
    R: Read + Seek,
{
```

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
...
+    let mut magic = [0u8; 4];
+    reader.seek(SeekFrom::Start(0))?;
```
szokeasaurusrex: It looks like you are already performing the seeks outside the function calls, so I would remove this. You should also remove the `R: Seek` trait bound from the function.
```suggestion
```

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
...
+    // https://en.wikipedia.org/wiki/List_of_file_signatures
+    const ZIP_MAGIC: [u8; 4] = [0x50, 0x4B, 0x03, 0x04];
+    const ZIP_MAGIC_EMPTY: [u8; 4] = [0x50, 0x4B, 0x05, 0x06];
+    const ZIP_MAGIC_SPANNED: [u8; 4] = [0x50, 0x4B, 0x07, 0x08];
+    Ok(magic == ZIP_MAGIC || magic == ZIP_MAGIC_EMPTY || magic == ZIP_MAGIC_SPANNED)
+}
+
+pub fn is_apk_file<R: Read + Seek>(reader: &mut R) -> Result<bool> {
+    reader.seek(SeekFrom::Start(0))?;
+
```
szokeasaurusrex: Same suggestions here

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
...
+    // APK files must contain AndroidManifest.xml at the root of the zip file
+    let has_manifest = archive.by_name("AndroidManifest.xml").is_ok();
+
+    Ok(has_manifest)
+}
+
+pub fn is_aab_file<R: Read + Seek>(reader: &mut R) -> Result<bool> {
+    reader.seek(SeekFrom::Start(0))?;
+
```
szokeasaurusrex: Same suggestions here

src/commands/mobile_app/upload.rs
```diff
@@ -18,9 +24,56 @@ pub fn make_command(command: Command) -> Command {
         )
 }
 
-#[expect(clippy::unnecessary_wraps)]
-pub fn execute(_matches: &ArgMatches) -> Result<()> {
-    eprintln!("Uploading mobile app files to a project is not yet implemented.");
+pub fn execute(matches: &ArgMatches) -> Result<()> {
+    let path_strings = matches.get_many::<String>("paths").unwrap();
```
szokeasaurusrex: Please use `expect` rather than `unwrap` in any new code. `expect` is the pretty much the same as `unwrap`; the only difference is that it takes a string parameter containing a message, which is shown if the `expect` is triggered. You can use this string to document the precondition you expect is satisfied.

I realize we use `unwrap` in many places in the CLI codebase at the moment, but this is something I would like to move away from, as using `expect` forces us to document our assumptions.


```suggestion
    let path_strings = matches.get_many::<String>("paths").expect("clap enforces this as a required argument");
```

src/commands/mobile_app/upload.rs
```diff
@@ -18,9 +24,56 @@ pub fn make_command(command: Command) -> Command {
...
+
+    let mut paths: Vec<&Path> = Vec::new();
```
szokeasaurusrex: The compiler can infer this type

```suggestion
    let mut paths = Vec::new();
```

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
...
+    let mut archive = zip::ZipArchive::new(reader)?;
+
+    // AAB files must contain BundleConfig.pb and base/manifest/AndroidManifest.xml
+    let has_bundle_config = archive.by_name("BundleConfig.pb").is_ok();
+    let has_base_manifest = archive.by_name("base/manifest/AndroidManifest.xml").is_ok();
+
+    Ok(has_bundle_config && has_base_manifest)
+}
+
+pub fn is_xcarchive_directory<P: AsRef<Path>>(path: P) -> Result<bool> {
```
szokeasaurusrex: Also here


```suggestion
pub fn is_xcarchive_directory<P>(path: P) -> Result<bool> 
where
    P: AsRef<Path>
{
```

src/commands/mobile_app/upload.rs
```diff
@@ -18,9 +24,56 @@ pub fn make_command(command: Command) -> Command {
...
+    }
+
+    eprintln!("Uploading mobile app files to a project is not yet implemented.");
     Ok(())
 }
+
+fn validate_is_mobile_app(path: &Path) -> Result<()> {
+    let file = File::open(path)?;
+    let mut reader = BufReader::new(file);
+
```
szokeasaurusrex: [suggestion] Elsewhere in the codebase, we typically use [Symbolic's `ByteView` struct](https://docs.rs/symbolic/latest/symbolic/common/struct.ByteView.html) rather than using readers and seeking.

[`ByteView::open`](https://docs.rs/symbolic/latest/symbolic/common/struct.ByteView.html#method.open) memory-maps the file, and allows you to dereference it as a `[u8]` (byte slice) as if the whole file was in memory, giving you random access to all of the file. 

Although I won't insist on it, I would highly recommend switching to using `ByteView`. You would need to modify the functions you are calling here to take a `&[u8]` instead of a generic `Read + Seek` as they are now.

rbro112: Will do, appreciate the suggestion!

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
...
+    let path = path.as_ref();
+
+    // XCArchive should have Info.plist and a .app file in Products/Applications/
+    let info_plist = path.join("Info.plist");
+    let applications_dir = path.join("Products").join("Applications");
```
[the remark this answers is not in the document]
rbro112: Missed this, added in next PR (#2532)

src/utils/mobile_app/validation.rs
```diff
@@ -0,0 +1,63 @@
+use std::path::Path;
+use symbolic::common::ByteView;
+
+use anyhow::Result;
+
+pub fn is_zip_file(byteview: &ByteView) -> bool {
```
szokeasaurusrex, line 6: Since `ByteView` dereferences into a `&[u8]`, it would be preferable to take a `&[u8]` rather than a `&ByteView` for the functions in this file. That makes the API more flexible, since you could also just pass a generic byte slice, even something which is not a `ByteView` already. 

I'll make a separate PR for this change

tests/integration/mobile_app/upload.rs
```diff
@@ -1,6 +1,39 @@
...
+#[test]
+fn command_mobile_app_upload_invalid_xcarchive() {
+    TestManager::new()
+        .assert_cmd(vec![
+            "mobile-app",
+            "upload",
+            "tests/integration/_fixtures/mobile_app/invalid_xcarchive",
+        ])
+        .run_and_assert(AssertCommand::Failure);
+}
```
szokeasaurusrex, line 39: @rbro112 Nice! Are you also planning to test the successful case, or is this waiting for additional logic to be added?

src/utils/mobile_app/validation.rs
szokeasaurusrex: #2541

rbro112: Thanks for the fix!

PR #2532, opened 2025-05-29
Add Zip normalization of provided paths before uploading

rbro112:
For ease of parsing artifacts on the backend, we'll "normalize" all provided paths to a basic zip file. This ensures that no matter the artifact uploaded (files: APK, AAB, directory: XCArchive), we can process all as a zip file on the backend. This was an optimization done with Emerge that greatly simplified backend artifact handling.

Skipping tests here to be added at a later time when I leverage a full integration test to validate the chunked uploads match the expected zip format.

### DIFF
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -1,11 +1,8 @@
 use std::io::Write;
 use std::path::Path;
 
-use anyhow::anyhow;
-use anyhow::Context;
-use anyhow::Result;
-use clap::ArgAction;
-use clap::{Arg, ArgMatches, Command};
+use anyhow::{anyhow, Context as _, Result};
+use clap::{Arg, ArgAction, ArgMatches, Command};
 use symbolic::common::ByteView;
 use zip::write::SimpleFileOptions;
 use zip::ZipWriter;
@@ -46,15 +43,25 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         validate_is_mobile_app(path, &byteview)?;
 
         let normalized_zip = if path.is_file() {
-            normalize_file(path, &byteview)
+            normalize_file(path, &byteview).with_context(|| {
+                format!(
+                    "Failed to generate uploadable bundle for file {}",
+                    path.display()
+                )
+            })?
         } else if path.is_dir() {
-            normalize_directory(path)
+            normalize_directory(path).with_context(|| {
+                format!(
+                    "Failed to generate uploadable bundle for directory {}",
+                    path.display()
+                )
+            })?
         } else {
             Err(anyhow!(
                 "Path {} is neither a file nor a directory, cannot upload",
                 path.display()
-            ))
-        }?;
+            ))?
+        };
 
         normalized_zips.push(normalized_zip);
     }
@@ -121,9 +128,7 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
     {
         let entry_path = entry.path();
         if entry_path.is_file() {
-            let relative_path = entry_path.strip_prefix(path).with_context(|| {
-                format!("Failed to get relative path for {}", entry_path.display())
-            })?;
+            let relative_path = entry_path.strip_prefix(path)?;
 
             zip.start_file(
                 relative_path.to_string_lossy(),
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -30,7 +30,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .get_many::<String>("paths")
         .expect("paths argument is required");
 
-    let mut normalized_zips = Vec::new();
+    let mut normalized_zips = vec![];
     for path_string in path_strings {
         let path: &Path = path_string.as_ref();
 
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -32,7 +32,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .expect("paths argument is required");
 
     let mut paths = Vec::new();
-    let mut normalized_zips: Vec<TempFile> = Vec::new();
+    let mut normalized_zips = Vec::new();
     for path_string in path_strings {
         let path: &Path = path_string.as_ref();
 
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -35,7 +35,6 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let mut normalized_zips = Vec::new();
     for path_string in path_strings {
-        println!("Processing path: {}", path_string);
         let path: &Path = path_string.as_ref();
 
         if !path.exists() {
@@ -94,7 +93,6 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
 
 // For APK and AAB files, we'll copy them directly into the zip
 fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
-    println!("Normalizing file: {}", path.display());
     let temp_file = TempFile::create()?;
     let mut zip = ZipWriter::new(temp_file.open()?);
 
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -105,7 +105,7 @@ fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
         .with_context(|| format!("Failed to get relative path for {}", path.display()))?;
 
     zip.start_file(file_name, SimpleFileOptions::default())?;
-    zip.write(bytes)?;
+    zip.write_all(bytes)?;
 
     zip.finish()?;
     Ok(temp_file)
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -1,3 +1,4 @@
+use std::io::Write;
 use std::path::Path;
 
 use anyhow::anyhow;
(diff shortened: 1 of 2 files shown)


src/commands/mobile_app/upload.rs
```diff
@@ -77,3 +80,45 @@ fn validate_is_mobile_app(path: &Path) -> Result<()> {
         path.display()
     ))
 }
+
+/// Normalizes a mobile app file into a zip file to ensure consistent artifact parsing logic on the backend.
+fn normalize_mobile_app(path: &Path) -> Result<TempFile> {
+    let temp_file = TempFile::create()?;
+    let mut zip = ZipWriter::new(temp_file.open()?);
+
+    if path.is_file() {
```
trevor-e, line 89: Should this handle the case where the user already zipped the file? Seems like this might re-zip it.

src/commands/mobile_app/upload.rs
```diff
@@ -77,3 +80,45 @@ fn validate_is_mobile_app(path: &Path) -> Result<()> {
...
+                    relative_path.to_string_lossy(),
+                    SimpleFileOptions::default(),
+                )?;
+                std::io::copy(&mut file, &mut zip)?;
+            }
+        }
+    }
+
+    zip.finish()?;
+    Ok(temp_file)
```
trevor-e, line 123: I'm new to Rust error handling but does this function handle errors properly? E.g. if opening the file or building the zip fails in some way, I thought this function would return an error result type.

rbro112: It does, the `?` at the end of `finish` is rust's error handling propogation: https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator

trevor-e: Wow that is extremely subtle lol

src/commands/mobile_app/upload.rs
```diff
@@ -27,7 +30,7 @@ pub fn make_command(command: Command) -> Command {
 pub fn execute(matches: &ArgMatches) -> Result<()> {
     let path_strings = matches.get_many::<String>("paths").unwrap();
 
-    let mut paths: Vec<&Path> = Vec::new();
+    let mut normalized_zips: Vec<TempFile> = Vec::new();
     for path_string in path_strings {
```
szokeasaurusrex, line 34: Type can be inferred here

```suggestion
    let mut normalized_zips = Vec::new();
```

szokeasaurusrex, line 34: Type can be inferred here


```suggestion
    let mut normalized_zips = Vec::new();
```

src/commands/mobile_app/upload.rs
szokeasaurusrex: Technically we could encounter this case if the contents located at the path are modified by another process while the CLI is executing. However, my guess is it likely does not make too much sense to handle this case

src/commands/mobile_app/upload.rs
```diff
@@ -77,3 +80,45 @@ fn validate_is_mobile_app(path: &Path) -> Result<()> {
         path.display()
     ))
 }
+
+/// Normalizes a mobile app file into a zip file to ensure consistent artifact parsing logic on the backend.
+fn normalize_mobile_app(path: &Path) -> Result<TempFile> {
```
szokeasaurusrex: If you go with my suggestion of using `ByteView` from #2531, I would recommend replacing the `path` parameter with the `&[u8]` file contents slice (or, if the path is still needed to name the zip file, I would also pass the `&[u8]` and use it as the source for the copy operation)

src/commands/mobile_app/upload.rs
```diff
@@ -77,3 +80,45 @@ fn validate_is_mobile_app(path: &Path) -> Result<()> {
...
+    let temp_file = TempFile::create()?;
```
szokeasaurusrex, line 121: Something I would generally like to improve in Sentry CLI (especially with any new code we add) is to make error messages clearer and more actionable to users. 

Many error messages in Sentry CLI end up containing very little information; this makes them confusing to users and difficult for us to debug. 

The `anyhow` crate allows us to add context to any error which is then displayed as the error message, alongside the original error message. You can add context to a result using `.context` (eagerly evaluated) or `.with_context` (lazily evaluated). To use this feature, you need to add an import of `anyhow::Context as _`.

Please add context, such as the following, to any errors where the default error message contains too little information to be helpful to users (this is often the case for errors arising from libraries or general CLI utility functions).

```suggestion
    let temp_file = TempFile::create()
        .with_context(|| format!("Failed to create temp file while processing {}", path.display()))?;
```

szokeasaurusrex (commented):
Please see my comment about error messages; I would prefer we improve these before merging this PR

src/commands/mobile_app/upload.rs
rbro112: Updated! Let me know if you'd like me to update any of the new error messages and the propagation I have in place.

szokeasaurusrex: I don't think I am seeing the changed error messages; this tempfile::create still does not have a `with_context` on it. Did you perhaps forget to commit something?

szokeasaurusrex: However, I am realizing now, since there are many places in this function where an error could get raised, it might make more sense just to place the context where this function gets called. I'll add the suggestion above

src/commands/mobile_app/upload.rs
```diff
@@ -28,43 +33,54 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        let normalized_zip = if path.is_file() {
+            normalize_file(path, &byteview)
+        } else if path.is_dir() {
+            normalize_directory(path)
+        } else {
+            Err(anyhow!(
+                "Path {} is neither a file nor a directory, cannot upload",
+                path.display()
+            ))
+        }?;
```
szokeasaurusrex: [nit] while having the question mark operator here is technically correct, I think the code would be more readable if you place it on each line that can cause an error

src/commands/mobile_app/upload.rs
```diff
@@ -28,43 +33,54 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-        paths.push(path);
+        let byteview = ByteView::open(path)?;
+
+        validate_is_mobile_app(path, &byteview)?;
+
+        let normalized_zip = if path.is_file() {
+            normalize_file(path, &byteview)
+        } else if path.is_dir() {
+            normalize_directory(path)
+        } else {
```
szokeasaurusrex: As I mention in a comment below, it might make sense to add context to any errors coming from these functions

```suggestion
            normalize_file(path, &byteview).with_context(|| format!("Failed to normalize file {}", path.display()))
        } else if path.is_dir() {
            normalize_directory(path).with_context(|| format!("Failed to normalize directory {}", path.display()))
        } else {
```

src/commands/mobile_app/upload.rs
```diff
@@ -1,12 +1,17 @@
+use std::io::Write;
 use std::path::Path;
 
 use anyhow::anyhow;
+use anyhow::Context;
 use anyhow::Result;
```
szokeasaurusrex: [nit] Now that we have multiple imports from this same module, I would suggest grouping them like so:

```suggestion
use anyhow::{anyhow, Context, Result};
```

As an FYI, since `rustfmt` does not automatically group imports in any particular way, I have lately been following [this recommendation](https://canonical.github.io/rust-best-practices/import-discipline.html#import-form), which I learned of at a Rust conference (emphasis added):

> Excessive repetition harms readability by adding lots of visual noise. For this reason, avoid Java-style imports where every single imported item gets its own line with a complete and exhaustive path and instead use the cleaner nested syntax. **If a path contains n parts, merge the first n-1, so that only the final part is grouped with its siblings.**

szokeasaurusrex: [nit/rust fyi] As another aside, since we don't actually use `Context` directly, only the trait implementaions, we can import it as `_` so that the `Context` name is inaccessible in this module:

```suggestion
use use anyhow::{anyhow, Context as _, Result}; 
```

Honestly I am unsure which way is "better," but I also just recently learned about this and thought I'd share

src/commands/mobile_app/upload.rs
```diff
@@ -28,43 +33,54 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .get_many::<String>("paths")
         .expect("paths argument is required");
 
-    let mut paths = Vec::new();
+    let mut normalized_zips = Vec::new();
```
szokeasaurusrex: [rust fyi]
You can also use the `vec!` macro here; feel free to use whichever you'd prefer, just thought I would mention this in case you were unaware.

```suggestion
    let mut normalized_zips = vec![];
```

szokeasaurusrex (approved):
Looks good once these comments are addressed

src/commands/mobile_app/upload.rs
rbro112: Yeah that was my thought process, the caller of these would handle the error vs the specific errors within the function itself rather than handle each individual error. Most of these errors are related to internal implementation details and while I agree with having user-friendly messages, I think it would be more user-friendly to add debug logging and not expose the full details to the user of each individual error.

src/commands/mobile_app/upload.rs
rbro112: Done! As for the other errors, I added with with_context handling to the caller of the normalize functions as per your similar sentiment, there's quite a few cases that can throw errors so wrapping each one is a bit cumbersome and not necessarily more user-friendly IMO

src/commands/mobile_app/upload.rs
rbro112: Done!

src/commands/mobile_app/upload.rs
rbro112: Nice, thanks for sharing these - still learning about all the syntactical sugar rust offers so keep these coming 🙏

PR #2542, opened 2025-06-05
WIP ref(mobile-app): Wireup mobile app upload

rbro112:
WIP, full description to come

### DIFF
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -32,7 +32,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .expect("paths argument is required");
 
     let mut paths = Vec::new();
-    let mut normalized_zips: Vec<TempFile> = Vec::new();
+    let mut normalized_zips = Vec::new();
     for path_string in path_strings {
         let path: &Path = path_string.as_ref();
 
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -35,7 +35,6 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let mut normalized_zips = Vec::new();
     for path_string in path_strings {
-        println!("Processing path: {}", path_string);
         let path: &Path = path_string.as_ref();
 
         if !path.exists() {
@@ -94,7 +93,6 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
 
 // For APK and AAB files, we'll copy them directly into the zip
 fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
-    println!("Normalizing file: {}", path.display());
     let temp_file = TempFile::create()?;
     let mut zip = ZipWriter::new(temp_file.open()?);
 
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -105,7 +105,7 @@ fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
         .with_context(|| format!("Failed to get relative path for {}", path.display()))?;
 
     zip.start_file(file_name, SimpleFileOptions::default())?;
-    zip.write(bytes)?;
+    zip.write_all(bytes)?;
 
     zip.finish()?;
     Ok(temp_file)
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -1,3 +1,4 @@
+use std::io::Write;
 use std::path::Path;
 
 use anyhow::anyhow;
@@ -31,7 +32,6 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .get_many::<String>("paths")
         .expect("paths argument is required");
 
-    let mut paths = Vec::new();
     let mut normalized_zips = Vec::new();
     for path_string in path_strings {
         let path: &Path = path_string.as_ref();
@@ -40,8 +40,28 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             return Err(anyhow!("Path does not exist: {}", path.display()));
         }
 
-        validate_is_mobile_app(path)?;
-        let normalized_zip = normalize_mobile_app(path)?;
+        let byteview = ByteView::open(path)?;
+
+        validate_is_mobile_app(path, &byteview)?;
+
+        let normalized_zip = if path.is_file() {
+            normalize_file(path, &byteview)
+        } else if path.is_dir() {
+            normalize_directory(path)
+        } else {
+            Err(anyhow!(
+                "Path {} is neither a file nor a directory, cannot upload",
+                path.display()
+            ))
+        }
+        .unwrap_or_else(|e| {
+            panic!(
+                "Failed to create uploadable zip while processing {}: {}",
+                path.display(),
+                e
+            )
+        });
+
         normalized_zips.push(normalized_zip);
     }
 
@@ -54,21 +74,19 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
     Ok(())
 }
 
-fn validate_is_mobile_app(path: &Path) -> Result<()> {
+fn validate_is_mobile_app(path: &Path, byte_view: &ByteView) -> Result<()> {
     // Check for XCArchive (directory) first
-    if path.is_dir() && is_xcarchive_directory(path)? {
+    if path.is_dir() && is_xcarchive_directory(path) {
         return Ok(());
     }
 
-    let byteview = ByteView::open(path)?;
-
     // Check if the file is a zip file (then AAB or APK)
-    if is_zip_file(&byteview) {
-        if is_aab_file(&byteview)? {
+    if is_zip_file(byte_view) {
+        if is_aab_file(byte_view)? {
             return Ok(());
         }
 
-        if is_apk_file(&byteview)? {
+        if is_apk_file(byte_view)? {
             return Ok(());
         }
     }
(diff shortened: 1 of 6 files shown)


src/commands/mobile_app/upload.rs
```diff
@@ -24,13 +33,32 @@ pub fn make_command(command: Command) -> Command {
...
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
     let path_strings = matches
         .get_many::<String>("paths")
         .expect("paths argument is required");
 
+    let sha = matches
+        .get_one::<String>("sha")
+        .cloned()
+        .or_else(|| vcs::find_head().ok());
```
szokeasaurusrex: [rust fyi] To avoid cloning in a situation like this, where you sometimes have a reference (in this case, the `matches.get_one("sha")`), but sometimes need to store owned data (the `vcs::find_head().ok()`), you can use a [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html). 

A `Cow` is a smart pointer, implemented as an enum, which can store owned or borrowed data, using the `Cow::Owned` and `Cow::Borrowed` enum variants, respectively. 

You would use a `Cow` here as follows:

```suggestion
    let sha = matches
        .get_one("sha")
        .map(String::as_str)
        .map(Cow::Borrowed)
        .or_else(|| vcs::find_head().ok().map(Cow::Owned));
```

You'll also need to import `Cow` by adding `use std::borrow::Cow` to your imports.

szokeasaurusrex: Since cloning probably does not add too much overhead here, as it happens at most once when this command is run, it is fine to keep as is, just wanted to mention the possibility of using `Cow` in this situation

src/commands/mobile_app/upload.rs
```diff
@@ -24,13 +33,32 @@ pub fn make_command(command: Command) -> Command {
...
+
+    let build_configuration = matches
+        .get_one::<String>("build_configuration")
```
szokeasaurusrex: Specifying the type here is unnecessary, as the compiler can infer this is a `String` due to the `map(String::as_str)` call on the following line

```suggestion
        .get_one("build_configuration")
```

src/commands/mobile_app/upload.rs
```diff
@@ -77,15 +105,20 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
     }
 
-    for zip in normalized_zips {
-        println!("Created normalized zip at: {}", zip.path().display());
-        // TODO: Upload the normalized zip to the chunked uploads API
+    let config = Config::current();
+    let (org, project) = config.get_org_and_project(matches)?;
+
+    for (path, zip) in normalized_zips {
+        println!("Uploading file: {}", path.display());
```
szokeasaurusrex: I think printing this line for every file might make the output overly verbose. I would consider instead converting this to an info log.

By default, info logs are not shown, since we have the default log level at warning. Users need to set `SENTRY_LOG_LEVEL=info` or pass `--log-level=info` (or `debug`) to get info logs.

I think it is reasonable to print a success message at the end to inform users that the upload has succeeded. Perhaps we can even list all the files uploaded if there is a small number of them (perhaps less than 3); otherwise, the success message would only list the count of files uploaded.

```suggestion
        log::info!("Uploading file: {}", path.display());
```

src/commands/mobile_app/upload.rs
```diff
@@ -77,15 +105,20 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        let bytes = ByteView::open(zip.path())?;
+        upload_file(&bytes, &org, &project, sha.as_deref(), build_configuration)
+            .with_context(|| format!("Failed to upload file at path {}", path.display()))?;
+        println!("Successfully uploaded file at path {}", path.display());
```
szokeasaurusrex: Likewise here, `log::info!` may be more appropriate:


```suggestion
        log::info!("Successfully uploaded file at path {}", path.display());
```

src/commands/mobile_app/upload.rs
```diff
@@ -77,15 +105,20 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-        println!("Created normalized zip at: {}", zip.path().display());
-        // TODO: Upload the normalized zip to the chunked uploads API
+    let config = Config::current();
+    let (org, project) = config.get_org_and_project(matches)?;
+
+    for (path, zip) in normalized_zips {
+        println!("Uploading file: {}", path.display());
+        let bytes = ByteView::open(zip.path())?;
+        upload_file(&bytes, &org, &project, sha.as_deref(), build_configuration)
+            .with_context(|| format!("Failed to upload file at path {}", path.display()))?;
```
szokeasaurusrex: As written right now, handling this error with the question mark operator will cause us to immediately crash if any of the files cannot be uploaded. That would lead to the situation where all of the files before the failing file would be uploaded, but none of the ones afterwards would be.

I am wondering whether you think it might make more sense instead to attempt to upload all subsequent files after encountering an error, then after having attempted to upload everything, we would print out all the errors that occurred, listing the files we failed to upload, and exiting with nonzero status.

What do you think?

szokeasaurusrex: My initial thought is that this could be annoying in situations where the upload fails e.g. due to not having authorization to perform the requests. We could maybe just do a single request before the upload to ensure the token has the necessary permissions in order to avoid attempting the upload on every file without getting the same error for every attempted upload.

However, in situations where uploading only one of the files fails (e.g. because the file is invalid or something file-specific), still attempting to upload all other files probably results in a better user experience.

src/commands/mobile_app/upload.rs
```diff
@@ -133,7 +166,13 @@ fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
         .with_context(|| format!("Failed to get relative path for {}", path.display()))?;
 
     debug!("Adding file to zip: {}", file_name);
-    zip.start_file(file_name, SimpleFileOptions::default())?;
+
+    // Need to set the last modified time to a fixed value to ensure consistent checksums
+    let options = SimpleFileOptions::default()
+        .compression_method(zip::CompressionMethod::Stored)
+        .last_modified_time(DateTime::default());
```
szokeasaurusrex: Why are these "consistent checksums" necessary? For something in the backend? Seems like setting a hardcoded last modified time (which is apparently in 1980) could result in some strange behavior; I am not super familiar with zip files, but is there a way to clear the field instead?

src/commands/mobile_app/upload.rs
```diff
@@ -149,6 +188,9 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
     let mut zip = ZipWriter::new(temp_file.open()?);
 
     let mut file_count = 0;
+
+    // Collect and sort entries for deterministic ordering
```
szokeasaurusrex, line 228: The deterministic ordering is needed for the consistent checksums, I am assuming? Might be worth clarifying the comment if so

tests/integration/mobile_app/upload.rs
```diff
@@ -37,3 +41,109 @@ fn command_mobile_app_upload_invalid_xcarchive() {
...
+                        "missingChunks": []
+                    }"#
+                }
+                .into()
+            })
+            .expect(2),
+        )
+        .register_trycmd_test("mobile_app/mobile_app-upload-apk.trycmd")
+        .with_default_token();
+}
```
szokeasaurusrex, line 149: Just wondering, did you find it easy to implement these tests? I implemented this framework for testing chunk uploads a few months ago; it's kinda complex but it was the best I could come up with at that time. If you have any suggestions, please feel free to share, I am happy to improve this when I have time

szokeasaurusrex (approved):
Looks good! Please see my comments before merging, especially the one asking about how we want to be handling errors here

src/commands/mobile_app/upload.rs
```diff
@@ -149,6 +188,9 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
...
+    let mut entries = Vec::new();
```
szokeasaurusrex: Unfortunately, I cannot add this as a suggestion because part of the code is not being changed in this PR, but I would replace this line, the for loop on the following lines, and the following `sort_by` code with the snippet below:

```rust
    let entries = walkdir::WalkDir::new(path)
        .follow_links(true)
        .into_iter()
        .filter_map(Result::ok)
        .filter(|entry| entry.path().is_file())
        .map(|entry| {
            let entry_path = entry.into_path();
            let relative_path = entry_path.strip_prefix(path)?.to_owned();
            Ok((entry_path, relative_path))
        })
        .collect::<Result<Vec<_>>>()?
        .into_iter()
        .sorted_by(|(_, a), (_, b)| a.cmp(b));
```

Note that you'll also need to `use itertools::Itertools as _` to use the `sorted_by` on the final line.

Imo it is more idiomatic in Rust to construct a vector using iterator methods than to create a new vector and push to it from a `for` loop. It also avoids the need for a mutable variable, which good, since we don't have to keep track of where and how the vector can be mutated.

szokeasaurusrex: As a bonus, that snippet also demonstrates how you can deconstruct the structure in the sorting closure rather than needing to access the fields with `.1`. I think it is more readable this way, but we can also keep using the `.1` if you prefer

src/commands/mobile_app/upload.rs
rbro112: Thanks for the suggestion here, didn't know about `Cow`, very useful!

src/commands/mobile_app/upload.rs
rbro112: Yep, needed for the backend. Essentially the file ordering and last modified date affects the checksum. Since we're always creating a new zip file for the normalization here, the last modified time on each zip entry is "unstable" (i.e. defaults to now), but never actually used by us or the underlying filestore. Therefore, to ensure checksum stability, we essentially have to keep file ordering and the modified timestamps stable. This is important as the underlying chunked uploading will skip uploading chunks the backend already has, which I believe we'll want to support here.

I'll beef up the comments on this and also investigate if we can skip the `last_modified_time` entirely as I do agree that the default timestamp is not the perfect solution here. Nonetheless, @runningcode had to deal with a similar issue before in work with Gradle/caching and setting a default date was the workaround.

tests/integration/mobile_app/upload.rs
rbro112: Generally yes! The debugging of the tests was a tad difficult - I was trying to get dumps of the stdout/stderr using `TRYCMD=dump` but was noticing that didn't dump for non-`.trycmd` tests (e.g. `assert_cmd(...)` tests). That took a bit to realize as I was initially writing these as `assert_cmd` tests.

Once I was able to figure that out, it was pretty smooth sailing and generally the API here is pretty easy to use. The helpers on the mock endpoints are nice and `with_matcher` even helped me figure out the unstable checksum issue here.

I can follow with a PR to add some notes to the docs that I found useful (like the dump command and how to use).

src/commands/mobile_app/upload.rs
rbro112: Thanks for the tip! Definitely prefer your solution here, the `.1` is not important to me

src/commands/mobile_app/upload.rs
rbro112: Yeah I like that, I'll modify this to more gracefully handle the errors and log them out.

src/commands/mobile_app/upload.rs
szokeasaurusrex: Makes sense! Although just FYI, I am not 100% sure on this, but I believe that chunks are only stored for a short amount of time (maybe 12-24 hours) and I am unsure if they are persisted after assembly.

@jan-auer is my understanding here correct?

tests/integration/mobile_app/upload.rs
szokeasaurusrex: Yeah, `assert_cmd` and `trycmd` are completely separate libraries. Until a few months ago, all our tests were `trycmd` tests, but I found this setup a bit difficult when more precise assertions were needed for a specific command. `assert_cmd` is better designed for that use case, which is why I added it for these tests.

So basically, my general recommendation (and also what the respective libraries seem to recommend in their docs) is to use `assert_cmd` for a test of one specific command where you need to check for specific side effects, and use `trycmd` when it is only important to assert output but not necessarily other behavior (e.g. asserting that the `--help` text is correct)

src/commands/mobile_app/upload.rs
```diff
@@ -77,15 +105,54 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    if uploaded_paths.len() < 3 {
+        for path in &uploaded_paths {
+            println!("  - {}", path.display());
+        }
     }
 
-    eprintln!("Uploading mobile app files to a project is not yet implemented.");
+    if uploaded_paths.is_empty() {
+        bail!("Failed to upload any files");
+    }
```
szokeasaurusrex, line 155: I think we may want to error with nonzero exit status also if any of the files failed to upload, not just if nothing was uploaded. Otherwise, users who e.g. run this from CI will easily miss the warning message which says some fails errored.

What do you think?

src/commands/mobile_app/upload.rs
```diff
@@ -176,3 +265,117 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
...
+        sha.unwrap_or("unknown"),
+        build_configuration.unwrap_or("unknown")
+    );
+
+    let api = Api::current();
+    let authenticated_api = api.authenticated()?;
+
+    let chunk_upload_options = authenticated_api
+        .get_chunk_upload_options(org)?
+        .expect("Chunked uploading is not supported");
```
szokeasaurusrex, line 289: We should return an error, rather than using `expect` here. `expect` panics, and a Rust program should only ever panic if it ends up in an unexpected state, since panics are unrecoverable.

If a panic occurs, it necessarily indicates a bug in the program. Because of this, Sentry CLI has a custom panic hook which automatically augments any panic error message with a notice that they have encountered an "Internal error" and that they should report a bug.

However, in this case, it is expected that users could be trying to upload to a server which does not support chunk uploading. So, we should instead return an error here.

Since this is already merged, I'll fix this.

szokeasaurusrex: I also don't see where we check that the options are checked to ensure the server supports receiving mobile apps. It would be important to check that as well, not just make sure that we have chunk options.

PR #2552, opened 2025-06-10
meta: Add some testing notes to docs

rbro112:
After adding some tests in a recent PR, figured it could be good to add some notes about debugging and the `dump` command usage to the docs.


szokeasaurusrex (approved):
fwiw I am unsure how many folks actually read this document; other parts also seem a bit out of date (e.g. the API Mocks section)

rbro112:
@szokeasaurusrex fair, although I read it when adding test 😄 If there's a better place to add documentation around testing/the CLI happy to do so!

szokeasaurusrex:
Okay, I suppose if you read it than it is doing its job! Maybe we just need to think about properly updating these docs. I'll add it to the backlog

PR #2553, opened 2025-06-11
feat(login): Warn when overwriting global org auth token

szokeasaurusrex:
Closes #2573


szokeasaurusrex:
Fixed in https://github.com/getsentry/sentry-cli/pull/2554

PR #2559, opened 2025-06-13
ci: Add macos intel runner and use macOS 15

noahsmartin:
To ensure CI also compiles with Intel macOS


szokeasaurusrex (commented):
Not opposed to this, but let's first fully resolve our discussion on the #2545 PR and on our Slack discussion, before deciding whether testing on Intel is truly necessary

noahsmartin:
@szokeasaurusrex I looked into that but only saw how to compile the release build not run the tests like this CI workflow was doing, without running on an intel Mac. If you have ideas please go ahead and push to this PR/make a new one, either way is fine with me

szokeasaurusrex:
After looking into this some more, I don't think it is necessary to test the `x86_64` macOS binary. We also [release Linux binaries for several different architectures](https://github.com/getsentry/sentry-cli/blob/3b81fbc03b40fcfaa291b88b92475297f11efb7d/.github/workflows/build.yml#L14-L25), yet we only test one Linux binary (for whichever architecture is the default on the `ubuntu-24.04` GitHub runner). 

Issues preventing only one architecture are quite rare – so far, this has occurred only one time in the past year and a half that I have been maintaining Sentry CLI, so I think it is okay if we only catch such issues in the release process. No need to run extra CI actions for each pull request, which would only very rarely uncover problems, which would anyways get caught before release.

PR #2563, opened 2025-06-20
ci: Add a new option `silentLogs` for cleaner CI/build output

RulaKhaled:
This PR introduces a new silentLogs option that provides a middle ground between full verbose output and complete silence. When enabled, it suppresses verbose command output while still displaying essential feedback like success messages and error information.

```
const cli = new SentryCli('config', { silentLogs: true });
await cli.releases.new('v1.0.0');
// Output: "✓ Release v1.0.0 created"

await cli.sourcemaps.upload();  
// Output: "✓ Source maps uploaded"
```

### DIFF
--- a/js/helper.js
+++ b/js/helper.js
@@ -287,11 +287,17 @@ function getPath() {
  *  - `'rejectOnError'` to inherit stdio and reject the promise if the command
  *    exits with a non-zero exit code.
  * @param {boolean} silent Disable stdout for silents build (CI/Webpack Stats, ...)
+ * @param {boolean} silentLogs Show only errors and success messages, hide verbose logs
  * @param {string} [configFile] Relative or absolute path to the configuration file.
  * @param {Object} [config] More configuration to pass to the CLI
  * @returns {Promise.<string>} A promise that resolves to the standard output.
  */
-async function execute(args, live, silent, configFile, config = {}) {
+async function execute(args, live, silent, silentLogs, configFile, config = {}) {
+  // Ensure silent takes precedence over silentLogs
+  if (silent) {
+    silentLogs = false;
+  }
+
   const env = { ...process.env };
   if (configFile) {
     env.SENTRY_PROPERTIES = configFile;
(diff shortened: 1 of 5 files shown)


js/helper.js
```diff
@@ -348,6 +391,84 @@ function getProjectFlagsFromOptions({ projects = [] } = {}) {
...
+
+  // Only show success messages for high-impact operations customers care about
+  switch (command) {
+    case 'releases':
+      if (args[1] === 'new' && args[2]) {
+        return `✓ Release ${args[2]} created`;
+      } else if (args[1] === 'finalize' && args[2]) {
+        return `✓ Release ${args[2]} finalized`;
+      } else if (args[1] === 'files' && args[3] === 'upload-sourcemaps' && args[2]) {
+        return `✓ Source maps uploaded to release ${args[2]}`;
```
szokeasaurusrex: [nit] We should align the output here with the output for `sourcemaps upload` (note that `sentry-cli releases files upload-sourcemaps` is deprecated). Either both should include the release (if present), or likely simpler is just to remove the output from both.

js/helper.js
```diff
@@ -348,6 +391,84 @@ function getProjectFlagsFromOptions({ projects = [] } = {}) {
...
+      break;
+
+    case 'upload-proguard':
+      return `✓ ProGuard mappings uploaded`;
+
+    case 'upload-dif':
+      return `✓ Debug information files uploaded`;
+
+    case 'upload-dsym':
+      return `✓ dSYM files uploaded`;
```
szokeasaurusrex: These commands are both deprecated aliases of `debug-files upload`, so output should be the same

```suggestion
    case 'upload-dif':
    case 'upload-dsym':
      return `✓ Debug files uploaded`;
```

js/helper.js
```diff
@@ -348,6 +391,84 @@ function getProjectFlagsFromOptions({ projects = [] } = {}) {
...
+      if (args[1] === 'new') {
+        return `✓ Deploy created`;
+      }
+      break;
+
+    case 'mobile-app':
+      if (args[1] === 'upload') {
+        return `✓ Mobile app uploaded`;
+      }
+      break;
```
szokeasaurusrex: Let's not add this yet, this command is still a work in progress.

```suggestion
```

js/helper.js
```diff
@@ -348,6 +391,84 @@ function getProjectFlagsFromOptions({ projects = [] } = {}) {
...
+
+    case 'send-event':
+      return `✓ Event sent`;
+
+    case 'send-envelope':
+      return `✓ Envelope sent`;
+
+    case 'send-metric':
+      return `✓ Metric sent`;
```
szokeasaurusrex: Maybe not necessary to add this either, as metrics are deprecated, and will be rejected if users attempt to send them to Sentry SaaS

```suggestion
```

js/helper.js
```diff
@@ -348,6 +391,84 @@ function getProjectFlagsFromOptions({ projects = [] } = {}) {
   return projects.reduce((flags, project) => flags.concat('-p', project), []);
 }
 
+/**
+ * Determines an appropriate success message based on the command.
+ *
+ * @param {string[]} args Command line arguments passed to sentry-cli
+ * @returns {string|null} Success message to display or null if no message needed
```
szokeasaurusrex, line 414: Maybe it is preferable to type this just as string, using empty string instead of null? Not sure how we typically do it in JS, so I'll defer to you on this.

```suggestion
 * @returns {string} Success message to display or empty string if no message needed
```

js/helper.js
```diff
@@ -282,11 +282,17 @@ function getPath() {
  * @param {string[]} args Command line arguments passed to `sentry-cli`.
  * @param {boolean} live We inherit stdio to display `sentry-cli` output directly.
  * @param {boolean} silent Disable stdout for silents build (CI/Webpack Stats, ...)
+ * @param {boolean} silentLogs Show only errors and success messages, hide verbose logs
  * @param {string} [configFile] Relative or absolute path to the configuration file.
  * @param {Object} [config] More configuration to pass to the CLI
  * @returns {Promise.<string>} A promise that resolves to the standard output.
  */
-async function execute(args, live, silent, configFile, config = {}) {
+async function execute(args, live, silent, silentLogs, configFile, config = {}) {
```
szokeasaurusrex, line 295: Is this part of public API?

RulaKhaled: I think users should only use the execute method provided by the SentryCli class

js/helper.js
RulaKhaled: Returning null expresses that "there is no success message to show" rather than "the success message is empty".
It is more consistent with JavaScript best practices for optional values.

szokeasaurusrex:
@RulaKhaled I am currently doing a clean-up of stale draft PRs in the repo. 

Are you still working on this change? Would you like help to complete it?

If not, please close the PR 🙏

Issue #2577, opened 2025-07-06
sentry-cli login Fails on macOS but Works on Windows and Ubuntu


sxsx2yzyz:
### CLI Version

sentry-cli 2.46.0

### Operating System and Architecture

- [x] macOS (arm64)
- [ ] macOS (x86_64)
- [ ] Linux (i686)
- [ ] Linux (x86_64)
- [ ] Linux (armv7)
- [ ] Linux (aarch64)
- [ ] Windows (i686)
- [ ] Windows (x86_64)

### Operating System Version

14.6.1 (23G93)

### CLI Command

sentry-cli --url https://<my-sentry-url> login --auth-token [REDACTED]

### Exact Reproduction Steps

1. Use macOS device with sentry-cli version 2.46.0 installed.
2. Ensure `.sentryclirc` file with the following contents exists at `/Users/<REDACTED_USER>/.sentryclirc`:

```bash
[defaults]
url = https://<REDACTED_DOMAIN>
[log]
level = trace
```

3. Execute the following command (replace `<REDACTED_TOKEN>` with a valid token):
```bash
sentry-cli --url https://<REDACTED_DOMAIN> login --auth-token <REDACTED_TOKEN>
```

4.	Observe the output error.
5.	Repeat the same command on Windows or Ubuntu machines with the same token and verify it succeeds.
6.	Optionally, disconnect internet on macOS device and rerun the command to observe error behavior.

### Expected Results

1. The sentry-cli login command should successfully authenticate against the self-hosted Sentry server (https://<REDACTED_DOMAIN>) on macOS.
2. The command should complete without errors and confirm a successful login.
3. The same command and token should work consistently across all operating systems (macOS, Windows, Ubuntu).
4. When internet is disconnected, the command should fail with a network-related error or retry, but should not immediately return an “Invalid token” error.

### Actual Results

1.On macOS, running sentry-cli login with a valid token returns:

```bash
error: Invalid token: API request failed
```

2.The error occurs regardless of internet connection status (both connected and disconnected).
3.The same token and command succeed without issue on Windows and Ubuntu.
4.Logs show the command attempts the request, but fails locally with “Invalid token” without proper network validation.

### Logs

```bash
<REDACTED_USER>@MacBook-Pro-2:led_ble_wifi_esp32_8_4zp <REDACTED_USER>$ sentry-cli --url https://<REDACTED_DOMAIN> login --auth-token <REDACTED_TOKEN>
[...]

PR #2578, opened 2025-07-07
fix(macOS): Fix sentry-cli login failure behind proxy on macOS

sxsx2yzyz:
## Problem
- `sentry-cli login` fails with "Invalid token: API request failed" on macOS behind proxy
- Works fine on Windows/Linux with same token and network config
- curl command works normally

## Root Cause
Static linking of curl (`static-curl`, `static-ssl`) has compatibility issues with HTTPS CONNECT tunnels on macOS, causing `perform()` to fail after proxy connection is established.

## Solution
- Remove static linking features from curl dependency
- Use system curl library instead
- Revert all temporary macOS-specific hacks

## Testing
- ✅ Login works behind HTTP/HTTPS proxy on macOS
- ✅ No regression on Linux/Windows
- ✅ All other commands work normally

## Related Issue
Closes #[2577]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ clap = { version = "4.1.6", default-features = false, features = [
 ] }
 clap_complete = "4.4.3"
 console = "0.15.5"
-curl = { version = "0.4.46", features = ["static-curl", "static-ssl"] }
+curl = { version = "0.4.46" }
 dirs = "4.0.0"
 dotenvy = "0.15.7"
 elementtree = "1.2.3"


szokeasaurusrex:
Hi @sxsx2yzyz, while I appreciate the contribution, switching away from the statically-linked CURL-library seems like a high-risk change which could introduce other unforeseen problems. Also, in Sentry CLI, we generally have been trying to statically link most libraries which are needed.

So, we are going to need to find another solution for #2577. If there is any information/links you found during your investigation of potential solutions which could be relevant, please feel free to share them, thanks!

PR #2580, opened 2025-07-08
feat(api): Improve API error messages

szokeasaurusrex:
Include the source of the error in `Display` output (when present), and also improve how we display generic `RequestFailed` error messages.

For example, prior to this change, if calling `sentry-cli login` with an invalid token, one would receive the following error message:

```
Invalid token: API request failed
```

After the change, the error message is:

```
Invalid token: API request failed, due to the following error: sentry reported an error: Invalid token (http status: 401)
```

This change should help debug #2577, as in that issue, `sentry-cli` is failing due to a strange generic `RequestFailed` API error.


szokeasaurusrex:
closing following offline discussion about not putting source in Display impl

PR #2583, opened 2025-07-08
feat(mobile-app): Make missing path not panic

rbro112:
Updates the mobile-app upload command to no longer panic upon a missing path argument, rather log an error and exit gracefully. Adds a test to validate this behavior.


rbro112:
* **#2583** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2583?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/commands/mobile_app/upload.rs
```diff
@@ -301,14 +305,12 @@ fn upload_file(
...
-        .ok_or_else(|| {
-            anyhow!(
-                "The Sentry server lacks chunked uploading support, which \
+    let chunk_upload_options = api.get_chunk_upload_options(org)?.ok_or_else(|| {
+        anyhow!(
+            "The Sentry server lacks chunked uploading support, which \
                 is required for mobile app uploads. {SELF_HOSTED_ERROR_HINT}"
-            )
-        })?;
+        )
```
rbro112, line 312: Looks like an autoformat update.

src/commands/mobile_app/upload.rs
```diff
@@ -49,9 +49,13 @@ pub fn make_command(command: Command) -> Command {
...
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
-    let path_strings = matches
-        .get_many::<String>("paths")
-        .expect("paths argument is required");
+    let path_strings: Vec<_> = match matches.get_many::<String>("paths") {
+        Some(paths) => paths.collect(),
+        None => {
+            eprintln!("path argument is required");
+            return Ok(());
```
szokeasaurusrex: We should instead return an error so that we exit with nonzero status

rbro112: I took this impl from the proguard mapping upload, will update but notable that that command does not return an error.

src/commands/mobile_app/upload.rs
```diff
@@ -49,9 +49,12 @@ pub fn make_command(command: Command) -> Command {
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
-    let path_strings = matches
-        .get_many::<String>("paths")
-        .expect("paths argument is required");
+    let path_strings: Vec<_> = match matches.get_many::<String>("paths") {
+        Some(paths) => paths.collect(),
+        None => {
+            return Err(anyhow!("path argument is required"));
```
szokeasaurusrex: Sorry I did not realize this earlier in the review, but I believe there should be a way to get clap to enforce the path argument as required? 

If you cannot get it to work, I'd try to write a better error message, as path is a positional argument, and as written, the error message could be confusing to users since there is no --path argument. 

Something like "you must provide at least one path to upload" I think would be better

src/commands/mobile_app/upload.rs
```diff
@@ -49,9 +50,13 @@ pub fn make_command(command: Command) -> Command {
...
-    let path_strings = matches
-        .get_many::<String>("paths")
-        .expect("paths argument is required");
+    let path_strings: Vec<_> = match matches.get_many::<String>("paths") {
+        Some(paths) => paths.collect(),
+        None => {
+            // Clap will handle the case where no paths are provided.
+            return Ok(());
+        }
+    };
```
szokeasaurusrex: Since clap is handling this already (or at least, as it should be), I would go back to the expect (which panics) since getting into the state where we don't have paths at this point is unexpected and would indicate a bug in Sentry CLI. You can update the expect message to state that Clap is enforcing that the argument is required.

szokeasaurusrex (approved):
See comment before merge, looks good otherwise!

PR #2585, opened 2025-07-08
fix(launchpad): Ensure mobile artifact's zip structures are preserved

NicoHinderling:
WIthout this, the code would remove the xcarchive wrapper directory, which causes issues when trying to parse the artifact in launchpad

### DIFF
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -399,6 +399,7 @@ fn poll_assemble(
     Ok(())
 }
 
+#[cfg(not(windows))]
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -250,8 +250,16 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
         .into_iter()
         .filter_map(Result::ok)
         .filter(|entry| entry.path().is_file())
-        .map(|entry| entry.into_path())
-        .sorted_by(|a, b| a.cmp(b));
+        .map(|entry| {
+            let entry_path = entry.into_path();
+            let relative_path = entry_path.strip_prefix(
+                path.parent().ok_or_else(|| anyhow!("Cannot determine parent directory for path: {}", path.display()))?
+            )?.to_owned();
+            Ok((entry_path, relative_path))
+        })
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .sorted_by(|(_, a), (_, b)| a.cmp(b));
 
     // Need to set the last modified time to a fixed value to ensure consistent checksums
     // This is important as an optimization to avoid re-uploading the same chunks if they're already on the server
@@ -260,13 +268,10 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
         .compression_method(zip::CompressionMethod::Stored)
         .last_modified_time(DateTime::default());
 
-    for entry_path in entries {
-        let zip_path = entry_path.strip_prefix(
-            path.parent().ok_or_else(|| anyhow!("Failed to get parent directory"))?
-        )?.to_owned();
-        debug!("Adding file to zip: {}", zip_path.display());
+    for (entry_path, relative_path) in entries {
+        debug!("Adding file to zip: {}", relative_path.display());
 
-        zip.start_file(zip_path.to_string_lossy(), options)?;
+        zip.start_file(relative_path.to_string_lossy(), options)?;
         let file_byteview = ByteView::open(&entry_path)?;
         zip.write_all(file_byteview.as_slice())?;
         file_count += 1;
@@ -418,7 +423,6 @@ mod tests {
         let mut archive = ZipArchive::new(zip_file)?;
         let file = archive.by_index(0)?;
         let file_path = file.name();
-        
         assert_eq!(file_path, "MyApp.xcarchive/Products/app.txt");
         Ok(())
     }
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -409,7 +409,7 @@ mod tests {
     fn test_normalize_directory_preserves_top_level_directory_name() -> Result<()> {
         let temp_dir = crate::utils::fs::TempDir::create()?;
         let test_dir = temp_dir.path().join("MyApp.xcarchive");
-        fs::create_dir_all(&test_dir.join("Products"))?;
+        fs::create_dir_all(test_dir.join("Products"))?;
         fs::write(test_dir.join("Products").join("app.txt"), "test content")?;
 
         let result_zip = normalize_directory(&test_dir)?;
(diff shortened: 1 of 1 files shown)


src/commands/mobile_app/upload.rs
```diff
@@ -252,7 +259,9 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
         .map(|entry| {
             let entry_path = entry.into_path();
             let relative_path = entry_path.strip_prefix(path)?.to_owned();
-            Ok((entry_path, relative_path))
+            // Preserve the directory structure by including the directory name
+            let full_relative_path = Path::new(dir_name).join(relative_path);
+            Ok((entry_path, full_relative_path))
```
szokeasaurusrex: Isn't the entry_path already equivalent to what you would be computing with the full_relative_path? Unless I am missing something, it looks like you are stripping the prefix only to add the same one back 😂

```suggestion
            Ok(entry_path)
```

szokeasaurusrex: Of course, you will also need to make some other modifications in response to this change, but the Rust type checker should complain for anything that this change breaks.

src/commands/mobile_app/upload.rs
```diff
@@ -249,14 +249,10 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
...
         .filter_map(Result::ok)
         .filter(|entry| entry.path().is_file())
-        .map(|entry| {
-            let entry_path = entry.into_path();
-            let relative_path = entry_path.strip_prefix(path)?.to_owned();
-            Ok((entry_path, relative_path))
-        })
+        .map(|entry| Ok(entry.into_path()))
         .collect::<Result<Vec<_>>>()?
         .into_iter()
```
szokeasaurusrex: Wrapping this in an `Ok` is unnecessary, because it is an infallible operation. Also, no need to collect into a vector just to make an iterator again.

```suggestion
        .map(|entry| entry.into_path())
```

src/commands/mobile_app/upload.rs
```diff
@@ -265,10 +261,13 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
...
-    for (entry_path, relative_path) in entries {
-        debug!("Adding file to zip: {}", relative_path.display());
+    for entry_path in entries {
+        let zip_path = entry_path.strip_prefix(
+            path.parent().ok_or_else(|| anyhow!("Failed to get parent directory"))?
+        )?.to_owned();
+        debug!("Adding file to zip: {}", zip_path.display());
 
-        zip.start_file(relative_path.to_string_lossy(), options)?;
+        zip.start_file(zip_path.to_string_lossy(), options)?;
```
szokeasaurusrex: How is this `zip_path` different from the `relative_path` we had before?

szokeasaurusrex (commented):
Please see inline comments.

Also, I am struggling to see how the behavior has changed following these changes. As far as I can tell, this code should behave exactly the same as what we had before because it looks to me like the `zip_path` is equivalent to the previous `relative_path`. 

Can you please explain what the desired difference in behavior is? If possible, please add a test case, which fails against the original code, but passes against the new code, to verify the desired behavior. This will also help me understand the purpose of this change, which I am currently struggling to grasp.

src/commands/mobile_app/upload.rs
NicoHinderling: The top level path is no longer being stripped.

Example: `entry_path = /full/path/to/MyApp.xcarchive/Products/app.txt`

relative_path stripped the entire path (/full/path/to/MyApp.xcarchive), losing the directory name
zip_path strips only the parent path (/full/path/to), keeping the directory name

Before: `relative_path = Products/app.txt`
After: `zip_path = MyApp.xcarchive/Products/app.txt`

NicoHinderling:
> Please see inline comments.
> 
> Also, I am struggling to see how the behavior has changed following these changes. As far as I can tell, this code should behave exactly the same as what we had before because it looks to me like the `zip_path` is equivalent to the previous `relative_path`.
> 
> Can you please explain what the desired difference in behavior is? If possible, please add a test case, which fails against the original code, but passes against the new code, to verify the desired behavior. This will also help me understand the purpose of this change, which I am currently struggling to grasp.

It's about preserving the top directory in the path. I've added a unit test 👍

src/commands/mobile_app/upload.rs
```diff
@@ -401,3 +398,28 @@ fn poll_assemble(
 
     Ok(())
 }
+
+#[cfg(not(windows))]
```
szokeasaurusrex, line 407: Why not run on Windows?

src/commands/mobile_app/upload.rs
```diff
@@ -401,3 +398,28 @@ fn poll_assemble(
...
+        let test_dir = temp_dir.path().join("MyApp.xcarchive");
+        fs::create_dir_all(test_dir.join("Products"))?;
+        fs::write(test_dir.join("Products").join("app.txt"), "test content")?;
+
+        let result_zip = normalize_directory(&test_dir)?;
+        let zip_file = fs::File::open(result_zip.path())?;
+        let mut archive = ZipArchive::new(zip_file)?;
+        let file = archive.by_index(0)?;
+        let file_path = file.name();
+
```
szokeasaurusrex: not sure why the formatter did not catch this

```suggestion

```

src/commands/mobile_app/upload.rs
```diff
@@ -266,10 +260,13 @@ fn normalize_directory(path: &Path) -> Result<TempFile> {
         .compression_method(zip::CompressionMethod::Stored)
         .last_modified_time(DateTime::default());
 
-    for (entry_path, relative_path) in entries {
-        debug!("Adding file to zip: {}", relative_path.display());
+    for entry_path in entries {
+        let zip_path = entry_path.strip_prefix(
+            path.parent().ok_or_else(|| anyhow!("Failed to get parent directory"))?
+        )?.to_owned();
```
szokeasaurusrex: There's no need to call `to_owned` here; it just causes an unnecessary memory allocation and the code compiles without it. 

```suggestion
            )?;
```

src/commands/mobile_app/upload.rs
szokeasaurusrex: Okay! I think I see the difference now. Whereas previously, you called `strip_prefix(path)`, you now call `strip_prefix(path.parent)`, correct?

Assuming the above is correct, I have two questions:
  1. Couldn't you have simply added the path.parent call where we were previously computing the `relative_path`? It seems like you are also refactoring the code by moving this changed strip_prefix into the for loop. Is this refactor causing some other functionality change which I am missing?
  2. Is it reasonable to assume that the path should always have a parent? E.g., what if we are running this from the root directory? Not that I think it would be super common, but perhaps there are some scenarios where this could occur (e.g. if running from within a Docker container)?

src/commands/mobile_app/upload.rs
NicoHinderling: The unit test just fails because instead of `MyApp.xcarchive/Products/app.txt` it's something like `C:\\MyApp.xcarchive\\Products\\app.txt`. Lets just skip

src/commands/mobile_app/upload.rs
NicoHinderling: 1. No I think you're right that the for loop refactor was not necessary. I've changed it to be this way, apologies for the thrash.
2. Yes it is but I've added a `ok_or_else` just in case

PR #2600, opened 2025-07-14
fix: ignore non dsym files

itaybre:
This is not a PR ready to merge, but shows what we need to check for to ignore non dsym files

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1158,6 +1158,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "goblin"
+version = "0.10.0"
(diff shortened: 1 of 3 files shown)


szokeasaurusrex:
FWIW I don't believe we want to ignore non-dsym files by default, as this would be potentially breaking. Perhaps having an option to ignore such files could be helpful, although even there I am unsure, as users could simply specify to the path to the specific debug file(s) they want to upload

szokeasaurusrex:
@itaybre given lack of activity here, I am assuming we can probably close this PR. Please feel free to reopen if needed 🙏

PR #2601, opened 2025-07-15
feat(mobile-app): Remove unstable-mobile-app feature gating

rbro112:
We're ready to start end-to-end testing this, so removing the `unstable-mobile-app` flag that was added in https://github.com/getsentry/sentry-cli/pull/2582/ to re-enable the `mobile-app` command without the explicit flag.


rbro112:
* **#2601** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2601?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

szokeasaurusrex:
I will wait to review this until our sync meeting next week, so I can provide more informed feedback. Please let me know if you need me to review sooner

Cargo.toml
```diff
@@ -11,7 +11,7 @@ rust-version = "1.86"
 [dependencies]
 anylog = "0.6.3"
 anyhow = { version = "1.0.69", features = ["backtrace"] }
-apple-catalog-parsing = { path = "apple-catalog-parsing", optional = true }
```
rbro112, line 14: @szokeasaurusrex, I had to remove the optional flag to get this to compile properly. Admittedly I'm a bit unsure of the effect that removing this optionality would have as it pertains to the release (and how you had the dependency seemingly conditionally included with the `unstable-mobile-app` feature), so if you don't mind a check/fix here, that would be great.

.github/workflows/test.yml
```diff
@@ -16,20 +16,15 @@ jobs:
...
         os: [ubuntu-24.04, macos-14, windows-2022]
-        feature-args: ['', '-Funstable-mobile-app']
         include:
           - os: ubuntu-24.04
             display-os: Linux
           - os: macos-14
             display-os: macOS
           - os: windows-2022
             display-os: Windows
-          - feature-args: ''
```
rbro112, line 27: Removed this too as it was originally added in https://github.com/getsentry/sentry-cli/pull/2582, but the empty args could be useful for having around if future flags are added so `${{ matrix.feature-suffix }}` won't need to be added below again. Will leave to you @szokeasaurusrex if that's alright.

.github/workflows/lint.yml
```diff
@@ -26,8 +26,6 @@ jobs:
             display-os: Windows
           - feature-args: ''
```
rbro112: These were originally added in https://github.com/getsentry/sentry-cli/pull/2582, but I left these in as the empty args could be useful for having around if future flags are added so `${{ matrix.feature-suffix }}` won't need to be added below again. 

Will leave to you @szokeasaurusrex to remove if you'd prefer.

.github/workflows/test.yml
```diff
@@ -26,10 +26,8 @@ jobs:
             display-os: Windows
           - feature-args: ''
```
rbro112: Similar to `lint.yml`, left these in as I figured it's easier to leave around for future feature flags

Cargo.toml
szokeasaurusrex: I think this is fine. The crate should compile on all platforms; on non-macOS platforms it is just an empty crate.

I can try to see though if I can add it as a macOS-specific dependency

.github/workflows/lint.yml
szokeasaurusrex: Let's remove – otherwise, if we never end up using this again, future maintainers could be confused about why this even exists.

We can always add this back if needed

szokeasaurusrex (approved):
Looks good after my minor changes, thanks!

PR #2599, opened 2025-07-14
ci: Restructure CI process

szokeasaurusrex:
This PR contains several improvements to our CI process which aim to make the CI workflow cleaner and to make it easier to configure required tests in code.

## Restructuring

Before the change:

image

After the change:
image

After the change, with some of the groups expanded:
image

As can be observed in the screenshots, this restructuring makes it clearer what all the CI check is doing, when viewing the workflow summary in CI.

The restructure also moves the Swift tests from a separate workflow file into the main `ci.yml` file.

## Functional changes

This PR also adds a "Check required jobs" job, which passes only if all other jobs in `ci.yml` pass. We intend to make this check the only required status from `ci.yml` to make PRs mergeable. 

Making "Check required jobs" required will indirectly make the previously non-required checks required: Test Node, Test Swift, and the `-Funstable-mobile-app` tests. With this change, it is now also possible to configure the required checks in code.

Also, this PR adds an optimization, so that any in progress jobs are cancelled if a new commit is pushed to the branch.

Closes #2509
Closes #2558
Closes #2597

PR #2602, opened 2025-07-15
ci: Cancel in-progress checks if new commit pushed

szokeasaurusrex:


szokeasaurusrex:
Closing; I wanted to split this off from #2599, but we should only do it afterwards, since the "Lint" test passes if the dependent jobs are cancelled

PR #2605, opened 2025-07-15
feat(js): Add `live: rejectOnError` execution mode to `execute` method

Lms24:


js/__tests__/helper.test.js
```diff
@@ -22,6 +22,30 @@ describe('SentryCli helper', () => {
...
   });
 
+  describe('execute', () => {
+    test('execute with live=false returns stdout', async () => {
+      const output = await helper.execute(['--version'], false);
+      expect(output.trim()).toBe('sentry-cli DEV');
+    });
+
+    test('execute with live=true resolves without output', async () => {
+      // TODO (v3): This should resolve with a string, not undefined/void
```
szokeasaurusrex, line 32: Please open an issue for this, if you have not already

js/helper.js
```diff
@@ -321,15 +325,26 @@ async function execute(args, live, silent, configFile, config = {}) {
...
       const pid = childProcess.spawn(getPath(), args, {
         env,
         // stdin, stdout, stderr
         stdio: ['ignore', output, output],
       });
-      pid.on('exit', () => {
+      pid.on('exit', (exitCode) => {
+        if (live === 'rejectOnError') {
+          if (exitCode === 0) {
+            resolve('success (live mode)');
```
szokeasaurusrex, line 341: Do we need to pass a value here? If not, maybe I'd just resolve with `resolve()`, but I'll defer to you here since I am not really familiar with JS

js/releases/__tests__/index.test.js
```diff
@@ -7,14 +7,14 @@ describe('SentryCli releases', () => {
   test('call sentry-cli releases propose-version', () => {
     expect.assertions(1);
     const cli = new SentryCli();
-    return cli.releases.proposeVersion().then(version => expect(version).toBeTruthy());
+    return cli.releases.proposeVersion().then((version) => expect(version).toBeTruthy());
```
szokeasaurusrex: are these just formatting changes?

js/releases/index.js
```diff
@@ -199,7 +199,10 @@ class Releases {
 
       return uploadPaths.map((path) =>
         // `execute()` is async and thus we're returning a promise here
-        this.execute(helper.prepareCommand([...args, path], SOURCEMAPS_SCHEMA, newOptions), true)
```
szokeasaurusrex, line 202: so previously, we completely ignored the `live` argument?

szokeasaurusrex (approved):
Looks good, see comments prior to merge

szokeasaurusrex:
Also perhaps would be good for someone more well-versed in JS to also take a glance at this before we merge

js/__tests__/helper.test.js
Lms24: good point! done in https://github.com/getsentry/sentry-cli/issues/2606

js/helper.js
Lms24: Yes we should because the return type according to the type definition is `Promise<string>` and doesn't allow `undefined` or resolving without a value.

js/releases/__tests__/index.test.js
Lms24: hmm yes, good catch! I'll remove for this to keep the PR clutter-free (also not sure why it updated yarn lock. Will check and revert if possible).

js/releases/index.js
Lms24: Not really -- `options` was not supposed to include the `live` key. However, it is (for better or worse) simply typed as `object` so it allows for any key to be passed in. I used this as an easy way to allow anyone calling `uploadSourceMaps(release, options)` to change the `live` value if they want to. We'll use this in the bundler plugin to run `uploadSourceMaps` with `live: rejectOnError`.

PR #2607, opened 2025-07-15
chore(js): Unstale yarn lock

Lms24:
Looks like yarn.lock was stale and missed am entry for a new(?) binary

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -665,40 +665,45 @@
     estree-walker "^2.0.1"
     picomatch "^2.2.2"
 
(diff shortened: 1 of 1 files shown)


szokeasaurusrex (approved):
new binary is the Windows on ARM binary. 

Any clue why this was not getting updated? It probably should happen automatically, right?

Lms24:
You have to run `yarn install` once locally and push up the changes whenever you add pacakge (which I guess is what happened here). Usually, when you update or add a dependency, yarn.lock updates automatically, yes.

Issue #2606, opened 2025-07-15
[v3] Fix JS interface `execute` promise resolve values


Lms24:
### Problem Statement

According to our JS interface [type definitions](https://github.com/getsentry/sentry-cli/blob/49afbb4825fd6109c17bc01c53f8fb8e89e7b8fc/js/index.d.ts#L202-L237), the `execute` promise, as well as all abstractions (e.g. `releases.uploadSourceMaps`) building on top of it always have to return a `Promise<string>`. 
However, in reality, if you execute a command with `live: true` (the second `execute` param), we resolve with void/undefined, thereby violating our type definitions.

### Solution Brainstorm

To avoid behaviour breaking changes, we can leave this as-is for now but we should fix it in v3 of the CLI. We have two options:

- always resolve with a string (I'd prefer this slightly)
- change type definitions to allow for `Promise<string | undefined>`

szokeasaurusrex:
@Lms24 what is your suggestion for fixing this? Should we just make the behavior defined by `live === 'rejectOnError'` the default behavior for `live === true` and perhaps also remove the `'rejectOnError'` option?<br><br>[https://github.com/getsentry/sentry-cli/blob/51273eaf2c89be56c7a62a99c41b0b47f39a49c4/js/helper.js#L339-L350](https://github.com/getsentry/sentry-cli/blob/51273eaf2c89be56c7a62a99c41b0b47f39a49c4/js/helper.js#L339-L350)

Lms24:
Making `rejectOnError` the behaviour for `true` and removing this value sounds reasonable to me! We could always introduce something like `'ignoreRejection'` if there's a use case we need for live mode without promise rejections. I skimmed over our bundler plugins and didn't find anything where this change would be problematic.

Behaviour for `live: false` should remain unchanged though.

PR #2611, opened 2025-07-15
chore: Add `typescript` to type check JS Interface via JSDoc

### LINKED ISSUE #2604: Enable Type Checking in JS Interface
Our current type definitions of the `@sentry/cli` JS interface are not 100% correct/adhered to and allowed for some nasty bugs to sneak in (e.g. https://github.com/getsentry/sentry-cli/issues/2062). We should enable TypeScript type checking.

I think this would be an ideal case for JS-based TypeScript type checking where we don't transpile anything or write actual TS but type check via JSDoc. Given we already use type information in JSDoc this should be a seamless transition and we get type safety for cheap.

Lms24:
This PR adds `typescript` to the repo to type-check the JS interface of Sentry CLI. I decided to build on top of the already existing JSDoc type annotations but actually enforce them by running TS. 

This also surfaced an incorrect type annotation which I fixed as well, along with adding some minor annotations that were necessary. In light of #2606, I also had to ignore a type error but we'll fix this in the next major.

Added a type checking job to CI, which IIUC should also be required after #2599.

closes #2604


.github/workflows/test_node.yml
```diff
@@ -32,4 +48,4 @@ jobs:
       - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install @vercel/nft@0.22.1
         if: matrix.node-version == '10.x' || matrix.node-version == '12.x'
 
-      - run: npm test
+      - run: npm run test
```
szokeasaurusrex: just wondering – what exactly is the distinction here?

.github/workflows/test_node.yml
```diff
@@ -8,6 +8,22 @@ on:
...
+
+      - name: Use Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # 4.4.0
+        with:
+          node-version-file: package.json
+
+      # We need to skip the fallback download because downloading will fail on release branches because the new version isn't available yet.
+      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install
+
+      - run: npm run check:types
```
szokeasaurusrex, line 25: Further changes might be required to make this check required – as you can see, above, the output for this workflow is defined as the output for `test_node`.

I can try to play around with this on a branch based on top of this PR. But, let's hold off on merging until we have ensured that the type check action is indeed being enforced

js/releases/options/deploys.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @type {import('../../helper').OptionsSchema}
```
szokeasaurusrex, line 2: [question] just wondering, what does this do?

js/helper.js
```diff
@@ -330,6 +331,7 @@ async function execute(args, live, silent, configFile, config = {}) {
         stdio: ['ignore', output, output],
       });
       pid.on('exit', () => {
+        // @ts-expect-error - this is a TODO (v3) to fix and resolve a string here
```
szokeasaurusrex, line 334: If I understand correctly, this allows a TS error on the next line. 

Once the error is fixed, if someone forgets to remove this comment, would the type checker complain about this comment being unnecessary? Also, is there any way to more precisely specify the exact type of error we expect here, so that we don't accidentally ignore something we don't mean to?

package-lock.json
szokeasaurusrex, line 1: Did you intentionally commit this file? If so, why is it needed now, if we did not need it before?

tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "checkJs": true,
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "target": "ES2015",
+    "moduleResolution": "node",
+    "skipLibCheck": true
+  },
+  "include": ["js/**/*.js"],
+  "exclude": ["**/__tests__/**", "**/__mocks__/**"]
+}
```
szokeasaurusrex: [nit] Don't forget the newline at the end of the file 🙂

```suggestion
}

```

yarn.lock
szokeasaurusrex, line 1: are these changes intentional? If so, why are they needed here?

.github/workflows/test_node.yml
Lms24: no difference actually. I was just weirded out by `npm test` for a second because usually you have to call `npm run <script>` for all the scripts registered in `package.json`. However, `npm test` is an exception. Will revert.

js/releases/options/deploys.js
Lms24: This is similar to a typescript type cast where I tell the type checker that the object this file exports is of type `OptionsSchema` declared in `../../helper.js`. 

The reason we need this here is because typscript infers that the `type` property of the individual options is of type `string`. When we [import](https://github.com/getsentry/sentry-cli/blob/d227ec6ce013e0d8a6c04151ab0f5a17365e8097/js/releases/index.js#L17-L22) this file though, we want `type` to be only specific strings `'array'|'string'|'boolean'|'inverted-boolean'` (see [here](https://github.com/getsentry/sentry-cli/blob/d227ec6ce013e0d8a6c04151ab0f5a17365e8097/js/helper.js#L179-L182)). This way of specifying the schema is a bit unusual so we need to "help" TS to infer the type correctly here.

js/helper.js
Lms24: > Once the error is fixed, if someone forgets to remove this comment, would the type checker complain about this comment being unnecessary?

Yes, exactly. That's the advantage of `@ts-expect-error` over `@ts-ignore` (which doesn't raise an error if there's no type error to ignore anymore) 

> Also, is there any way to more precisely specify the exact type of error we expect here, so that we don't accidentally ignore something we don't mean to?

Good question! I looked it up but looks like this is still an open issue (see https://github.com/microsoft/TypeScript/issues/19139)

package-lock.json
Lms24: ah it all makes sense now (see private convo about package-lock). Bad cursor! 
thanks for flagging!

yarn.lock
Lms24: Most of them were not, thanks for flagging! I cleaned it up and now the lock file only contains the added dependencies from typescript and @types/node.

.github/workflows/test_node.yml
Lms24: I _think_ the final "Check required jobs" job would fail if only the type checking job failed. To me it looks like we're requiring the workflow result of `test_node` in the "Check required jobs" job and the workflow would fail if type checking failed. But I'd leave this up to you to confirm. I won't merge without your approval :)

szokeasaurusrex: Yep, you are correct, [verified on this PR](https://github.com/getsentry/sentry-cli/pull/2616). I guess I don't really know what the `output` does in this case; perhaps I should get rid of it (of course, this has nothing to do with your PR though, so I will handle it)

js/releases/options/deploys.js
szokeasaurusrex: Okay, I guess I kinda understand 😅

Appreciate the explanation

Lms24:
> One final question though (non-blocking for this PR): Does the type checker enforce that everything has types declared (at least, that the public API does)? Or, is it possible that we could expose something which is untyped?

It doesn't enforce that the types are declared publicly (i.e. for library consumers). At least not in the current setup. We could change this eventually though, by telling typescript to not only type-check but also emit type declaration files. Might be a nice follow-up but I don't have time for this today anymore. Opened https://github.com/getsentry/sentry-cli/issues/2617 to track.

Within the codebase, typescript simply infers the types of all non-explicitly annotated variables and exports. So if I have something like

```js
module.exports.x = "foo"
```

TS will infer that `x` is of type `string` and apply the inferred type if `x` is consumed in any other module.

PR #2619, opened 2025-07-16
feat(mobileapp): Update mobile app command for ipa uploads

noahsmartin:
Add support for uploading iOS .ipa files to the `mobile-app` command by converting them to xcarchive format.


src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+    let mut app_name = String::new();
+
+    // Extract .app from Payload/ directory
+    for i in 0..ipa_archive.len() {
+        let mut file = ipa_archive.by_index(i)?;
+        let file_path = file.name().to_string();
+
+        if let Some(stripped) = file_path.strip_prefix("Payload/") {
+            if let Some(app_folder_name) = stripped.strip_suffix(".app/") {
+                app_name = app_folder_name.to_string();
```
szokeasaurusrex: From what I can tell here, the `app_name` variable could get defined multiple times as we loop through this directory. However, the `Info.plist` will only get the final value of `app_name`.

Do we expect there to only be a single `app_name`? If so, we should refactor the code to check and enforce this, ideally via the Rust type system (which you can do by making `app_name` immutable, that way it can only be defined once).

src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+    let mut ipa_archive = zip::ZipArchive::new(cursor)?;
+
+    let mut app_name = String::new();
+
+    // Extract .app from Payload/ directory
+    for i in 0..ipa_archive.len() {
+        let mut file = ipa_archive.by_index(i)?;
+        let file_path = file.name().to_string();
+
+        if let Some(stripped) = file_path.strip_prefix("Payload/") {
```
szokeasaurusrex: There is absolutely no need to call `to_string` here. 

Also [nit], I would inline this variable since it is only used once.

```suggestion

        if let Some(stripped) = file.name().strip_prefix("Payload/") {
```

src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
 
     debug!("File format validation failed");
     Err(anyhow!(
-        "File is not a recognized mobile app format (APK, AAB, or XCArchive): {}",
+        "File is not a recognized mobile app format (APK, AAB, XCArchive, or IPA): {}",
         path.display()
     ))
 }
 
+fn ipa_to_xcarchive(ipa_path: &Path, ipa_bytes: &[u8]) -> Result<TempFile> {
```
szokeasaurusrex: It would be extremely helpful for future devs working on Sentry CLI (and also for me while reviewing this change) if you could please add a detailed documentation comment for this method which explains exactly what this function does. 

Specifically, I am neither familiar with the IPA nor the Xarchive formats, so it would be helpful if you could describe both here, or link to a resource which describes this distinction. Then, also please add a description of how we transform the IPA into an Xarchive

src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+    // Extract IPA file
+    let cursor = std::io::Cursor::new(ipa_bytes);
+    let mut ipa_archive = zip::ZipArchive::new(cursor)?;
+
+    let mut app_name = String::new();
+
+    // Extract .app from Payload/ directory
+    for i in 0..ipa_archive.len() {
+        let mut file = ipa_archive.by_index(i)?;
+        let file_path = file.name().to_string();
```
szokeasaurusrex: [change request]

Using `ZipFile::name` directly [is dangerous](https://docs.rs/zip/latest/zip/read/struct.ZipFile.html#warnings), as it could break out of the current directory (e.g. if the path is `Payload/../../../something`). We could then end up overwriting this path later.

Please use [`ZipFile::enclosed_name`](https://docs.rs/zip/latest/zip/read/struct.ZipFile.html#method.enclosed_name) instead, as the enclosed name is validated to ensure it does not break out of the current directory.

szokeasaurusrex: As a side note, we should probably make sure we don't use `ZipFile::name` anywhere else in the codebase.

src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+            }
+
+            if !file.is_dir() {
+                // Create the file path in the XCArchive structure
+                let target_path = applications_dir.join(stripped);
+
+                // Create parent directories
+                if let Some(parent) = target_path.parent() {
+                    std::fs::create_dir_all(parent)?;
+                }
```
szokeasaurusrex: Just checking, as it is unclear from the code comment: it is also okay if there are not any parent directories to create here, right?

src/commands/mobile_app/upload.rs
```diff
@@ -198,19 +198,106 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
 // For APK and AAB files, we'll copy them directly into the zip
+// For IPA files, we'll convert them to XCArchive structure first
 fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
     debug!("Creating normalized zip for file: {}", path.display());
 
+    // Check if this is an IPA file that needs conversion
+    if is_zip_file(bytes) && is_ipa_file(bytes)? {
+        debug!("Converting IPA file to XCArchive structure");
+        return ipa_to_xcarchive(path, bytes);
+    }
```
szokeasaurusrex: I think it would make more sense to leave this function unchanged, and instead check whether we are dealing with an IPA file prior to calling `normalize_file`

src/commands/mobile_app/upload.rs
szokeasaurusrex: Also, I think the `command/mobile_app/upload.rs` file is starting to contain more code than it should. We are not super good about this elsewhere in the codebase, but imo we should keep the code in the command very simple (just define the arguments and `execute` function); stuff like this function probably belongs under `utils`

szokeasaurusrex (changes requested):
Requesting changes primarily because of the use of `ZipFile::name`, which is potentially dangerous here.

I've left some other suggestions, too

src/utils/mobile_app/validation.rs
```diff
@@ -37,6 +37,32 @@ pub fn is_aab_file(bytes: &[u8]) -> Result<bool> {
...
+
+            // Check if there's a .app directory in Payload/
+            if name.starts_with("Payload/") && name.ends_with(".app/") {
+                has_app_in_payload = true;
+                break;
+            }
+        }
+    }
+
+    Ok(has_payload && has_app_in_payload)
```
szokeasaurusrex: Much simpler:

```suggestion
    let is_ipa = archive
        .file_names()
        .any(|name| name.starts_with("Payload/") && name.ends_with(".app/"));

    Ok(is_ipa)
```

szokeasaurusrex: Also [question]: Does the `.app` need to be immediatly under the `Payload` in order to be a valid IPA? As in `Payload/my_app_name.app`? Or is it okay to have `Payload/some_other_directory/my_app_name.app`?

Currently, the code (both your version and mine) would consider something to be an IPA if it only has a path like `Payload/some_other_directory/my_app_name.app` but not one like `Payload/my_app_name.app`.

tests/integration/mobile_app/upload.rs
```diff
@@ -55,6 +55,18 @@ fn command_mobile_app_upload_invalid_xcarchive() {
         .run_and_assert(AssertCommand::Failure);
 }
 
+#[test]
+fn command_mobile_app_upload_invalid_ipa() {
```
szokeasaurusrex, line 66: Let's please also test with a valid IPA file if possible

src/commands/mobile_app/upload.rs
noahsmartin: yep

src/commands/mobile_app/upload.rs
noahsmartin: 👍 gave this to cursor and it came up with a pretty detailed comment

noahsmartin:
Thanks @szokeasaurusrex just updated the PR with those changes

src/commands/mobile_app/upload.rs
```diff
@@ -36,7 +38,7 @@ pub fn make_command(command: Command) -> Command {
         .arg(
             Arg::new("paths")
                 .value_name("PATH")
-                .help("The path to the mobile app files to upload. Supported files include Apk, Aab or XCArchive.")
+                .help("The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive, or IPA.")
```
szokeasaurusrex:
```suggestion
                .help("The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive, and IPA.")
```

src/commands/mobile_app/upload.rs
```diff
@@ -95,12 +97,29 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                let xcarchive_path =
+                    ipa_to_xcarchive(path, &byteview, &temp_dir).with_context(|| {
+                        format!(
+                            "Failed to convert IPA to XCArchive for file {}",
+                            path.display()
+                        )
+                    })?;
+                normalize_directory(&xcarchive_path).with_context(|| {
+                    format!("Failed to normalize XCArchive for file {}", path.display())
+                })?
```
szokeasaurusrex: I think it is a lot cleaner to combine these lines. Imo it is okay if the error message is the same for both.

```suggestion
                ipa_to_xcarchive(path, &byteview, &temp_dir)
                    .and_then(|path| normalize_directory(&path))
                    .with_context(|| format!("Failed to process IPA file {}", path.display()))?
```

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+/// ```
+///
+/// # Transformation Process
+///
+/// This function performs the following steps:
+/// 1. Creates the XCArchive directory structure (`archive.xcarchive/Products/Applications/`)
+/// 2. Extracts the app name from the IPA by finding the shortest path ending with `.app/Info.plist`
+/// 3. Extracts all files from the IPA's `Payload/` directory into the XCArchive structure
+/// 4. Creates an `Info.plist` file for the XCArchive with the app path reference
+/// 5. Returns the path to the XCArchive directory structure
```
szokeasaurusrex: Thanks for adding the comment! Imo, the "Transformation Process" part is unnecessary, as one can simply read the code to understand this. 

The information about the format is more important to be able to understand this code's purpose.

```suggestion
```

src/commands/mobile_app/upload.rs
```diff
@@ -95,12 +97,29 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-                format!(
-                    "Failed to generate uploadable bundle for file {}",
-                    path.display()
-                )
-            })?
+
+            // Handle IPA files by converting them to XCArchive
+            if is_zip_file(&byteview) && is_ipa_file(&byteview)? {
+                debug!("Converting IPA file to XCArchive structure");
+                let temp_dir = crate::utils::fs::TempDir::create()?;
```
szokeasaurusrex: [nit] I think it is more readable if we import `TempDir` from `crate::utils::fs`. We already import `TempFile` from there
```suggestion
                let temp_dir = TempDir::create()?;
```

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+    let mut ipa_archive = ZipArchive::new(cursor)?;
+
+    let app_name = extract_app_name_from_ipa(&ipa_archive)?;
+
+    // Extract all files from the archive
+    for i in 0..ipa_archive.len() {
+        let mut file = ipa_archive.by_index(i)?;
+
+        if let Some(name) = file.enclosed_name() {
+            if let Ok(stripped) = name.strip_prefix("Payload/") {
```
szokeasaurusrex: [rust fyi] These can be combined into a single `if let` like so:

```suggestion
        if let Some(stripped) = file
            .enclosed_name()
            .and_then(|name| name.strip_prefix("Payload/").ok())
        {
```

Although I am also fine with keeping this as is

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+        "Created XCArchive Info.plist at: {}",
+        info_plist_path.display()
+    );
+    Ok(xcarchive_dir)
+}
+
+fn extract_app_name_from_ipa(archive: &ZipArchive<Cursor<&[u8]>>) -> Result<String> {
+    archive
+        .file_names()
+        .filter(|name| name.starts_with("Payload/") && name.ends_with(".app/Info.plist"))
```
szokeasaurusrex: Have you tested that this works as expected if running this command on Windows? Since on Windows, the paths might have `\` instead of `/` (though I am not sure how the `zip` crate handles this)

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+    std::fs::write(&info_plist_path, info_plist_content)?;
+
+    debug!(
+        "Created XCArchive Info.plist at: {}",
+        info_plist_path.display()
+    );
+    Ok(xcarchive_dir)
+}
+
+fn extract_app_name_from_ipa(archive: &ZipArchive<Cursor<&[u8]>>) -> Result<String> {
```
szokeasaurusrex: I am still waiting for an answer to [this question](https://github.com/getsentry/sentry-cli/pull/2619#discussion_r2212969623) from my previous review

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+    archive
+        .file_names()
+        .filter(|name| name.starts_with("Payload/") && name.ends_with(".app/Info.plist"))
+        .min_by_key(|name| name.len())
```
szokeasaurusrex: Why do we take the shortest-length name here?

src/utils/mobile_app/ipa.rs
```diff
@@ -0,0 +1,132 @@
...
+        .and_then(|name| name.strip_prefix("Payload/"))
+        .and_then(|name| name.split('/').next())
+        .and_then(|name| name.strip_suffix(".app"))
+        .map(|name| name.to_owned())
+        .ok_or_else(|| anyhow!("No .app found in IPA"))
+}
```
szokeasaurusrex: No need to clone the string inside this function.

However, you will need to adjust where the funciton is called with a `.to_owned` in order to satisfy the borrow checker. Main benefit of moving the clone to the call site is that if we were ever to use this function somewhere else, where the clone is not needed, we would avoid the clone. 

```suggestion
fn extract_app_name_from_ipa<'a>(archive: &'a ZipArchive<Cursor<&[u8]>>) -> Result<&'a str> {
    archive
        .file_names()
        .filter(|name| name.starts_with("Payload/") && name.ends_with(".app/Info.plist"))
        .min_by_key(|name| name.len())
        .and_then(|name| name.strip_prefix("Payload/"))
        .and_then(|name| name.split('/').next())
        .and_then(|name| name.strip_suffix(".app"))
        .ok_or_else(|| anyhow!("No .app found in IPA"))
}
```

src/utils/mobile_app/validation.rs
```diff
@@ -37,6 +37,19 @@ pub fn is_aab_file(bytes: &[u8]) -> Result<bool> {
     Ok(has_bundle_config && has_base_manifest)
 }
 
+pub fn is_ipa_file(bytes: &[u8]) -> Result<bool> {
```
szokeasaurusrex, line 41: Just wondering, why do we have this validation code in the first place? Why not just try to parse the file as all the different file types directly, then if we fail to, we assume that the file is invalid?

Having the separate validation logic seems like it is leading to duplicated code paths. We already check this structure in `ipa.rs`, in the `extract_app_name_from_ipa` method.

Fixing this (especially for the other file types) is of course out of scope for this PR, but hoping you can shed some light on this architectural decision.

src/utils/mobile_app/ipa.rs
szokeasaurusrex: Actually, this is a bug: https://github.com/getsentry/sentry-cli/pull/2629

szokeasaurusrex: Why not just check against a regex to extract the correct path? E.g. something like `Payload/[^/]+\.app/Info\.plist` could work I think

Also, how would we want to handle the case if there are multiple `.app` files nested directly under the `Payload`?

tests/integration/mobile_app/upload.rs
```diff
@@ -160,3 +172,57 @@ fn command_mobile_app_upload_apk_chunked() {
         .register_trycmd_test("mobile_app/mobile_app-upload-apk.trycmd")
         .with_default_token();
 }
+
+#[test]
+#[cfg(target_os = "macos")]
```
szokeasaurusrex, line 184: Why only run the test on macOS? The code we are testing is available on all platforms from what I can tell.

```suggestion
```

src/utils/mobile_app/ipa.rs
noahsmartin: It wouldn't really make sense to upload an IPA from windows, they are always on Macs and other things like the asset catalog support would not work on windows. So I will just add the code to avoid ever calling this path so users don't get confused about why thinks like asset catalogs stop working

src/utils/mobile_app/ipa.rs
noahsmartin: Ended up changing this back to using String when I updated the function to use regex

src/utils/mobile_app/validation.rs
noahsmartin: Discussed on a call, it's mostly because this was ported from our typescript code, we probably could refactor this and make it at least just file extension checking

tests/integration/mobile_app/upload.rs
noahsmartin: I updated the ipa support to be macOS only

noahsmartin:
Thanks @szokeasaurusrex just made some more updates. I also made the IPA support macOS only so people don't accidentally upload without the asset catalog logic running (the Swift code). I intend to make XCArchive support macOS only too but figured I would do that in a follow up PR to keep this one focused.

src/utils/mobile_app/mod.rs
```diff
@@ -2,8 +2,12 @@
 
 #[cfg(target_os = "macos")]
 mod apple;
+#[cfg(target_os = "macos")]
+mod ipa;
```
szokeasaurusrex: You forgot to delete these lines when you got rid of the `ipa` module. This is causing CI to fail.

```suggestion
```

src/commands/mobile_app/upload.rs
```diff
@@ -21,22 +21,28 @@ use crate::config::Config;
...
+use crate::utils::mobile_app::{handle_asset_catalogs, ipa_to_xcarchive, is_ipa_file};
 use crate::utils::mobile_app::{is_aab_file, is_apk_file, is_apple_app, is_zip_file};
 use crate::utils::progress::ProgressBar;
 use crate::utils::vcs;
 
 pub fn make_command(command: Command) -> Command {
+    #[cfg(target_os = "macos")]
+    let help_text = "The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive, and IPA.";
+    #[cfg(not(target_os = "macos"))]
+    let help_text = "The path to the mobile app files to upload. Supported files include Apk, Aab, and XCArchive.";
```
szokeasaurusrex: I'd define this as a constant – I think it is more idiomatic in this case

```suggestion
    #[cfg(target_os = "macos")]
    const HELP_TEXT: &str = "The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive, and IPA.";
    #[cfg(not(target_os = "macos"))]
    const HELP_TEXT: &str = "The path to the mobile app files to upload. Supported files include Apk, Aab, and XCArchive.";
```

src/commands/mobile_app/upload.rs
```diff
@@ -21,22 +21,28 @@ use crate::config::Config;
...
     command
         .about("[EXPERIMENTAL] Upload mobile app files to a project.")
         .org_arg()
         .project_arg(false)
         .arg(
             Arg::new("paths")
                 .value_name("PATH")
-                .help("The path to the mobile app files to upload. Supported files include Apk, Aab or XCArchive.")
+                .help(help_text)
```
szokeasaurusrex:
```suggestion
                .help(HELP_TEXT)
```

src/commands/mobile_app/upload.rs
```diff
@@ -186,9 +206,9 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
         return Ok(());
     }
 
-    // Check if the file is a zip file (then AAB or APK)
+    // Check if the file is a zip file (then AAB, APK, or IPA)
     if is_zip_file(bytes) {
-        debug!("File is a zip, checking for AAB/APK format");
+        debug!("File is a zip, checking for AAB/APK/IPA format");
```
szokeasaurusrex, line 212: I would consider updating this debug statement to print only AAB and APK on non-macOS platforms. Not strictly necessary though, imo, as this is just a debug message

src/commands/mobile_app/upload.rs
```diff
@@ -198,11 +218,23 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+            debug!("Detected IPA file");
+            return Ok(());
+        }
     }
 
     debug!("File format validation failed");
+    #[cfg(target_os = "macos")]
+    let format_list = "APK, AAB, XCArchive, or IPA";
+    #[cfg(not(target_os = "macos"))]
+    let format_list = "APK, AAB, or XCArchive";
```
szokeasaurusrex, line 237: As we have to conditionally print this list in several places, it might eventually make sense to extract this into a constant. I don't think we have to do it now necessarily but could make the code more easily maintainable, especially if we add more file formats in the future

src/commands/mobile_app/upload.rs
```diff
@@ -198,11 +218,23 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
...
+
     Err(anyhow!(
-        "File is not a recognized mobile app format (APK, AAB, or XCArchive): {}",
+        "File is not a recognized mobile app format ({}): {}",
+        format_list,
```
szokeasaurusrex: This can also be written like so – it is more readable imo. There is a lint which can enforce this, so I'll enable it


```suggestion
        "File is not a recognized mobile app format ({format_list}): {}",
```

src/utils/mobile_app/apple.rs
```diff
@@ -27,3 +33,121 @@ fn find_car_files(root: &Path) -> Vec<PathBuf> {
...
+
+    debug!(
+        "Created XCArchive Info.plist at: {}",
+        info_plist_path.display()
+    );
+    Ok(xcarchive_dir)
+}
+
+fn extract_app_name_from_ipa(archive: &ZipArchive<Cursor<&[u8]>>) -> Result<String> {
+    let pattern = Regex::new(r"^Payload/([^/]+)\.app/Info\.plist$")?;
```
szokeasaurusrex: As compiling regex is an expensive-ish operation, it is better to only do it once. We can achieve this by storing the pattern in a constant `LazyLock`, which is initialized on first access, like so:

```suggestion
    const PATTERN: LazyLock<Regex> = LazyLock::new(|| {
        Regex::new(r"^Payload/([^/]+)\.app/Info\.plist$").expect("regex is valid")
    });
```

You will need to update the usages of the variable, and also add an import of `LazyLock`

src/utils/mobile_app/apple.rs
```diff
@@ -27,3 +33,121 @@ fn find_car_files(root: &Path) -> Vec<PathBuf> {
...
+        if let Some(captures) = pattern.captures(name) {
+            matches.push(captures[1].to_string());
+        }
+    }
+
+    match matches.len() {
+        1 => Ok(matches[0].clone()),
+        _ => Err(anyhow!("IPA file did not contain exactly one .app.")),
+    }
+}
```
szokeasaurusrex: I would refactor your for loop using an iterator as follows (I think iterators are more idiomatic in Rust for such operations).

This also enables you to return a str reference rather than an owned String.

```suggestion
fn extract_app_name_from_ipa<'a>(archive: &'a ZipArchive<Cursor<&[u8]>>) -> Result<&'a str> {
    let pattern = Regex::new(r"^Payload/([^/]+)\.app/Info\.plist$")?;
    let matches = archive
        .file_names()
        .filter_map(|name| pattern.captures(name))
        .map(|c| c.get(1).expect("group 1 must be present").as_str())
        .take(2)  // If there are ≥2 matches, we already know the IPA is invalid
        .collect::<Vec<_>>();

    if let &[app_name] = matches.as_slice() {
        Ok(app_name)
    } else {
        Err(anyhow!("IPA did not contain exactly one .app."))
    }
}
```

src/utils/mobile_app/validation.rs
```diff
@@ -37,6 +37,20 @@ pub fn is_aab_file(bytes: &[u8]) -> Result<bool> {
...
+#[cfg(target_os = "macos")]
+pub fn is_ipa_file(bytes: &[u8]) -> Result<bool> {
+    let cursor = std::io::Cursor::new(bytes);
+    let archive = zip::ZipArchive::new(cursor)?;
+
+    let is_ipa = archive.file_names().any(|name| {
+        name.starts_with("Payload/")
+            && name.ends_with(".app/Info.plist")
+            && name.matches('/').count() == 2
+    });
```
szokeasaurusrex, line 49: we could reuse the regex here, but this is also fine

tests/integration/_cases/mobile_app/mobile_app-upload-help.trycmd
szokeasaurusrex, line 1: Please rename this file to indicate this is macOS specific

tests/integration/mobile_app/upload.rs
szokeasaurusrex: [question] do we expect that in the future users may want to be able to upload IPA files from a non-macOS platform, e.g. from a (non-macOS) CI runner? If so, would it be possible to add such support?

szokeasaurusrex (approved):
Hey, this looks good! I left some suggestions, please address those, and then feel free to merge

noahsmartin:
@szokeasaurusrex we wouldn't be able to run the asset catalog logic on non-macos which would result in incomplete features when the rest of our backend runs on the upload. I'd prefer to fail early here and require that the upload be on a Mac rather than possible show incorrect results later on, that's why I don't think we'd add non-macos support for these

PR #2621, opened 2025-07-16
ref: Reformat `src/commands`

szokeasaurusrex:
Previously, `cargo fmt` was unable to find these files, so CI could not enforce formatting. Fortunately, seems as though most contributors have been using rust-analyzer locally, so there are only a few formatter violations to fix here.

Prerequisite for #2622


loewenheim (approved):
PR itself looks good, but the title is a bit off—no file named `commands.rs` exists in the repo.

szokeasaurusrex:
Yeah, i meant just `src/commands`. Thanks for catching

PR #2623, opened 2025-07-16
ref: Clean up code

szokeasaurusrex:
It is a bit weird to have this block defining two variables inside the definition of another variable.
    
I think the following is a bit more readable. It also keeps the code DRY.

Review alongside #2618, this is basically the changes I would have suggested on that PR


loewenheim (approved):
I don't mind the variables in the block, but your version is still better even so.

Issue #2617, opened 2025-07-15
Use `typescript` to generate type declarations


Lms24:
### Description

Right now, we ship hand-crafted type declaration files for the JS interface. Since we added typescript now to type-check the JS codebase (#2611), we can probably also use it to generate type declaration files (details to be figured out). This would ensure that the manual declaration files don't diverge any more from our actual exports (which they do as tracked in #2606)

PR #2627, opened 2025-07-18
feat(js): Export `live` option type for `releases.uploadSourceMaps`

Lms24:
While working on #2605, I missed that we also need to add the new `live` option to the manually created *public* type declaration of the `SentryCLI().releases.uploadSourceMaps` method. This PR now adds the `live` option to this method. (feat because although there's no behaviour change, from a user perspective #2605 is only public with this change)

Very good reason to tackle #2617, since it will eliminate this kind of divergence. I gave it a quick try and got results pretty quickly. However, due to our internal typing currently being a lot more lenient than what the type declarations specify, I'd consider unifying the typing a breaking change. The goal is that our exported type declarations come from out internal type definitions BUT to get to this point, we first need to change these internal types to adhere to the currently exported declarations. 
@szokeasaurusrex let me know when you want to tackle v3, then I'll take another look at this.

### DIFF
--- a/js/releases/index.js
+++ b/js/releases/index.js
@@ -149,6 +149,7 @@ class Releases {
    *   ext: ['js', 'map', 'jsbundle', 'bundle'],  // override file extensions to scan for
    *   projects: ['node'],        // provide a list of projects
    *   decompress: false          // decompress gzip files before uploading
+   *   live: undefined            // whether to inherit stdio to display `sentry-cli` output directly.
    * });
    *
    * @param {string} release Unique name of the release.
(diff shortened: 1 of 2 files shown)


js/releases/index.js
```diff
@@ -149,6 +149,7 @@ class Releases {
    *   ext: ['js', 'map', 'jsbundle', 'bundle'],  // override file extensions to scan for
    *   projects: ['node'],        // provide a list of projects
    *   decompress: false          // decompress gzip files before uploading
+   *   live: undefined            // whether to inherit stdio to display `sentry-cli` output directly.
```
szokeasaurusrex: Why is live `undefined`? Seems wrong but maybe this is just some JS thing I don't understand 🫣

Lms24: Good catch, thanks! Forgot that the default value is actually `true` 🤦‍♂️

PR #2628, opened 2025-07-18
feat(js): Expose `rejectOnError` in public `execute` type declarations

Lms24:
Of course I missed more occurances of the `live` mode in type declarations 🤦‍♂️ 
(no urgency this time though)

This PR now adds the new `rejectOnError` to the remaining `.execute` type declarations so that this mode can also be set by library consumers directly calling `.execute` (e.g. `@sentry/react-router` SDK).

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -217,7 +217,7 @@ declare module '@sentry/cli' {
 
     newDeploy(release: string, options: SentryCliNewDeployOptions): Promise<string>;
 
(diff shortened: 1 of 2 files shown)


szokeasaurusrex:
@Lms24 is this actually a `feat` though? ~~Because the code itself is the same, these are only code comment, right?~~

Oh wait, there is some `ts` stuff

Lms24:
yes, it's as much a feat as #2627 was one. thx for the review!

PR #2632, opened 2025-07-21
feat(proguard): Use chunk uploading by default

### LINKED ISSUE #2328: Make chunk uploads the default for Proguard files
Chunk uploads are available for Proguard files, but they are not the default and have been labelled as experimental. Before we can make them the default, we must do the following:

- [ ] Check whether the server supports Proguard chunked uploads, and only use chunk uploading if yes
- [ ] Investigate and resolve the large memory usage of chunked uploads (#2291)

szokeasaurusrex:
Enable chunk uploads for Proguard files. This upload method avoids 502 errors when uploading large Proguard files, and enables interrupted uploads to be resumed.

Closes #2328

### DIFF
--- a/src/api/data_types/chunking/upload/capability.rs
+++ b/src/api/data_types/chunking/upload/capability.rs
@@ -33,6 +33,9 @@ pub enum ChunkUploadCapability {
     /// Upload of preprod artifacts
     PreprodArtifacts,
 
+    /// Upload of ProGuard mappings
+    Proguard,
+
     /// Any other unsupported capability (ignored)
     Unknown,
 }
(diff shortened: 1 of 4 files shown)


src/api/data_types/chunking/upload/capability.rs
```diff
@@ -53,6 +56,7 @@ impl<'de> Deserialize<'de> for ChunkUploadCapability {
             "bcsymbolmaps" => ChunkUploadCapability::BcSymbolmap,
             "il2cpp" => ChunkUploadCapability::Il2Cpp,
             "preprod_artifacts" => ChunkUploadCapability::PreprodArtifacts,
+            "proguard" => ChunkUploadCapability::Proguard,
```
szokeasaurusrex, line 59: The server already returns this since https://github.com/getsentry/sentry/commit/e0b7ff2fb72e49bfe862c6b90e42257990ea38e1

szokeasaurusrex:
We should release this in the next major, as we will want to remove release association from the CLI before swapping over to making chunk uploading the default. Otherwise, we would always have to wait for assembly to complete, which could arguably be a degradation compared to the current default non-chunked behavior which does not have this restriction

PR #2641, opened 2025-07-22
build: Delete `Makefile`

szokeasaurusrex:
The `Makefile` was added in https://github.com/getsentry/sentry-cli/commit/53a4f3b74a33711a8727c25cd425769f28344fcf, and at that time, it was called from CI. Since then, our CI no longer calls `make` to build, lint, and test Sentry CLI, so the `Makefile` no longer appears to serve much of a purpose.

So, let's delete the file to avoid confusing anyone as to its purpose.


szokeasaurusrex:
I also ran the release build workflow against this PR and it passes, indicating that we don't seem to need the `Makefile` in the release process, either

PR #2645, opened 2025-07-22
fix: Remove apple uploads from intel

noahsmartin:
Adding a test runner for intel macOS, also fixes the issue with intel macOS and the mobile app command by disabling the iOS features on intel

### DIFF
--- a/src/commands/mobile_app/upload.rs
+++ b/src/commands/mobile_app/upload.rs
@@ -21,19 +21,19 @@ use crate::config::Config;
 use crate::utils::args::ArgExt as _;
 use crate::utils::chunks::{upload_chunks, Chunk, ASSEMBLE_POLL_INTERVAL};
 use crate::utils::fs::get_sha1_checksums;
-#[cfg(target_os = "macos")]
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use crate::utils::fs::TempDir;
 use crate::utils::fs::TempFile;
-#[cfg(target_os = "macos")]
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use crate::utils::mobile_app::{handle_asset_catalogs, ipa_to_xcarchive, is_ipa_file};
 use crate::utils::mobile_app::{is_aab_file, is_apk_file, is_apple_app, is_zip_file};
 use crate::utils::progress::ProgressBar;
 use crate::utils::vcs;
 
 pub fn make_command(command: Command) -> Command {
-    #[cfg(target_os = "macos")]
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
     const HELP_TEXT: &str = "The path to the mobile app files to upload. Supported files include Apk, Aab, XCArchive, and IPA.";
-    #[cfg(not(target_os = "macos"))]
+    #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
     const HELP_TEXT: &str = "The path to the mobile app files to upload. Supported files include Apk, Aab, and XCArchive.";
     command
         .about("[EXPERIMENTAL] Upload mobile app files to a project.")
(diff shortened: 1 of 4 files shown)


apple-catalog-parsing/build.rs
```diff
@@ -94,4 +94,8 @@ fn main() {
     println!(
         "cargo:rustc-link-search={developer_dir_path}/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx"
     );
+
+    let lib_dir = "/usr/lib/swift";
+    println!("cargo:rustc-link-arg=-Wl,-rpath");
+    println!("cargo:rustc-link-arg=-Wl,{lib_dir}");
```
szokeasaurusrex: This would no longer be necessary if we restrict to ARM, correct?

.github/workflows/test.yml
```diff
@@ -41,3 +41,75 @@ jobs:
...
+          name: sentry-cli-macos14-build${{ matrix.feature-suffix }}-${{ matrix.base-os }}
+          path: ./bin
+
+      - name: Make Binary Executable
+        run: chmod +x ./bin/sentry-cli
+
+      - name: Test Binary Compatibility
+        run: |
+          # Ensure the cli runs
+          ./bin/sentry-cli --help
```
szokeasaurusrex: If we will make this feature ARM-specific, we should have test coverage for both the ARM-binary and the x86_64-binary; however, I would prefer that we try to keep it simpler than this (i.e. depending on fewer runner types, and perhaps only testing on the system we build for).

apple-catalog-parsing/build.rs
szokeasaurusrex: Or wait, this is the crate's build.rs, so I don't think these directives have any effect here

szokeasaurusrex (commented):
Left a couple comments, but this is not a full review, as the PR is still marked as draft

src/commands/mobile_app/upload.rs
```diff
@@ -208,9 +208,9 @@ fn validate_is_mobile_app(path: &Path, bytes: &[u8]) -> Result<()> {
 
     // Check if the file is a zip file (then AAB, APK, or IPA)
     if is_zip_file(bytes) {
-        #[cfg(target_os = "macos")]
+        #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
         debug!("File is a zip, checking for AAB/APK/IPA format");
-        #[cfg(not(target_os = "macos"))]
+        #[cfg(any(not(target_os = "macos"), not(target_arch = "aarch64")))]
```
szokeasaurusrex: Would prefer we flip this `any(not(), not())` to `not(all())`, as the latter is more clearly the inverse of the `cfg(all())` above. `not(all())` is also more concise, and we appear to be using it elsewhere

```suggestion
        #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
```

.github/workflows/test.yml
```diff
@@ -15,13 +15,15 @@ jobs:
...
-        os: [ubuntu-24.04, macos-14, windows-2022]
+        os: [ubuntu-24.04, macos-14, windows-2022, macos-14-large]
         feature-args: ['', '-Funstable-mobile-app']
         include:
           - os: ubuntu-24.04
             display-os: Linux
           - os: macos-14
             display-os: macOS
+          - os: macos-14-large
+            display-os: macOS (Intel)
```
szokeasaurusrex: As discussed in Slack, I would rather we test the Intel binary on the `macos-14` runner to avoid depending on additional runner images.

I suggest doing this like so: https://github.com/getsentry/sentry-cli/pull/2658 (this PR also adds linting for the Intel binary, which I think is also important to do).

szokeasaurusrex:
Looks good; however, CI is failing on #2658 (which this PR is depending on), since that PR needs this PR in order to pass. So, we will need to reorder the two PRs to the following:

```
master <- this PR <- #2658
```

Would you like to take care of this reordering or shall I?

PR #2650, opened 2025-07-24
lint all together

szokeasaurusrex:


szokeasaurusrex:
I don't think this is needed

PR #2652, opened 2025-07-25
meta: Enforce Sentry commit message conventions

szokeasaurusrex:
Co-authored-by: daniel.szoke <daniel.szoke@sentry.io>


szokeasaurusrex:
After discussion with @dingsdax, perhaps enforcing the commit message convention on each commit is overkill, as we typically squash and merge anyways.

We plan to introduce a new PR template, which would instruct contributors to follow the commit naming convention in the PR title, let's see if that solves the problem before reconsidering automated enforcement.

Issue #2661, opened 2025-07-28
Expose logs and error (events) via `sentry-cli`


AbhiPrasad:
### Description

Add a `sentry-cli logs` command that allows you to manage logs from the SDK.

Add support for live tailing of the logs and issues via `--live` argument on the `list` subcommand.

This is useful to hook into AI, but in certain environments (mobile, gaming) it also can provide useful visibility without needing to go into the Sentry UI.

We can also explore spans, but emitting full traces might be difficult.

szokeasaurusrex:
Hey @AbhiPrasad – I would like to clarify this wording:

> ... that allows you to _live-tail_ logs or error events from the SDK

What do you mean by live-tailing? Would the command essentially just print out logs and errors as they come in – meaning that the CLI would not exit, just continue polling the server for new logs and errors and print these to `stdout` as they come in?

If so, perhaps this feature is better implemented as an argument (perhaps named something like `--live`) to our already-exising `sentry-cli issues list` command, which as the name suggests, lists your issues. And, to get logs, we could add a similar command.

vgrozdanic:
I agree with @szokeasaurusrex, the `tail` has originated from the Slack thread, but given how we already do things in our CLI, just adding a new command for `logs` and `--live` argument makes the most sense IMO, especially since we don't have the real "tail" functionality, but we would just poll to show newer events.

That being said, i can take this issue over, unless one of you (or someone else) wants it :) 

Tomorrow i am planning to work on it.

AbhiPrasad:
I really like the idea of a `--live` argument, feel free to change my issue description to match that.

As for working on it, please go for it @vgrozdanic!

szokeasaurusrex:
Sure, you can try it @vgrozdanic!

I am assuming this would be your first contribution to Sentry CLI? If it is, please make sure you have `rust-analyzer` installed in your editor – it makes Rust development way easier – and also let me know if you need any assistance with anything, as I am happy to help! The contributing docs are somewhat out of date, but I recently added Cursor rules which you can reference.

AbhiPrasad:
I also realized, this could be a interesting way to validate things during onboarding.

For example if you install the sdk with the `sentry-wizard` and send test errors, we could automatically run `sentry-cli error --live` or `sentry-cli logs --live` in your terminal and validate if things were sent.

szokeasaurusrex:
Although, I don't think the `--live` argument is necessary for this. You could also just use the existing commands probably

[...]

PR #2664, opened 2025-07-30
feat(logs): Introduce logs command

vgrozdanic:
Add a new `logs` subcommand to the Sentry CLI, allowing users to manage and query logs from their organizations. 

This includes a `list` command to fetch and display log entries with options for filtering and pagination. 

Implemented common arguments for logs and integrated them into the command structure. Also added integration tests for the new functionality.

Part of https://github.com/getsentry/sentry-cli/issues/2661

Live tailing will be done as a part of separate PR to make this a bit easier to review, it's already very large change.


src/api/mod.rs
```diff
@@ -1226,6 +1226,58 @@ impl<'a> AuthenticatedApi<'a> {
         Ok(rv)
     }
 
+    /// Fetch organization events from the specified dataset
+    pub fn fetch_organization_events(
```
szokeasaurusrex, line 1230: Linter is already complaining about this, but I also noticed this seems like too many function arguments, and I would therefore suggest combining them into a struct

vgrozdanic: Yeah, makes sense, i also had a feeling like it's a lot of args, but also i wanted to make this function as flexible as possible for future usage

src/api/mod.rs
```diff
@@ -1225,6 +1225,75 @@ impl<'a> AuthenticatedApi<'a> {
...
+            params.push(format!("query={}", QueryArg(query)));
+        }
+
+        params.push(format!("per_page={}", options.per_page.unwrap_or(100)));
+        params.push(format!(
+            "statsPeriod={}",
+            options.stats_period.unwrap_or("1h")
+        ));
+        params.push("referrer=sentry-cli-tail".to_owned());
+        params.push(format!("sort={}", options.sort.unwrap_or("-timestamp")));
```
szokeasaurusrex: [suggestion] I would suggest we extract all of this logic into a method on the `FetchEventsOptions` struct, the main reason being, that we are mostly just processing that struct into a query string here, so it makes more sense to have that logic on that struct rather than on `AuthenticatedApi`

Also, using chained iterators is perhaps more idiomatic in Rust than constructing a vector this way, but this is also fine

src/api/mod.rs
```diff
@@ -1225,6 +1225,75 @@ impl<'a> AuthenticatedApi<'a> {
...
+    pub query: Option<&'a str>,
+    /// Number of events per page (default: 100)
+    pub per_page: Option<usize>,
+    /// Time period for stats (default: "1h")
+    pub stats_period: Option<&'a str>,
+    /// Sort order (default: "-timestamp")
+    pub sort: Option<&'a str>,
+}
+
+impl<'a> AuthenticatedApi<'a> {
```
szokeasaurusrex: [change request] There is already an existing `impl<'a> AuthenticatedApi<'a>` block above, please combine this with that block

Sadly, [Clippy cannot check for this](https://github.com/rust-lang/rust-clippy/issues/8714) :/

src/api/mod.rs
```diff
@@ -1225,6 +1225,75 @@ impl<'a> AuthenticatedApi<'a> {
...
+    /// Fetch organization events from the specified dataset
+    pub fn fetch_organization_events(
+        &self,
+        org: &str,
+        dataset: &str,
+        fields: &[&str],
+        options: FetchEventsOptions,
```
szokeasaurusrex: Pass by reference would make the API more flexible here, as we don't need to take ownership of the struct.

```suggestion
        options: &FetchEventsOptions,
```

src/commands/logs/common_args.rs
```diff
@@ -0,0 +1,13 @@
+use clap::Args;
+
+/// Common arguments for all logs subcommands.
+#[derive(Args)]
```
szokeasaurusrex: Nice 🚀

src/commands/logs/list.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[arg(help = "Maximum number of rows to print.")]
+    pub(super) max_rows: Option<usize>,
+
+    #[arg(long = "per-page", default_value = "100")]
+    #[arg(help = "Number of log entries per request (max 1000).")]
+    pub(super) per_page: usize,
+
+    #[arg(long = "query", default_value = "")]
+    #[arg(help = "Query to filter logs. Example: \"level:error\"")]
+    pub(super) query: String,
```
szokeasaurusrex:
```suggestion
    #[command(flatten)]
    common: CommonLogsArgs,

    #[arg(long = "max-rows")]
    #[arg(help = "Maximum number of rows to print.")]
    max_rows: Option<usize>,

    #[arg(long = "per-page", default_value = "100")]
    #[arg(help = "Number of log entries per request (max 1000).")]
    per_page: usize,

    #[arg(long = "query", default_value = "")]
    #[arg(help = "Query to filter logs. Example: \"level:error\"")]
    query: String,
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    #[arg(long = "live")]
+    #[arg(help = "Live-tail logs (not implemented yet).")]
+    pub(super) live: bool,
```
szokeasaurusrex: Since it is not implemented yet:

```suggestion
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+        Some(args.query.as_str())
+    };
+    let fields = [
+        "sentry.item_id",
+        "trace",
+        "severity",
+        "timestamp",
+        "message",
+    ];
```
szokeasaurusrex: Might make more sense to define this as a constant, preferably in the `fetch_organization_events` method, since we appear to always be calling it with the same value. We can always add this parameter back later if needed

src/api/mod.rs
```diff
@@ -1225,6 +1225,75 @@ impl<'a> AuthenticatedApi<'a> {
...
+    /// Sort order (default: "-timestamp")
+    pub sort: Option<&'a str>,
+}
+
+impl<'a> AuthenticatedApi<'a> {
+    /// Fetch organization events from the specified dataset
+    pub fn fetch_organization_events(
+        &self,
+        org: &str,
+        dataset: &str,
```
szokeasaurusrex: I would suggest defining an enum for the possible datasets, so we avoid passing invalid values here

src/commands/logs/list.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[command(flatten)]
+    pub(super) common: CommonLogsArgs,
+
+    #[arg(long = "max-rows")]
+    #[arg(help = "Maximum number of rows to print.")]
+    pub(super) max_rows: Option<usize>,
+
+    #[arg(long = "per-page", default_value = "100")]
+    #[arg(help = "Number of log entries per request (max 1000).")]
+    pub(super) per_page: usize,
```
szokeasaurusrex: This seems more like an internal implementation detail – why the decision to make it user-configurable?

Edit: if I am reading the code in the API correctly, it seems like this is the maximum number of logs that are returned from the server in the request. Since we only make one request, it would also then be the maximum number of logs which we display, right? If that is correct, please rename this argument to something like `max-logs`, and update the help to state this is the maximum number of logs we will retrieve. I would also combine this with the `max_rows` option in that case

src/commands/logs/list.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut table = Table::new();
+    table
+        .title_row()
+        .add("Item ID")
+        .add("Timestamp")
+        .add("Severity")
+        .add("Message")
+        .add("Trace");
+
+    let max_rows = std::cmp::min(logs.len(), args.max_rows.unwrap_or(usize::MAX));
```
szokeasaurusrex: `usize::MAX` seems like a really big default value. I would maybe reduce this?

src/commands/logs/common_args.rs
```diff
@@ -0,0 +1,13 @@
...
+pub(super) struct CommonLogsArgs {
+    #[arg(short = 'o', long = "org")]
+    #[arg(help = "The organization ID or slug.")]
+    pub(super) org: Option<String>,
+
+    #[arg(short = 'p', long = "project")]
+    #[arg(help = "The project ID or slug.")]
+    pub(super) project: Option<String>,
+}
```
szokeasaurusrex: I'd roll these arguments into the `list` command for now. We can always extract them once needed

src/commands/logs/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Args)]
+pub(super) struct LogsArgs {
+    #[command(subcommand)]
+    subcommand: LogsSubcommand,
+}
+
+#[derive(Subcommand)]
+#[command(about = "Manage logs in Sentry")]
+#[command(long_about = "Manage and query logs in Sentry. \
+This command provides access to log entries and supports live-tailing functionality.")]
```
szokeasaurusrex:
```suggestion
This command provides access to log entries.")]
```

src/commands/logs/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+enum LogsSubcommand {
+    #[command(about = LIST_ABOUT)]
+    #[command(long_about = format!("{LIST_ABOUT}. \
+Query and filter log entries from your Sentry projects. \
+Supports filtering by time period, log level, and custom queries."))]
```
szokeasaurusrex: Formatting nit – Rust ignores any whitespace before the first non-whitespace character after `\`-escaped newlines
```suggestion
    #[command(long_about = format!("{LIST_ABOUT}. \
        Query and filter log entries from your Sentry projects. \
        Supports filtering by time period, log level, and custom queries."))]
```

szokeasaurusrex: Also, I wonder whether there would be some way to put this about text in `list.rs`. If not, that's fine, but it would be preferable to have this in one place in that file imo

tests/integration/_cases/logs/logs-list-with-data.trycmd
```diff
@@ -0,0 +1,9 @@
+$ sentry-cli logs list --org wat-org --project wat-project
+? success
++------------------+---------------------+----------+----------------------+---------------------+
+| Item ID          | Timestamp           | Severity | Message              | Trace               |
++------------------+---------------------+----------+----------------------+---------------------+
+| test-item-id-001 | 2025-01-15T10:30:00 | info     | test_log_message_001 | test-trace-id-abc123 |
+| test-item-id-002 | 2025-01-15T10:31:00 | error    | test_error_message_002 | test-trace-id-def456 |
+| test-item-id-003 | 2025-01-15T10:32:00 | warning  | test_warning_message_003 | test-trace-id-ghi789 |
++------------------+---------------------+----------+----------------------+---------------------+
```
szokeasaurusrex: I wonder if there is a way in the table library to get the columns to expand to the length of the longest item

tests/integration/logs.rs
```diff
@@ -0,0 +1,37 @@
...
+        .mock_endpoint(
+            MockEndpointBuilder::new(
+                "GET",
+                "/api/0/organizations/wat-org/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=wat-project&per_page=100&statsPeriod=1h&referrer=sentry-cli-tail&sort=-timestamp",
+            )
+            .with_response_file("logs/get-logs.json"),
+        )
+        .register_trycmd_test("logs/logs-list-with-data.trycmd")
+        .with_default_token();
+}
```
szokeasaurusrex: You should be able to combine all these tests into one by using a glob pattern. Trycmd will pick up all the matching tests automatically

```suggestion
#[test]
fn command_logs() {
    TestManager::new()
        .mock_endpoint(
            MockEndpointBuilder::new(
                "GET",
                "/api/0/organizations/wat-org/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=wat-project&per_page=100&statsPeriod=1h&referrer=sentry-cli-tail&sort=-timestamp",
            )
            .with_response_file("logs/get-logs.json"),
        )
        .register_trycmd_test("logs/*.trycmd")
        .with_default_token();
}
```

src/commands/logs/list.rs
vgrozdanic: I didn't want to do it, since other datasets might not have those fields, and it probably doesn't make sense to call these fields for e.g. regular spans. This is very related to logs

tests/integration/_cases/logs/logs-list-with-data.trycmd
vgrozdanic: i guess this was just me needed to format this better in tests, it already happens in CLI

src/commands/logs/list.rs
szokeasaurusrex: Okay, fair enough. Although I still think this should probably be a constant defined with `const` rather than `let`

src/commands/logs/mod.rs
vgrozdanic: I couldn't find a way to make it work, probably there is a way, but i am lacking rust skills to do it

src/api/mod.rs
```diff
@@ -1390,6 +1413,71 @@ impl<'a> AuthenticatedApi<'a> {
...
+            params.push(format!("query={}", QueryArg(query)));
+        }
+
+        params.push(format!("per_page={}", self.per_page.unwrap_or(100)));
+        params.push(format!("statsPeriod={}", self.stats_period.unwrap_or("1h")));
+
+        params.push(format!("sort={}", self.sort.unwrap_or("-timestamp")));
+
+        params
+    }
```
szokeasaurusrex: [optional suggestion] Seems like we can use `serde` (the same crate we use to serialize/deserialize to JSON) to serialize the query params directly to a query string via the [`serde_qs` crate](https://crates.io/crates/serde_qs). Then, we would not need to manually implement this logic.

JoshFerge:
love this! what do you think about our other data types? what would be the best interface? something like this?

`sentry logs --live`
`sentry traces -live`
`sentry issues --live`
`sentry all --live` # eventually a command to live log all sentry telemetry

vgrozdanic:
> love this! what do you think about our other data types? what would be the best interface? something like this?
> 
> `sentry logs --live` `sentry traces -live` `sentry issues --live` `sentry all --live` # eventually a command to live log all sentry telemetry

@JoshFerge Yep, something like that we already had in mind, i have a second draft PR that builds on top of the work from this one (https://github.com/getsentry/sentry-cli/pull/2666) and adds live tailing only for logs, but after that it should be relatively straight forward to support tailing for other data types.

EDIT: not sure how "easy" it is to support traces and all, but issues should be no problem 😅

src/commands/logs/list.rs
```diff
@@ -0,0 +1,119 @@
...
+pub(super) fn execute(args: ListLogsArgs) -> Result<()> {
+    let config = Config::current();
+    let (default_org, default_project) = config.get_org_and_project_defaults();
+
+    let org = args
+        .org
+        .as_ref()
+        .or(default_org.as_ref())
+        .ok_or_else(|| {
+            anyhow::anyhow!("No organization specified. Use --org or set a default in config.")
```
szokeasaurusrex: I don't think many users would know how to set a default in the config. I would prefer we highlight only the `--org` argument here

```suggestion
            anyhow::anyhow!("No organization specified. Please specify an organization using the --org argument.")
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,119 @@
...
+        .fetch_organization_events(org, &options)?;
+
+    let mut table = Table::new();
+    table
+        .title_row()
+        .add("Item ID")
+        .add("Timestamp")
+        .add("Severity")
+        .add("Message")
+        .add("Trace");
```
szokeasaurusrex, line 112: How will this table output work with the `--live` option? Maybe, it would be better just to output logs line-by-line, rather than using a table, as that probably will work more easily with `--live`

tests/integration/_cases/logs/logs-list-basic.trycmd
```diff
@@ -0,0 +1,3 @@
+$ sentry-cli logs list --org wat-org --project 12345 --max-rows 0
+? success
+No logs found
```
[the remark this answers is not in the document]
szokeasaurusrex: Indeed, good catch from Cursor

tests/integration/_cases/logs/logs-list-with-custom-max-rows.trycmd
```diff
@@ -0,0 +1,5 @@
+$ sentry-cli logs list --org wat-org --project 12345 --max-rows 50
+? success
++------------------+---------------------+----------+----------------------+---------------------+
+| Item ID          | Timestamp           | Severity | Message              | Trace               |
++------------------+---------------------+----------+----------------------+---------------------+
```
szokeasaurusrex: This output looks wrong. I would expect it to be the same as `logs-list-with-data.trycmd`. Why is it different?

tests/integration/_cases/logs/logs-list-with-zero-max-rows.trycmd
```diff
@@ -0,0 +1,3 @@
+$ sentry-cli logs list --org wat-org --project 12345 --max-rows 0
```
szokeasaurusrex, line 2: imo, we should validate that users pass a value above 0 for `--max-rows` because it does not make sense to pass zero. In any case, I don't think this test is very helpful, as it does not really make any sense to do this

src/api/mod.rs
```diff
@@ -1390,6 +1413,71 @@ impl<'a> AuthenticatedApi<'a> {
...
+}
+
+impl Dataset {
+    /// Returns the string representation of the dataset
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Dataset::OurLogs => "ourlogs",
+        }
+    }
+}
```
szokeasaurusrex, line 1430: I'd add a `Display` implementation here. Then, you can just print the Dataset enum directly, without having to call `as_str` on it.
```suggestion
}

impl fmt::Display for Dataset {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{}", self.as_str())
    }
}

```

src/api/mod.rs
```diff
@@ -1390,6 +1413,71 @@ impl<'a> AuthenticatedApi<'a> {
...
+/// Available datasets for fetching organization events
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Dataset {
+    /// Our logs dataset
+    OurLogs,
+}
+
+impl Dataset {
+    /// Returns the string representation of the dataset
+    pub fn as_str(&self) -> &'static str {
```
szokeasaurusrex: I believe you can make this private

```suggestion
    fn as_str(&self) -> &'static str {
```

szokeasaurusrex (commented):
Looking better, still have a couple questions though

szokeasaurusrex:
Hey @vgrozdanic, I just attempted to test this PR locally, but it still does not seem to work as expected.

I wanted to [list these three logs](https://sentry-sdks.sentry.io/explore/logs/?logsFields=timestamp&logsFields=message&logsSortBys=-timestamp&project=4508694563782656&statsPeriod=14d).

image

To list the logs, I ran the following command.

```
➜  sentry-cli git:(vg/add-logs-command) cargo run -- logs list -p 4508694563782656 --org sentry-sdks
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/sentry-cli logs list -p 4508694563782656 --org sentry-sdks`
No logs found
```

As you can see, the command says no logs were found, even though we should have found the three logs we can see on the page in Sentry.

If you rerun the same command using `--log-level=debug` to observe the HTTP requests and responses, you will see that the server is returning the logs. So, it seems like the problem lies with the code not deserializing the server's response correctly.

```
➜  sentry-cli git:(vg/add-logs-command) cargo run -- logs list -p 4508694563782656 --org sentry-sdks --log-level=debug
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/sentry-cli logs list -p 4508694563782656 --org sentry-sdks --log-level=debug`
  INFO    2025-08-01 11:27:37.152130 +02:00 Loaded config from /Users/dszoke/Development/sentry-cli/.sentryclirc
  DEBUG   2025-08-01 11:27:37.152813 +02:00 sentry-cli version: 2.50.2, platform: "darwin", architecture: "arm64"
  INFO    2025-08-01 11:27:37.154792 +02:00 sentry-cli was invoked with the following command line: "target/debug/sentry-cli" "logs" "list" "-p" "4508694563782656" "--org" "sentry-sdks" "--log-level=debug"
  DEBUG   2025-08-01 11:27:37.155904 +02:00 request GET https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp
  DEBUG   2025-08-01 11:27:37.155945 +02:00 using token authentication
  DEBUG   2025-08-01 11:27:37.155990 +02:00 retry number 0, max retries: 5
  DEBUG   2025-08-01 11:27:37.232294 +02:00 > GET /api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp HTTP/1.1
  DEBUG   2025-08-01 11:27:37.232335 +02:00 > Host: sentry.io
  DEBUG   2025-08-01 11:27:37.232343 +02:00 > Accept: */*
  DEBUG   2025-08-01 11:27:37.232349 +02:00 > Connection: TE
  DEBUG   2025-08-01 11:27:37.232356 +02:00 > TE: gzip
  DEBUG   2025-08-01 11:27:37.232392 +02:00 > User-Agent: sentry-cli/2.50.2
  DEBUG   2025-08-01 11:27:37.234180 +02:00 > Authorization: Bearer sntryu_c***
  DEBUG   2025-08-01 11:27:37.660784 +02:00 < HTTP/1.1 200 OK
  DEBUG   2025-08-01 11:27:37.660834 +02:00 < server: nginx
  DEBUG   2025-08-01 11:27:37.660855 +02:00 < date: Fri, 01 Aug 2025 09:27:37 GMT
  DEBUG   2025-08-01 11:27:37.660867 +02:00 < content-type: application/json
  DEBUG   2025-08-01 11:27:37.660877 +02:00 < vary: Accept-Encoding,Accept-Language, Cookie
  DEBUG   2025-08-01 11:27:37.660893 +02:00 < link: <https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp&cursor=0:0:1>; rel="previous"; results="false"; cursor="0:0:1", <https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp&cursor=0:100:0>; rel="next"; results="false"; cursor="0:100:0"
  DEBUG   2025-08-01 11:27:37.660904 +02:00 < allow: GET, HEAD, OPTIONS
  DEBUG   2025-08-01 11:27:37.660913 +02:00 < access-control-allow-methods: GET, HEAD, OPTIONS
  DEBUG   2025-08-01 11:27:37.660952 +02:00 < access-control-allow-headers: X-Sentry-Auth, X-Requested-With, Origin, Accept, Content-Type, Authentication, Authorization, Content-Encoding, sentry-trace, baggage, X-CSRFToken
  DEBUG   2025-08-01 11:27:37.660963 +02:00 < access-control-expose-headers: X-Sentry-Error, X-Sentry-Direct-Hit, X-Hits, X-Max-Hits, Endpoint, Retry-After, Link
  DEBUG   2025-08-01 11:27:37.660972 +02:00 < access-control-allow-origin: *
  DEBUG   2025-08-01 11:27:37.660981 +02:00 < x-sentry-rate-limit-remaining: 996
  DEBUG   2025-08-01 11:27:37.660989 +02:00 < x-sentry-rate-limit-limit: 1000
  DEBUG   2025-08-01 11:27:37.660997 +02:00 < x-sentry-rate-limit-reset: 1754040600
  DEBUG   2025-08-01 11:27:37.661006 +02:00 < x-sentry-rate-limit-concurrentremaining: 14
  DEBUG   2025-08-01 11:27:37.661014 +02:00 < x-sentry-rate-limit-concurrentlimit: 15
  DEBUG   2025-08-01 11:27:37.661023 +02:00 < content-language: en
  DEBUG   2025-08-01 11:27:37.661032 +02:00 < reporting-endpoints: default=https://sentry.my.sentry.io/api/0/reporting-api-experiment/
  DEBUG   2025-08-01 11:27:37.661040 +02:00 < x-frame-options: deny
  DEBUG   2025-08-01 11:27:37.661048 +02:00 < x-content-type-options: nosniff
  DEBUG   2025-08-01 11:27:37.661057 +02:00 < x-xss-protection: 1; mode=block
  DEBUG   2025-08-01 11:27:37.661075 +02:00 < content-security-policy: base-uri 'none'; default-src 'none'; connect-src 'self' *.algolia.net *.algolianet.com *.algolia.io sentry.io *.sentry.io s1.sentry-cdn.com o1.ingest.sentry.io api2.amplitude.com app.pendo.io data.pendo.io reload.getsentry.net t687h3m0nh65.statuspage.io sentry.zendesk.com ekr.zdassets.com maps.googleapis.com; frame-src app.pendo.io demo.arcade.software js.stripe.com sentry.io 'self'; style-src * 'unsafe-inline'; frame-ancestors 'self' *.sentry.io; script-src 'self' 'unsafe-inline' 'report-sample' s1.sentry-cdn.com js.sentry-cdn.com browser.sentry-cdn.com statuspage-production.s3.amazonaws.com static.zdassets.com aui-cdn.atlassian.com connect-cdn.atl-paas.net js.stripe.com 'strict-dynamic' cdn.pendo.io data.pendo.io pendo-io-static.storage.googleapis.com pendo-static-5634074999128064.storage.googleapis.com; media-src *; object-src 'none'; img-src * blob: data:; worker-src blob:; font-src * data:; report-uri https://o1.ingest.sentry.io/api/54785/security/?sentry_key=f724a8a027db45f5b21507e7142ff78e&sentry_release=81196427e84449af89ec87c74201522e28d79b6b
  DEBUG   2025-08-01 11:27:37.661091 +02:00 < x-envoy-upstream-service-time: 296
  DEBUG   2025-08-01 11:27:37.661100 +02:00 < x-sentry-proxy-url: http://10.2.0.67:8999/api/0/organizations/sentry-sdks/events/
  DEBUG   2025-08-01 11:27:37.661131 +02:00 < x-envoy-attempt-count: 1
  DEBUG   2025-08-01 11:27:37.661147 +02:00 < x-served-by: frontend-default-6b97b7cfd8-59bxm
  DEBUG   2025-08-01 11:27:37.661159 +02:00 < strict-transport-security: max-age=31536000; includeSubDomains; preload
  DEBUG   2025-08-01 11:27:37.661168 +02:00 < via: 1.1 google
  DEBUG   2025-08-01 11:27:37.661177 +02:00 < Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
  DEBUG   2025-08-01 11:27:37.661187 +02:00 < Transfer-Encoding: chunked
  DEBUG   2025-08-01 11:27:37.663707 +02:00 response status: 200
  DEBUG   2025-08-01 11:27:37.663749 +02:00 body: {"data":[{"sentry.item_id":"d1befd406c0620b5047757aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"error","timestamp":"2025-08-01T09:25:48+00:00","message":"Database connection failed","tags[sentry.timestamp_precise,number]":1.75404034824715e+18},{"sentry.item_id":"975283f1a4430cac5a7c57aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"info","timestamp":"2025-08-01T09:25:48+00:00","message":"Hello, world!","tags[sentry.timestamp_precise,number]":1.754040348247139e+18},{"sentry.item_id":"9959d7e2eb1b929dd77556aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"info","timestamp":"2025-08-01T09:25:48+00:00","message":"Hello, world!","tags[sentry.timestamp_precise,number]":1.7540403482469248e+18}],"meta":{"fields":{"sentry.item_id":"string","trace":"string","severity":"string","timestamp":"string","message":"string","tags[sentry.timestamp_precise,number]":"number"},"units":{"sentry.item_id":null,"trace":null,"severity":null,"timestamp":null,"message":null,"tags[sentry.timestamp_precise,number]":null},"isMetricsData":false,"isMetricsExtractedData":false,"tips":{},"datasetReason":"unchanged","dataset":"ourlogs","dataScanned":"full","accuracy":{"confidence":[{},{},{}]}},"confidence":[{},{},{}]}
No logs found
  INFO    2025-08-01 11:27:37.664836 +02:00 Skipping update nagger update check
```

vgrozdanic:
> Hey @vgrozdanic, I just attempted to test this PR locally, but it still does not seem to work as expected.
> 
> I wanted to [list these three logs](https://sentry-sdks.sentry.io/explore/logs/?logsFields=timestamp&logsFields=message&logsSortBys=-timestamp&project=4508694563782656&statsPeriod=14d).
> 
> image
> To list the logs, I ran the following command.
> 
> ```
> ➜  sentry-cli git:(vg/add-logs-command) cargo run -- logs list -p 4508694563782656 --org sentry-sdks
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.48s
>      Running `target/debug/sentry-cli logs list -p 4508694563782656 --org sentry-sdks`
> No logs found
> ```
> 
> As you can see, the command says no logs were found, even though we should have found the three logs we can see on the page in Sentry.
> 
> If you rerun the same command using `--log-level=debug` to observe the HTTP requests and responses, you will see that the server is returning the logs. So, it seems like the problem lies with the code not deserializing the server's response correctly.
> 
> ```
> ➜  sentry-cli git:(vg/add-logs-command) cargo run -- logs list -p 4508694563782656 --org sentry-sdks --log-level=debug
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
>      Running `target/debug/sentry-cli logs list -p 4508694563782656 --org sentry-sdks --log-level=debug`
>   INFO    2025-08-01 11:27:37.152130 +02:00 Loaded config from /Users/dszoke/Development/sentry-cli/.sentryclirc
>   DEBUG   2025-08-01 11:27:37.152813 +02:00 sentry-cli version: 2.50.2, platform: "darwin", architecture: "arm64"
>   INFO    2025-08-01 11:27:37.154792 +02:00 sentry-cli was invoked with the following command line: "target/debug/sentry-cli" "logs" "list" "-p" "4508694563782656" "--org" "sentry-sdks" "--log-level=debug"
>   DEBUG   2025-08-01 11:27:37.155904 +02:00 request GET https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp
>   DEBUG   2025-08-01 11:27:37.155945 +02:00 using token authentication
>   DEBUG   2025-08-01 11:27:37.155990 +02:00 retry number 0, max retries: 5
>   DEBUG   2025-08-01 11:27:37.232294 +02:00 > GET /api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp HTTP/1.1
>   DEBUG   2025-08-01 11:27:37.232335 +02:00 > Host: sentry.io
>   DEBUG   2025-08-01 11:27:37.232343 +02:00 > Accept: */*
>   DEBUG   2025-08-01 11:27:37.232349 +02:00 > Connection: TE
>   DEBUG   2025-08-01 11:27:37.232356 +02:00 > TE: gzip
>   DEBUG   2025-08-01 11:27:37.232392 +02:00 > User-Agent: sentry-cli/2.50.2
>   DEBUG   2025-08-01 11:27:37.234180 +02:00 > Authorization: Bearer sntryu_c***
>   DEBUG   2025-08-01 11:27:37.660784 +02:00 < HTTP/1.1 200 OK
>   DEBUG   2025-08-01 11:27:37.660834 +02:00 < server: nginx
>   DEBUG   2025-08-01 11:27:37.660855 +02:00 < date: Fri, 01 Aug 2025 09:27:37 GMT
>   DEBUG   2025-08-01 11:27:37.660867 +02:00 < content-type: application/json
>   DEBUG   2025-08-01 11:27:37.660877 +02:00 < vary: Accept-Encoding,Accept-Language, Cookie
>   DEBUG   2025-08-01 11:27:37.660893 +02:00 < link: <https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp&cursor=0:0:1>; rel="previous"; results="false"; cursor="0:0:1", <https://sentry.io/api/0/organizations/sentry-sdks/events/?dataset=ourlogs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=4508694563782656&per_page=100&statsPeriod=1h&sort=-timestamp&cursor=0:100:0>; rel="next"; results="false"; cursor="0:100:0"
>   DEBUG   2025-08-01 11:27:37.660904 +02:00 < allow: GET, HEAD, OPTIONS
>   DEBUG   2025-08-01 11:27:37.660913 +02:00 < access-control-allow-methods: GET, HEAD, OPTIONS
>   DEBUG   2025-08-01 11:27:37.660952 +02:00 < access-control-allow-headers: X-Sentry-Auth, X-Requested-With, Origin, Accept, Content-Type, Authentication, Authorization, Content-Encoding, sentry-trace, baggage, X-CSRFToken
>   DEBUG   2025-08-01 11:27:37.660963 +02:00 < access-control-expose-headers: X-Sentry-Error, X-Sentry-Direct-Hit, X-Hits, X-Max-Hits, Endpoint, Retry-After, Link
>   DEBUG   2025-08-01 11:27:37.660972 +02:00 < access-control-allow-origin: *
>   DEBUG   2025-08-01 11:27:37.660981 +02:00 < x-sentry-rate-limit-remaining: 996
>   DEBUG   2025-08-01 11:27:37.660989 +02:00 < x-sentry-rate-limit-limit: 1000
>   DEBUG   2025-08-01 11:27:37.660997 +02:00 < x-sentry-rate-limit-reset: 1754040600
>   DEBUG   2025-08-01 11:27:37.661006 +02:00 < x-sentry-rate-limit-concurrentremaining: 14
>   DEBUG   2025-08-01 11:27:37.661014 +02:00 < x-sentry-rate-limit-concurrentlimit: 15
>   DEBUG   2025-08-01 11:27:37.661023 +02:00 < content-language: en
>   DEBUG   2025-08-01 11:27:37.661032 +02:00 < reporting-endpoints: default=https://sentry.my.sentry.io/api/0/reporting-api-experiment/
>   DEBUG   2025-08-01 11:27:37.661040 +02:00 < x-frame-options: deny
>   DEBUG   2025-08-01 11:27:37.661048 +02:00 < x-content-type-options: nosniff
>   DEBUG   2025-08-01 11:27:37.661057 +02:00 < x-xss-protection: 1; mode=block
>   DEBUG   2025-08-01 11:27:37.661075 +02:00 < content-security-policy: base-uri 'none'; default-src 'none'; connect-src 'self' *.algolia.net *.algolianet.com *.algolia.io sentry.io *.sentry.io s1.sentry-cdn.com o1.ingest.sentry.io api2.amplitude.com app.pendo.io data.pendo.io reload.getsentry.net t687h3m0nh65.statuspage.io sentry.zendesk.com ekr.zdassets.com maps.googleapis.com; frame-src app.pendo.io demo.arcade.software js.stripe.com sentry.io 'self'; style-src * 'unsafe-inline'; frame-ancestors 'self' *.sentry.io; script-src 'self' 'unsafe-inline' 'report-sample' s1.sentry-cdn.com js.sentry-cdn.com browser.sentry-cdn.com statuspage-production.s3.amazonaws.com static.zdassets.com aui-cdn.atlassian.com connect-cdn.atl-paas.net js.stripe.com 'strict-dynamic' cdn.pendo.io data.pendo.io pendo-io-static.storage.googleapis.com pendo-static-5634074999128064.storage.googleapis.com; media-src *; object-src 'none'; img-src * blob: data:; worker-src blob:; font-src * data:; report-uri https://o1.ingest.sentry.io/api/54785/security/?sentry_key=f724a8a027db45f5b21507e7142ff78e&sentry_release=81196427e84449af89ec87c74201522e28d79b6b
>   DEBUG   2025-08-01 11:27:37.661091 +02:00 < x-envoy-upstream-service-time: 296
>   DEBUG   2025-08-01 11:27:37.661100 +02:00 < x-sentry-proxy-url: http://10.2.0.67:8999/api/0/organizations/sentry-sdks/events/
>   DEBUG   2025-08-01 11:27:37.661131 +02:00 < x-envoy-attempt-count: 1
>   DEBUG   2025-08-01 11:27:37.661147 +02:00 < x-served-by: frontend-default-6b97b7cfd8-59bxm
>   DEBUG   2025-08-01 11:27:37.661159 +02:00 < strict-transport-security: max-age=31536000; includeSubDomains; preload
>   DEBUG   2025-08-01 11:27:37.661168 +02:00 < via: 1.1 google
>   DEBUG   2025-08-01 11:27:37.661177 +02:00 < Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
>   DEBUG   2025-08-01 11:27:37.661187 +02:00 < Transfer-Encoding: chunked
>   DEBUG   2025-08-01 11:27:37.663707 +02:00 response status: 200
>   DEBUG   2025-08-01 11:27:37.663749 +02:00 body: {"data":[{"sentry.item_id":"d1befd406c0620b5047757aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"error","timestamp":"2025-08-01T09:25:48+00:00","message":"Database connection failed","tags[sentry.timestamp_precise,number]":1.75404034824715e+18},{"sentry.item_id":"975283f1a4430cac5a7c57aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"info","timestamp":"2025-08-01T09:25:48+00:00","message":"Hello, world!","tags[sentry.timestamp_precise,number]":1.754040348247139e+18},{"sentry.item_id":"9959d7e2eb1b929dd77556aef3649801","trace":"a2328953e06b4f7100bcea441c357259","severity":"info","timestamp":"2025-08-01T09:25:48+00:00","message":"Hello, world!","tags[sentry.timestamp_precise,number]":1.7540403482469248e+18}],"meta":{"fields":{"sentry.item_id":"string","trace":"string","severity":"string","timestamp":"string","message":"string","tags[sentry.timestamp_precise,number]":"number"},"units":{"sentry.item_id":null,"trace":null,"severity":null,"timestamp":null,"message":null,"tags[sentry.timestamp_precise,number]":null},"isMetricsData":false,"isMetricsExtractedData":false,"tips":{},"datasetReason":"unchanged","dataset":"ourlogs","dataScanned":"full","accuracy":{"confidence":[{},{},{}]}},"confidence":[{},{},{}]}
> No logs found
>   INFO    2025-08-01 11:27:37.664836 +02:00 Skipping update nagger update check
> ```

@szokeasaurusrex rust skill issue - i have found a bug in my code, working on a fix... I have found a couple of more mistakes in my code, but soon will push the updated version. Sorry to bother you this much, as you can tell, i haven't written much rust so far 😅

szokeasaurusrex:
It's okay, no worries @vgrozdanic!

But please, once you have fixed the problem, make sure you have a test case which validates that the command handles this type of response correctly

tests/integration/_cases/logs/logs-list-with-custom-max-rows.trycmd
vgrozdanic: Somehow this test wasn't even running because i have screwed it up, now all of the tests should be fixed

src/commands/logs/list.rs
vgrozdanic: I would still leave it as a table, since all of the others `list` commands print out the table.

For `--live` i am planning to just print a table header, and body without the bottom border, so we would just continue printing new rows on every re-fetch

src/api/mod.rs
vgrozdanic: I wouldn't do it as a part of this PR. I agree that this would be a nice improvement, but i would rather do it as a follow up, and not introduce a new crate, and make this PR even more complicated and prolong the time until it is done :) 

Happy to do it as a refactor follow up

tests/integration/logs.rs
vgrozdanic, line 1: I left each test in their own function because that most of them require mocking of endpoints, and i find it easier to understand and read if each of them is a special test case. Especially if one of them starts failing, it is easier to know if mock is wrong, or the expected output has changed, or smth 3rd.

src/commands/logs/list.rs
JoshFerge: we could specify how they could do that, like `(see sentry-cli config --help) to learn how to configure your cli`

szokeasaurusrex: @JoshFerge let's keep it simple please. Also, `sentry-cli config` is not a real command

JoshFerge: i'm not saying it needs to be done in this PR, nor was i suggesting it was a real command. -- just an idea for a UX flow. many CLIs function in this way where it describes how to set a default org if you don't have one set. while specifying how to make your query work without a default.


>I don't think many users would know how to set a default in the config

this seems like a problem! why don't we make it easier for them! :)

szokeasaurusrex: Sounds reasonable, we have more urgent priorities in Sentry CLI, though

tests/integration/logs.rs
szokeasaurusrex: Yeah, that is reasonable

src/api/mod.rs
szokeasaurusrex: I may also be able to do it, should be easy

src/commands/logs/list.rs
szokeasaurusrex: Fair enough

src/commands/logs/list.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let logs_to_show = &logs[..args.max_rows.min(logs.len())];
+    for log in logs_to_show {
```
szokeasaurusrex: Let's please rewrite this way, it is easier to read (this is what I meant with my `take` suggestion in Slack)

```suggestion
    for log in logs.iter().take(args.max_rows) {
```

tests/integration/_cases/logs/logs-help.trycmd
```diff
@@ -0,0 +1,32 @@
+```
+$ sentry-cli logs --help
+? success
+Manage and query logs in Sentry. This command provides access to log entries.
+
+Usage: sentry-cli logs [OPTIONS] [COMMAND]
```
szokeasaurusrex: As `[EXE]` adds `.exe` on Windows, but evaluates to an empty string on other platforms, I believe that we can remove the `logs-help-windows.trycmd` if we make this edit here

```suggestion
Usage: sentry-cli[EXE] logs [OPTIONS] [COMMAND]
```

tests/integration/logs.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[test]
+fn command_logs_help() {
+    let manager = TestManager::new();
+
+    #[cfg(not(windows))]
+    manager.register_trycmd_test("logs/logs-help.trycmd");
+    #[cfg(windows)]
+    manager.register_trycmd_test("logs/logs-help-windows.trycmd");
```
szokeasaurusrex: If you make the requested change to `logs-help.trycmd`, this should work

```suggestion
    manager.register_trycmd_test("logs/logs-help.trycmd");
```

szokeasaurusrex (approved):
Looks good, still some comments to address before merge (but I'll take care of it)

src/commands/logs/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(Subcommand)]
+#[command(about = "Manage logs in Sentry")]
+#[command(long_about = "Manage and query logs in Sentry. \
+    This command provides access to log entries.")]
+enum LogsSubcommand {
+    #[command(about = LIST_ABOUT)]
+    #[command(long_about = format!("{LIST_ABOUT}. \
+    Query and filter log entries from your Sentry projects. \
+    Supports filtering by time period, log level, and custom queries."))]
```
szokeasaurusrex: Filtering by time period does not seem to be possible

src/commands/logs/list.rs
```diff
@@ -0,0 +1,131 @@
...
+    args: &ListLogsArgs,
+) -> Result<()> {
+    let options = FetchEventsOptions {
+        dataset: Dataset::OurLogs,
+        fields,
+        project_id: Some(project),
+        cursor: None,
+        query,
+        per_page: Some(args.max_rows),
+        stats_period: Some("1h"),
```
szokeasaurusrex: This value should either be user-configurable, or we should state in the command's `--help` text that we only display logs from the last hour

szokeasaurusrex (changes requested):
While playing around with this PR, I noticed that it appears we only support displaying logs from the last hour. We should probably instead allow the time period to be user-configured, either in this PR or in a future PR.

If we add the ability to user-configure the time period in a future PR, this PR should state clearly in the help text that for now, only logs from the past hour are displayed

src/commands/logs/list.rs
shellmayr: So since the way we query the API is descending by time, and we limit to 100 rows via the `per_page` parameter, we will always get the 100 most recent logs. I did a quick test on our API with 90d and 1h, and the timing is not really different, so I'd propose defaulting to 90d and not making it user-configurable for now, that would always give the user the most recent logs. What do you think about that @szokeasaurusrex?

szokeasaurusrex: @shellmayr, is the `stats_period` even needed in the query string? If not, let's just delete the struct field completely and not send the parameter, assuming that the default time period the server assumes is reasonable. Otherwise, I would just hardcode the `90d` default

shellmayr: @szokeasaurusrex I think it defaults to 14 days if not explicitly set

szokeasaurusrex: Hmm okay, I guess either the server default or a 90 day default is reasonable in that case

shellmayr: I'd rather it be explicit, so the CLI doesn't rely on current, unspecified behaviour - that way the behaviour should stay the same if something changes in the API configuration.

szokeasaurusrex: Very wise

src/commands/logs/list.rs
```diff
@@ -0,0 +1,130 @@
...
+fn validate_max_rows(s: &str) -> Result<usize, String> {
+    let value = s
+        .parse::<usize>()
+        .map_err(|_| "invalid number".to_owned())?;
+    if value == 0 {
+        Err("max-rows must be greater than 0".to_owned())
+    } else {
+        Ok(value)
+    }
+}
```
szokeasaurusrex: The following includes several improvements:
1. We define a constant for the allowed range of max_rows.
2. The validation returns an `anyhow::Result`, maintaining consistency with the rest of the codebase.
3. The validation also checks that `max_rows` is below the maximum value of 1000.

```suggestion
const MAX_ROWS_RANGE: std::ops::RangeInclusive<usize> = 1..=1000;

/// Validate that max_rows is in the allowed range
fn validate_max_rows(s: &str) -> Result<usize> {
    let value = s.parse()?;
    if MAX_ROWS_RANGE.contains(&value) {
        Ok(value)
    } else {
        Err(anyhow::anyhow!(
            "max-rows must be between {} and {}",
            MAX_ROWS_RANGE.start(),
            MAX_ROWS_RANGE.end()
        ))
    }
}
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[arg(help = "The organization ID or slug.")]
+    org: Option<String>,
+
+    #[arg(short = 'p', long = "project")]
+    #[arg(help = "The project ID (slug not supported).")]
+    project: Option<String>,
+
+    #[arg(long = "max-rows", default_value = "100")]
+    #[arg(value_parser = validate_max_rows)]
+    #[arg(help = "Maximum number of log entries to fetch and display (max 1000).")]
```
szokeasaurusrex:
```suggestion
    #[arg(help = format!("Maximum number of log entries to fetch and display (max {})", MAX_ROWS_RANGE.end()))]
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,130 @@
...
+    let (default_org, default_project) = config.get_org_and_project_defaults();
+
+    let org = args
+        .org
+        .as_ref()
+        .or(default_org.as_ref())
+        .ok_or_else(|| {
+            anyhow::anyhow!("No organization specified. Please specify an organization using the --org argument.")
+        })?
+        .to_owned();
```
szokeasaurusrex: I believe this `to_owned` call is unnecessary. Removing it may require you to remove the `&` sign when passing it to `execute_simple_fetch` in order to appease the linter

```suggestion
        })?;
```

src/commands/logs/list.rs
```diff
@@ -0,0 +1,130 @@
...
+    let project = args
+        .project
+        .as_ref()
+        .or(default_project.as_ref())
+        .ok_or_else(|| {
+            anyhow::anyhow!("No project specified. Use --project or set a default in config.")
+        })?
+        .to_owned();
```
szokeasaurusrex:
```suggestion
        })?;
```

src/api/mod.rs
```diff
@@ -1390,6 +1413,78 @@ impl<'a> AuthenticatedApi<'a> {
     }
 }
 
+/// Available datasets for fetching organization events
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Dataset {
+    /// Our logs dataset
+    OurLogs,
```
szokeasaurusrex: `ourlogs` is – as of a couple days ago – [deprecated in the backend](https://github.com/getsentry/sentry/blob/afdf3166fdfdfe5480c49c5b7a2fec90d92d92d4/src/sentry/snuba/utils.py#L33-L35), from what I can tell

```suggestion
    /// Logs dataset
    Logs,
```

src/api/mod.rs
```diff
@@ -1390,6 +1413,78 @@ impl<'a> AuthenticatedApi<'a> {
...
+}
+
+impl Dataset {
+    /// Returns the string representation of the dataset
+    fn as_str(&self) -> &'static str {
+        match self {
+            Dataset::OurLogs => "ourlogs",
```
szokeasaurusrex:
```suggestion
            Dataset::Logs => "logs",
```

src/api/mod.rs
```diff
@@ -1390,6 +1413,78 @@ impl<'a> AuthenticatedApi<'a> {
...
+/// Options for fetching organization events
+pub struct FetchEventsOptions<'a> {
+    /// Dataset to fetch events from
+    pub dataset: Dataset,
+    /// Fields to include in the response
+    pub fields: &'a [&'a str],
+    /// Project ID to filter events by
+    pub project_id: Option<&'a str>,
+    /// Cursor for pagination
+    pub cursor: Option<&'a str>,
```
szokeasaurusrex, line 1447: From what I can tell, this only ever gets set to `None`, so let's please remove the field for now

```suggestion
```

src/api/mod.rs
```diff
@@ -1390,6 +1413,78 @@ impl<'a> AuthenticatedApi<'a> {
...
+}
+
+/// Options for fetching organization events
+pub struct FetchEventsOptions<'a> {
+    /// Dataset to fetch events from
+    pub dataset: Dataset,
+    /// Fields to include in the response
+    pub fields: &'a [&'a str],
+    /// Project ID to filter events by
+    pub project_id: Option<&'a str>,
```
szokeasaurusrex: Is this optional as far as the backend is concerned? If yes, let's also make it optional when running the command (currently, it is required to specify a project). If not, let's make the struct field required:


```suggestion
    pub project_id: &'a str,
```

src/api/mod.rs
```diff
@@ -1390,6 +1413,78 @@ impl<'a> AuthenticatedApi<'a> {
...
+    /// Cursor for pagination
+    pub cursor: Option<&'a str>,
+    /// Query string to filter events
+    pub query: Option<&'a str>,
+    /// Number of events per page (default: 100)
+    pub per_page: Option<usize>,
+    /// Time period for stats (default: "1h")
+    pub stats_period: Option<&'a str>,
+    /// Sort order (default: "-timestamp")
+    pub sort: Option<&'a str>,
```
szokeasaurusrex: For now, all of these fields appear to always be being set to a `Some` variant, so the default value never ends up being used. 

Therefore, I would like to see the struct fields be made non-optional, so we can also get rid of the defaults in the code, and make the API simpler. We can always later make the fields optional, as needed.
```suggestion
    /// Query string to filter events
    pub query: &'a str,
    /// Number of events per page
    pub per_page: usize,
    /// Time period for stats
    pub stats_period: &'a str,
    /// Sort order
    pub sort: &'a str,
```

tests/integration/_cases/logs/logs-list-basic.trycmd
szokeasaurusrex, line 1: Can we please rename this file to something which indicates that the server is responding without any logs? Something like `logs-list-no-logs.trycmd`

tests/integration/_cases/logs/logs-list-basic.trycmd
```diff
@@ -0,0 +1,6 @@
+```
+$ sentry-cli logs list --org wat-org --project 12345 --max-rows 1
```
szokeasaurusrex: Since what we are fundamentally testing here is the behavior when the server replies without any logs, I would remove the `--max-rows` parameter, as it is irrelevant to what we are fundamentally trying to test here.


```suggestion
$ sentry-cli logs list --org wat-org --project 12345
```

szokeasaurusrex (commented):
I think this needs a couple final polishing touches but otherwise looks close to done! Thanks Simon

src/commands/logs/list.rs
shellmayr: changed!

src/commands/logs/list.rs
shellmayr: changed!

src/commands/logs/list.rs
shellmayr: good catch - changed this one as well!

src/api/mod.rs
shellmayr: This will be needed in the follow-up PR at the latest, and is a central part of our API patterns, and since it will be used in https://github.com/getsentry/sentry-cli/pull/2666 - I'd prefer we put it in and not remove it and add it in again in a few days

src/api/mod.rs
shellmayr: FWIW I'm not sure whether users know what a _project_id_ actually is, because this is not the slug, but the numeric identifier. Do we have a facility in sentry-cli to map from a project slug to a project id?

shellmayr: & Agreed to optional, but I think then the output needs to be adapted to include the project slug, so that users can understand which project a specific log entry is from.

szokeasaurusrex: > Do we have a facility in sentry-cli to map from a project slug to a project id?

No, we want to add this in a future PR though, so that the logs command can also take project slugs (rather would not add it to this PR, as it is already quite big as is)

All other CLI commands support project slugs natively – the backend APIs we call can take either the ID or the slug

src/api/mod.rs
shellmayr: @szokeasaurusrex Calls to the endpoint don't _need_ to have a query, stats_period, per_page or sort. What are we achieving by making them non-optional?

src/api/mod.rs
shellmayr: Let's make this required for now, we can also make the API more lenient later, for now I'd like this first version to go out ASAP

src/api/mod.rs
szokeasaurusrex: I want to keep the Sentry CLI code simple. Making these optional means we have to add code paths to Sentry CLI, which for now, are not being used, because we always set the `query`, `stats_period`, `per_page`, and `sort`. Rust, however, still forces us to handle the cases of these being `None` (even though, in reality, they never are `None`). 

As a result, [we have code paths, which are completely unused](https://github.com/getsentry/sentry-cli/blob/74f859624e684896b1b94451ac765a846221778b/src/api/mod.rs#L1473-L1480), to handle the `None` cases. In the case of the `query`, we only provide it if it is there. For `stats_period`, `per_page`, and `sort`, we have default values defined, which are never used. From reading the code, however, it is not immediately clear that these default values are never used.

The problem with keeping them as `Option` is just that it increases complexity and it makes it harder to see where the actual values get defined. I understand that the values are optional from the server perspective, but that absolutely does not mean that we need to make them optional in the CLI's API

szokeasaurusrex: @shellmayr As an alternative, I would also be okay with just hardcoding the defaults and removing these fields from the struct completely. I'm okay with whatever you think is more reasonable, just want to avoid the `Option` when its adding currently unnecessary complexity by handling cases which never occur.

tests/integration/logs.rs
```diff
@@ -0,0 +1,39 @@
...
+        .with_default_token();
+}
+
+#[test]
+fn command_logs_no_logs_found() {
+    TestManager::new()
+        .mock_endpoint(
+            MockEndpointBuilder::new(
+                "GET", 
+                "/api/0/organizations/wat-org/events/?dataset=logs&field=sentry.item_id&field=trace&field=severity&field=timestamp&field=message&project=12345&per_page=100&statsPeriod=90d&sort=-timestamp"
```
szokeasaurusrex: Hmm @shellmayr, I might have been wrong about the `query` not needing to be optional. Looks like this test is failing now because the CLI is adding an empty `query=&` parameter to this URL, causing the `501` error

shellmayr: I think that should be fine, fixing that test now

szokeasaurusrex: I think it is the same problem with the test above it

PR #2666, opened 2025-07-31
feat(logs): support log streaming 

vgrozdanic:
Adds support for live streaming of logs by polling the API every 2 seconds (by default).

Polling interval is configurable via command line argument.

### DIFF
--- a/src/commands/logs/list.rs
+++ b/src/commands/logs/list.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-use std::num::{NonZero, NonZeroUsize};
+use std::num::NonZeroUsize;
 use std::time::Duration;
 
 use anyhow::Result;
@@ -12,10 +12,10 @@ use crate::utils::formatting::Table;
 
 const MAX_ROWS_RANGE: std::ops::RangeInclusive<usize> = 1..=1000;
 /// Validate that max_rows is in the allowed range
-fn validate_max_rows(s: &str) -> Result<usize> {
-    let value = s.parse()?;
+fn validate_max_rows(s: &str) -> Result<NonZeroUsize> {
+    let value = s.parse::<usize>()?;
     if MAX_ROWS_RANGE.contains(&value) {
-        Ok(value)
+        NonZeroUsize::new(value).ok_or_else(|| anyhow::anyhow!("max-rows must be greater than 0"))
     } else {
         Err(anyhow::anyhow!(
             "max-rows must be between {} and {}",
@@ -63,7 +63,7 @@ pub(super) struct ListLogsArgs {
     #[arg(long = "max-rows", default_value = "100")]
     #[arg(value_parser = validate_max_rows)]
     #[arg(help = format!("Maximum number of log entries to fetch and display (max {}).", MAX_ROWS_RANGE.end()))]
-    max_rows: usize,
+    max_rows: NonZeroUsize,
 
     #[arg(long = "query", default_value = "")]
     #[arg(help = "Query to filter logs. Example: \"level:error\"")]
@@ -133,7 +133,7 @@ fn execute_single_fetch(
         project_id,
         cursor: None,
         query,
-        per_page: args.max_rows,
+        per_page: args.max_rows.get(),
         stats_period: "90d",
         sort: "-timestamp",
     };
@@ -151,7 +151,7 @@ fn execute_single_fetch(
         .add("Message")
         .add("Trace");
 
-    for log in logs.iter().take(args.max_rows) {
+    for log in logs.iter().take(args.max_rows.get()) {
         let row = table.add_row();
         row.add(&log.item_id)
             .add(&log.timestamp)
@@ -256,8 +256,7 @@ fn execute_live_streaming(
     fields: &[&str],
     args: &ListLogsArgs,
 ) -> Result<()> {
-    let mut deduplicator =
-        LogDeduplicator::new(NonZero::new(args.max_rows).expect("max-rows should be non-zero"));
+    let mut deduplicator = LogDeduplicator::new(args.max_rows);
     let poll_duration = Duration::from_secs(args.poll_interval);
     let mut new_only_tracker = ConsecutiveNewOnlyTracker::new(3); // Warn after 3 consecutive batches of only new logs
 
@@ -274,7 +273,7 @@ fn execute_live_streaming(
             project_id: project,
             cursor: None,
             query,
-            per_page: args.max_rows,
+            per_page: args.max_rows.get(),
             stats_period: "10m",
             sort: "-timestamp",
         };


src/commands/logs/list.rs
```diff
@@ -129,3 +146,257 @@ fn execute_single_fetch(
...
+        for log in new_logs {
+            if !self.seen_ids.contains(&log.item_id) {
+                self.seen_ids.insert(log.item_id.clone());
+                self.buffer.push(log.clone());
+                unique_logs.push(log);
+            }
+        }
+
+        // Maintain buffer size limit by removing oldest entries
+        while self.buffer.len() > self.max_size {
```
vgrozdanic: this is done to prevent OOM in case this keeps running for a long time

src/utils/formatting.rs
```diff
@@ -98,6 +98,75 @@ impl Table {
...
+        // Use the same base format as print_rows_only but add header separator
+        let mut format = *prettytable::format::consts::FORMAT_NO_BORDER;
+        format.column_separator('|');
+        format.padding(1, 1);
+        format.borders('|'); // Add left and right borders
+                             // Add separator line under the header
+        format.separator(
+            prettytable::format::LinePosition::Title,
+            prettytable::format::LineSeparator::new('-', '+', '+', '+'),
+        );
```
vgrozdanic: We need to do custom formatting, because every other formatting adds empty lines after printing out the table. We want it to seem like it's infinite table, without empty lines in between printing multiple rows, so we need to use `FORMAT_NO_BORDER`, and define our own line separator for header

szokeasaurusrex:
Deferring review until after HackWeek – please ping me if you need a review more urgently

src/commands/logs/list.rs
```diff
@@ -143,10 +168,327 @@ fn execute_single_fetch(
...
+            if let Some(removed_log) = self.buffer.pop_front() {
+                self.seen_ids.remove(&removed_log.item_id);
+            }
+        }
+
+        unique_logs
+    }
+}
```
szokeasaurusrex: [change request] If I am understanding the intention behind the code correctly, I think what you want here is an LRU cache, which is implemented in the [`lru` crate](https://crates.io/crates/lru).

The current implementation with the separate `HashSet` and `VecDeque` is brittle. Consider, for instance, the case where the same log is infinitely repeated. The first 10k instances are dedeuplicated, but then, the `VecDeque` is full, so on every iteration, we pop the log from the `VecDeque` and remove it from the `HashSet` (even though it is the only item in the `HashSet`). As a result, from the 10001-th iteration onwards, we print the log every time.

src/commands/logs/list.rs
```diff
@@ -143,10 +168,327 @@ fn execute_single_fetch(
...
+
+/// Tracks consecutive batches of all-new logs and manages warning state.
+///
+/// A batch is "all-new" when every fetched log is unique (no duplicates).
+/// This struct tracks how many consecutive all-new batches we've seen and
+/// warns when the count reaches the threshold, suggesting the user might be
+/// missing some logs due to overly broad filtering.
```
szokeasaurusrex, line 207: It is not clear to me why we want this warning, could you explain it? We could also discuss in a meeting so I can better understand the goal

szokeasaurusrex: Does every log being unique mean that none of the logs have been seen in a previous fetch? If so, what is the expected behavior if the most recent fetch itself contains duplicated logs? Or, is this not possible, since every log (even if it is the same message) has the same ID?

src/utils/formatting.rs
szokeasaurusrex: Can we revisit whether it makes sense to output streamed logs in the table format? I really think it would be better, certainly in terms of making the implementation easier, but possibly also from a usability standpoint, if we just outputted them in an unstructured format. Or, if we also want to support a structured format for AI agents, I would suggest a separate mode that outputs in JSON or CSV, which could be enabled with a `--structured` flag.

My concern with table format is that the implementation is complex, it's not human readable unless you widen your terminal window, and also it probably is not a very standard format for LLMs compared with something like JSON or CSV. I think we can do better by having the default log format look like the logs that Sentry CLI outputs (human-readable), with a structured logging option.

What are your thoughts?

szokeasaurusrex (changes requested):
Requesting changes, as the deduplication logic seems strange/possibly wrong, and we can improve it. Please see inline comment for details 👍

src/commands/logs/list.rs
szokeasaurusrex: Discussed offline

src/commands/logs/list.rs
szokeasaurusrex: Discussed offline, @vgrozdanic explained the implementation, and now I see it is correct.

Cargo.toml
```diff
@@ -80,6 +80,7 @@ data-encoding = "2.3.3"
 magic_string = "0.3.4"
 chrono-tz = "0.8.4"
 secrecy = "0.8.0"
+lru = "0.16.0"
```
vgrozdanic, line 83: after offline discussion with @szokeasaurusrex - we decided to use `lru` crate instead of rolling our own solution for this

src/commands/logs/list.rs
```diff
@@ -143,10 +169,298 @@ fn execute_single_fetch(
...
+/// Manages deduplication of log entries using an LRU cache
+struct LogDeduplicator {
+    /// LRU cache of seen log IDs
+    seen_ids: LruCache<String, ()>,
+}
+
+const MAX_DEDUP_BUFFER_SIZE: usize = 10_000;
+
+impl LogDeduplicator {
+    fn new(max_size: usize) -> Self {
```
szokeasaurusrex: tbh, I would suggest we just take a `NonZeroUsize` here, then we don't need to handle the error case (which just happens if the value is zero).

```suggestion
    fn new(max_size: NonZeroUsize) -> Self {
```

src/commands/logs/list.rs
```diff
@@ -143,10 +169,295 @@ fn execute_single_fetch(
...
+
+fn execute_live_streaming(
+    api: &Api,
+    org: &str,
+    project: Option<&str>,
+    query: &str,
+    fields: &[&str],
+    args: &ListLogsArgs,
+) -> Result<()> {
+    let mut deduplicator = LogDeduplicator::new(MAX_DEDUP_BUFFER_SIZE);
```
szokeasaurusrex: Why not have the max size here be the set to the number of logs we are fetching per request? We probably would need to add that as a parameter to the signature.

szokeasaurusrex: Just my thinking process: If we are always fetching the most recent $x$ logs, and we assume that logs can neither be deleted nor can there timestamps be modified, then I am pretty sure the following holds:

1. Suppose fetch $n$ includes logs $\ell_1, \ell_2, ..., \ell_m$, where $m$ is maximum number of logs.
2. Suppose fetch $n+1$ then includes logs $\ell_{m+1}, \ell_{m+2}, ..., \ell_{2m}$. Importantly, no logs are repeated.
3. If the timestamps cannot be changed and no logs can be changed, then it is not possible for fetch $n+2$ to contain any of $\ell_1, \ell_2, ..., \ell_m$, as they would have also been included in $n+1$.
4. That means, we only need to cache at most $m$ entries.

vgrozdanic: Yeah, makes sense, i'll update the code

src/commands/logs/list.rs
```diff
@@ -143,10 +169,290 @@ fn execute_single_fetch(
...
+fn execute_live_streaming(
+    api: &Api,
+    org: &str,
+    project: Option<&str>,
+    query: &str,
+    fields: &[&str],
+    args: &ListLogsArgs,
+) -> Result<()> {
+    let mut deduplicator =
+        LogDeduplicator::new(NonZero::new(args.max_rows).expect("max-rows should be non-zero"));
```
szokeasaurusrex: It would be ideal to update the `max_rows` field in the struct to also have a `NonZeroUsize` type. That way, we don't need the `expect`, because the nonzero value is enforced through the type system. But, we can do that in a future PR.

vgrozdanic: i have updated it as a part of this PR

PR #2667, opened 2025-07-31
feat(launchpad): Add asset catalog files to zip without adding to folder

noahsmartin:
Previously the assets were written to the folder and then all the files in the folder were zipped. However, this meant the original location the user passed to the CLI had extra files in it after running the command, which is confusing. This changes it to write the assets to a temporary location, and then add to the zip from the temporary location instead.


trevor-e (approved):
The Swift changes LGTM

src/commands/mobile_app/upload.rs
```diff
@@ -276,38 +271,45 @@ fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
     Ok(temp_file)
 }
 
+fn sort_entries(path: &Path) -> Result<std::vec::IntoIter<(PathBuf, PathBuf)>> {
```
szokeasaurusrex: You should return an `impl Iterator` here; that way, if the concrete return type ever changes (which is quite common with small changes in iterators), the function return type can still remain the same.

```suggestion
fn sort_entries(path: &Path) -> Result<impl Iterator<Item = (PathBuf, PathBuf)>> {
```

szokeasaurusrex: Also, I think it would make sense to move this function (and probably many of the other functions in this file) into the `utils`, as I could see these functions being used elsewhere in the codebase, e.g. if we add new mobile-app commands. However, this can be done in a separate PR.

src/commands/mobile_app/upload.rs
```diff
@@ -276,38 +271,45 @@ fn normalize_file(path: &Path, bytes: &[u8]) -> Result<TempFile> {
...
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    if is_apple_app(path) {
+        handle_asset_catalogs(path, temp_dir.path());
+    }
+    normalize_directory(path, temp_dir.path())
+}
+
 // For XCArchive directories, we'll zip the entire directory
-fn normalize_directory(path: &Path) -> Result<TempFile> {
+fn normalize_directory(path: &Path, parsed_assets_path: &Path) -> Result<TempFile> {
```
szokeasaurusrex: I am a bit confused by how the behavior is changing here.

If I understand correctly, the `parsed_assets_path` may not exist, in which case `normalize_directory` behaves the same as it did before. Otherwise, we do something with the `parsed_assets_path`.

If my understanding is correct, I think the code would be easier to understand and maintain if instead, we had a `normalize_directory` function which takes only the single `path` parameter. Then, we could have a separate function which also handles the `parsed_assets_path` – that function would internally call the `normalize_directory` function and the `parsed_assets_path` on this new function would always exist. This would avoid the branching we have in this function as modified in this PR

tests/integration/mobile_app/upload.rs
```diff
@@ -92,6 +92,25 @@ fn command_mobile_app_upload_apk_all_uploaded() {
         .with_default_token();
 }
 
+#[test]
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+fn command_mobile_app_upload_xcarchive_all_uploaded() {
```
szokeasaurusrex: What exactly is this test meant to be checking, and how does it relate to the changes in this PR?

As far as I can tell, this simply checks that the CLI recognizes when nothing needs to be uploaded. I am unsure how the test relates to the functionality introduced in this PR, it seems independent

noahsmartin: Yes good catch, this wasn't needed, removed

szokeasaurusrex (approved):

PR #2683, opened 2025-08-07
chore(launchpad): Remove assembly polling for mobile app + print artifact id if present

NicoHinderling:
In https://github.com/getsentry/sentry/pull/97435 we make a change to simplify the uploading which depends on us no longer worrying about the assembly status. The user will be able to follow up via the website / github hooks if there are any issues with the upload further down the road


rbro112 (approved):

src/api/data_types/chunking/mobile_app.rs
```diff
@@ -17,7 +15,6 @@ pub struct ChunkedMobileAppRequest<'a> {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AssembleMobileAppResponse {
-    pub state: ChunkedFileState,
     pub missing_chunks: Vec<Digest>,
-    pub detail: Option<String>,
```
NicoHinderling, line 22: we aren't using these two fields anymore

src/api/data_types/chunking/mobile_app.rs
```diff
@@ -17,7 +15,6 @@ pub struct ChunkedMobileAppRequest<'a> {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AssembleMobileAppResponse {
-    pub state: ChunkedFileState,
```
szokeasaurusrex, line 20: I doubt we want to be ignoring this, as the `state` can indicate an error

src/commands/mobile_app/upload.rs
```diff
@@ -399,61 +406,7 @@ fn upload_file(
...
-    }
-
-    if response.state.is_pending() {
-        info!("File upload complete (processing pending on server)");
-    } else {
-        info!("File processing complete");
-    }
-
-    Ok(())
+    Ok(response.artifact_id)
```
szokeasaurusrex, line 409: You need to make a call to the assemble endpoint after uploading the chunks, so that the server begins the assembly process. I don't expect these changes to work, as the code is currently written.

NicoHinderling:
Going to merge these changes into https://github.com/getsentry/sentry-cli/pull/2675

PR #2688, opened 2025-08-08
feat(logs): allow project slugs in logs list

shellmayr:
- Allow project slugs & continue allowing project ids for project parameter in logs list

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1454,7 +1454,7 @@ pub struct FetchEventsOptions<'a> {
     /// Fields to include in the response
     pub fields: &'a [&'a str],
     /// Project ID to filter events by
-    pub project_id: &'a str,
+    pub project: Option<&'a str>,
     /// Cursor for pagination
     pub cursor: Option<&'a str>,
     /// Query string to filter events
@@ -1480,8 +1480,14 @@ impl<'a> FetchEventsOptions<'a> {
             params.push(format!("cursor={}", QueryArg(cursor)));
         }
 
-        params.push(format!("project={}", QueryArg(self.project_id)));
-        params.push(format!("query={}", QueryArg(self.query)));
+        if let Some(project) = self.project {
+            if !project.is_empty() {
+                params.push(format!("project={}", QueryArg(project)));
+            }
+        }
+        if !self.query.is_empty() {
+            params.push(format!("query={}", QueryArg(self.query)));
+        }
         params.push(format!("per_page={}", self.per_page));
         params.push(format!("statsPeriod={}", QueryArg(self.stats_period)));
         params.push(format!("sort={}", QueryArg(self.sort)));
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1454,7 +1454,7 @@ pub struct FetchEventsOptions<'a> {
     /// Fields to include in the response
     pub fields: &'a [&'a str],
     /// Project ID to filter events by
-    pub project: Option<&'a str>,
+    pub project_id: Option<&'a str>,
     /// Cursor for pagination
     pub cursor: Option<&'a str>,
     /// Query string to filter events
@@ -1480,7 +1480,7 @@ impl<'a> FetchEventsOptions<'a> {
             params.push(format!("cursor={}", QueryArg(cursor)));
         }
 
-        if let Some(project) = self.project {
+        if let Some(project) = self.project_id {
             if !project.is_empty() {
                 params.push(format!("project={}", QueryArg(project)));
             }
--- a/src/commands/logs/list.rs
+++ b/src/commands/logs/list.rs
@@ -142,3 +142,34 @@ fn execute_single_fetch(
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_is_numeric_project_id_purely_numeric() {
+        assert!(is_numeric_project_id("123456"));
+        assert!(is_numeric_project_id("1"));
+        assert!(is_numeric_project_id("999999999"));
+    }
+
+    #[test]
+    fn test_is_numeric_project_id_alphanumeric() {
+        assert!(!is_numeric_project_id("abc123"));
+        assert!(!is_numeric_project_id("123abc"));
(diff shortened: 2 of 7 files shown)


src/api/mod.rs
```diff
@@ -1480,8 +1480,14 @@ impl<'a> FetchEventsOptions<'a> {
             params.push(format!("cursor={}", QueryArg(cursor)));
         }
 
-        params.push(format!("project={}", QueryArg(self.project_id)));
-        params.push(format!("query={}", QueryArg(self.query)));
+        if let Some(project) = self.project {
+            if !project.is_empty() {
+                params.push(format!("project={}", QueryArg(project)));
+            }
+        }
```
szokeasaurusrex: Is there any semantic difference between a `self.project` value of `None` and `Some("")`? If not, I would suggest that we make project non-optional, as originally, and just use the empty string to indicate a missing project.

src/commands/logs/list.rs
```diff
@@ -37,7 +42,7 @@ pub(super) struct ListLogsArgs {
     org: Option<String>,
 
     #[arg(short = 'p', long = "project")]
-    #[arg(help = "The project ID (slug not supported).")]
+    #[arg(help = "The project ID (numeric) or project slug (string).")]
```
szokeasaurusrex: For consistency with other commands:

```suggestion
    #[arg(help = "The project ID or slug.")]
```

src/commands/logs/list.rs
```diff
@@ -70,29 +75,41 @@ pub(super) fn execute(args: ListLogsArgs) -> Result<()> {
...
     } else {
-        Some(args.query.as_str())
+        // For project slugs, include in query string
+        let query = if args.query.is_empty() {
+            format!("project:{project}")
+        } else {
+            format!("project:{project} {}", args.query)
+        };
+        (query, None)
     };
```
szokeasaurusrex: You can greatly simplify the `if` block by using `Cow`, which is a smart-pointer which can store either a borrowed type (`&String` in this case) or an owned type (`String`).

```suggestion
    let (query, project_id) = if is_numeric_project_id(project) {
        // For numeric project IDs, pass as project parameter
        (Cow::Borrowed(&args.query), Some(project.as_str()))
    } else {
        // For project slugs, include in query string
        let query = if args.query.is_empty() {
            format!("project:{project}")
        } else {
            format!("project:{project} {}", args.query)
        };
        (Cow::Owned(query), None)
    };
```

You need to also import `std::borrow::Cow` at the beginning of the file.

src/commands/logs/list.rs
```diff
@@ -70,29 +75,41 @@ pub(super) fn execute(args: ListLogsArgs) -> Result<()> {
...
+    execute_single_fetch(&api, org, &query, project_id, LOG_FIELDS, &args)
 }
 
 fn execute_single_fetch(
     api: &Api,
     org: &str,
-    project: &str,
-    query: Option<&str>,
+    query: &str,
+    project_id: Option<&str>,
```
szokeasaurusrex: [nit/optional] to keep the diff simpler

```suggestion
    project_id: Option<&str>,
    query: &str,
```

src/api/mod.rs
```diff
@@ -1454,7 +1454,7 @@ pub struct FetchEventsOptions<'a> {
     /// Fields to include in the response
     pub fields: &'a [&'a str],
     /// Project ID to filter events by
-    pub project_id: &'a str,
+    pub project: Option<&'a str>,
```
szokeasaurusrex: As we only accept project IDs here, I don't think the rename makes sense.

```suggestion
    pub project_id: Option<&'a str>,
```

src/commands/logs/list.rs
shellmayr: nice, TIL!

src/commands/logs/list.rs
```diff
@@ -20,6 +22,11 @@ fn validate_max_rows(s: &str) -> Result<usize> {
     }
 }
 
+/// Check if a project identifier is numeric (project ID) or string (project slug)
+fn is_numeric_project_id(project: &str) -> bool {
+    project.chars().all(|c| c.is_ascii_digit())
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: this may be a valid concern, at least the part about an empty project. We should validate that users cannot provide an empty string for this argument

src/commands/logs/list.rs
```diff
@@ -70,29 +77,36 @@ pub(super) fn execute(args: ListLogsArgs) -> Result<()> {
...
+    let (query, project_id) = if is_numeric_project_id(project) {
+        (Cow::Borrowed(&args.query), Some(project.as_str()))
     } else {
-        Some(args.query.as_str())
+        let query = if args.query.is_empty() {
+            format!("project:{project}")
+        } else {
+            format!("project:{project} {}", args.query)
+        };
+        (Cow::Owned(query), None)
```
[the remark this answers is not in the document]
shellmayr: This case can not happen, as numeric project ids are not allowed.

src/api/mod.rs
shellmayr: Is there a reason we don't want to allow project to be None, and show logs from any project?

src/commands/logs/list.rs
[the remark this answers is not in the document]
shellmayr: Fixed 👍

src/api/mod.rs
szokeasaurusrex: Yes, from a user perspective, I'd definitely like to see a feature where we can show logs from all projects, if no specific project is provided, just like in the Sentry web UI assuming this is possible.

What I am wondering with this comment is whether there is any difference between a project value of `None` versus `Some("")` (i.e. empty string). Right now, this if statement seems to treat both the same way. All I am saying is that, if empty string is the same as `None`, we can just make the project's type a simple `String` rather than `Option<String>` and only send the project to Sentry if it is non-empty.

Alternatively, we keep it as `Option<String>`, and assume that any `Some` value is non-empty.

In any case, this is not a major concern, I was just wondering if there is some semantic difference between the two cases

shellmayr: At the moment - no real semantic difference, just wondering about the desired behaviour here

PR #2693, opened 2025-08-08
Improve `--no-upload` flag for `sentry-cli dif upload`

P1n3appl3:
According to the docs, `--no-upload` is supposed to "run all steps for the processing", but currently it exits early and doesn't actually do any searching for debug files.

This PR defers the exit until _right_ before files are actually uploaded, which makes the command actually output helpful information for troubleshooting.


loewenheim (approved):

szokeasaurusrex (approved):
This is a breaking change, as the change moves the check for `--no-upload` after the request to the chunk upload endpoint, which requires a valid auth token, that we did not need previously. 

However, given that the `--no-upload` help text states that we run all processing steps, I believe it is reasonable that we handle this as a bug fix

PR #2697, opened 2025-08-11
feat(preprod): use artifact url from response

NicoHinderling:
I thought I pushed these changes to https://github.com/getsentry/sentry-cli/pull/2675 on sunday but I must have only committed them to my local

### DIFF
--- a/src/api/data_types/chunking/mobile_app.rs
+++ b/src/api/data_types/chunking/mobile_app.rs
@@ -20,5 +20,5 @@ pub struct AssembleMobileAppResponse {
     pub state: ChunkedFileState,
     pub missing_chunks: Vec<Digest>,
     pub detail: Option<String>,
-    pub artifact_id: Option<String>,
+    pub artifact_url: Option<String>,
 }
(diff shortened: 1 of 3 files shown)


szokeasaurusrex (approved):
small nit: I would rename the PR to `ref(preprod): Rename variables to indicate they store URLs`.

From what I can tell, this PR is a refactor, and would not actually change any behavior. The current title makes it seem otherwise.

src/api/data_types/chunking/mobile_app.rs
```diff
@@ -20,5 +20,5 @@ pub struct AssembleMobileAppResponse {
     pub state: ChunkedFileState,
     pub missing_chunks: Vec<Digest>,
     pub detail: Option<String>,
-    pub artifact_id: Option<String>,
+    pub artifact_url: Option<String>,
```
szokeasaurusrex, line 23: Ah wait, this produces a behavior change. We now are trying to deserialize a field named `artifact_url` instead of `artifact_id`

szokeasaurusrex:
CI is failing, likely because you have to update the mocks to return the `artifact_url` in addition to, or instead of the `artifact_id`

szokeasaurusrex:
Closed in favor of #2698

PR #2706, opened 2025-08-12
feat(mobile-app): Add default vcs base_ref parsing for mobile-app subcommand

rbro112:
Adds default handling of `head_ref` (branch name) to the `mobile-app upload` subcommand. This will leverage the `head.shorthand()` value if no value is provided by the user.

Also handles the detached HEAD state with some logging that will require the user to specify using the command arguments.

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -140,8 +140,23 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             .map(Cow::Borrowed)
             .or_else(|| {
                 // Try to get the current ref from the VCS if not provided
+                // Note: git_repo_head_ref will return an error for detached HEAD states,
+                // which .ok() converts to None - this prevents sending "HEAD" as a branch name
+                // In that case, the user will need to provide a valid branch name.
                 repo_ref
-                    .and_then(|r| git_repo_head_ref(r).ok())
+                    .and_then(|r| match git_repo_head_ref(r) {
+                        Ok(ref_name) => {
+                            debug!("Found current branch reference: {}", ref_name);
+                            Some(ref_name)
+                        }
+                        Err(e) => {
+                            debug!(
+                                "No valid branch reference found (likely detached HEAD): {}",
+                                e
+                            );
+                            None
+                        }
+                    })
                     .map(Cow::Owned)
             });
 
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -112,7 +112,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         let repo_ref = repo.as_ref();
         let remote_url = repo_ref.and_then(|repo| git_repo_remote_url(repo, &cached_remote).ok());
 
-        let vcs_provider: Option<Cow<'_, str>> = matches
+        let vcs_provider = matches
             .get_one("vcs_provider")
             .map(String::as_str)
             .map(Cow::Borrowed)
@@ -123,7 +123,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
                     .map(Cow::Owned)
             });
 
-        let head_repo_name: Option<Cow<'_, str>> = matches
+        let head_repo_name = matches
             .get_one("head_repo_name")
             .map(String::as_str)
             .map(Cow::Borrowed)
@@ -134,14 +134,14 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
                     .map(Cow::Owned)
             });
 
-        let head_ref: Option<Cow<'_, str>> = matches
+        let head_ref = matches
             .get_one("head_ref")
             .map(String::as_str)
             .map(Cow::Borrowed)
             .or_else(|| {
                 // Try to get the current ref from the VCS if not provided
                 // Note: git_repo_head_ref will return an error for detached HEAD states,
-                // which .ok() converts to None - this prevents sending "HEAD" as a branch name
+                // which the error handling converts to None - this prevents sending "HEAD" as a branch name
                 // In that case, the user will need to provide a valid branch name.
                 repo_ref
                     .and_then(|r| match git_repo_head_ref(r) {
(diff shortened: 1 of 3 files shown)


rbro112:
* **#2706** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2706?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

szokeasaurusrex:
Deferring review until after HackWeek – please ping me if you need a review more urgently

src/commands/mobile_app/upload.rs
```diff
@@ -134,13 +135,37 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
                     .map(Cow::Owned)
             });
 
-        (vcs_provider, head_repo_name)
+        let head_ref: Option<Cow<'_, str>> = matches
```
szokeasaurusrex: Type annotation unnecessary here. 
```suggestion
        let head_ref = matches
```

szokeasaurusrex: I have noticed Cursor tends to add unnecessary type annotations quite frequently, and sadly it is not possible to lint against them when the type involves generics (as it does here), so I have tried adding a Cursor rule in https://github.com/getsentry/sentry-cli/pull/2717. 

Hopefully that helps going forward!

src/commands/mobile_app/upload.rs
```diff
@@ -134,13 +135,37 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+            .get_one("head_ref")
+            .map(String::as_str)
+            .map(Cow::Borrowed)
+            .or_else(|| {
+                // Try to get the current ref from the VCS if not provided
+                // Note: git_repo_head_ref will return an error for detached HEAD states,
+                // which .ok() converts to None - this prevents sending "HEAD" as a branch name
+                // In that case, the user will need to provide a valid branch name.
```
szokeasaurusrex: This comment appears to be outdated, as we are no longer calling `.ok()` on `None` here

src/utils/vcs.rs
```diff
@@ -234,6 +234,24 @@ pub fn git_repo_remote_url(
         .ok_or_else(|| git2::Error::from_str("No remote URL found"))
 }
 
+#[cfg(feature = "unstable-mobile-app")]
+pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String, git2::Error> {
```
szokeasaurusrex: We should just use the normal `anyhow::Result` here

```suggestion
pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
```

src/utils/vcs.rs
```diff
@@ -234,6 +234,24 @@ pub fn git_repo_remote_url(
...
+    let head = repo.head()?;
+
+    // Only return a reference name if we're not in a detached HEAD state
+    // In detached HEAD state, head.shorthand() returns "HEAD" which is not a valid branch name
+    if head.is_branch() {
+        head.shorthand()
+            .map(|s| s.to_owned())
+            .ok_or_else(|| git2::Error::from_str("No HEAD reference found"))
```
```suggestion
            .ok_or_else(|| anyhow::anyhow!("No HEAD reference found"))
```

src/utils/vcs.rs
```diff
@@ -234,6 +234,24 @@ pub fn git_repo_remote_url(
...
+    } else {
+        // In detached HEAD state, return an error to indicate no valid branch reference
+        Err(git2::Error::from_str(
+            "HEAD is detached - no branch reference available",
+        ))
```
```suggestion
        Err(anyhow::anyhow!(
            "HEAD is detached - no branch reference available",
        ))
```

szokeasaurusrex (commented):
Some minor comments, otherwise lgtm!

src/commands/mobile_app/upload.rs
rbro112: Ha, funny enough this was manual. Removed!

src/utils/vcs.rs
```diff
@@ -1190,3 +1208,32 @@ fn test_generate_patch_ignore_missing() {
...
+        "foo.js",
+        b"console.log(\"Hello, world!\");",
+        "\"initial commit\"",
+    );
+
+    let repo = git2::Repository::open(dir.path()).expect("Failed");
+
+    // Test on a branch (should succeed)
+    let head_ref = git_repo_head_ref(&repo).expect("Should get branch reference");
+    assert_eq!(head_ref, "master");
```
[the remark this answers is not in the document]
szokeasaurusrex: @rbro112 this is a legit concern, ~~as we may want to rename the "master" branch to "main" in the (near) future~~

Nvm the strike through part, I see now that we are creating a new repo. Still I think the comment is valid, as it still feels like we are asserting internal Git2 behavior here. Would be better to initialize the repo with a specific default branch name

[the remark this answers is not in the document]
rbro112: I went ahead and added some code above to default the main branch to "main" but I do feel this is a bit beyond the scope of my change as these tests run within the scope of `sentry-cli`, they use the default branch naming (which is `master`). But agreed that these should run successfully agnostic of env they run in.

src/commands/build/upload.rs
```diff
@@ -133,13 +134,37 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
-        (vcs_provider, head_repo_name)
+        let head_ref = matches
+            .get_one("head_ref")
+            .map(String::as_str)
+            .map(Cow::Borrowed)
+            .or_else(|| {
+                // Try to get the current ref from the VCS if not provided
+                // Note: git_repo_head_ref will return an error for detached HEAD states,
+                // which the error handling converts to None - this prevents sending "HEAD" as a branch name
+                // In that case, the user will need to provide a valid branch name.
```
szokeasaurusrex, line 145: [question] Does this imply that the `head_ref` is required? From what I can tell right now, we don't validate its presence locally, so if it is required by the server, we would likely get a 4xx response when attempting the upload. If that is the case, we should likely return an error here, rather than swallow the error.

rbro112: It is not required, it is optional on the backend so no response failure if not provided.

PR #2710, opened 2025-08-13
chore(metrics): Remove send_metric command

vgrozdanic:
Metrics have been deprecated on october 7th 2024, and since then it is not possible to send metrics. 
This PR:
- cleans up command `send_metric` which is invalid since last October.
- upgrades `sentry-core` package to the latest version which has deprecated metrics


szokeasaurusrex:
We will defer merging this PR until the next major.

I have linked this PR under #2713, so we remember to merge it when we are releasing the next major. In the meantime, I will mark this PR as a draft

lcian:
Thanks @vgrozdanic!
In general it would be better to just upgrade the `sentry` dependency instead of adding/upgrading the one to `sentry-core`.
If you do it this way, you end up with a mix: https://github.com/getsentry/sentry-cli/blob/8f0cf5ae001d835bc06de488bff6a9dec00c3e78/Cargo.lock#L2786-L2811

In this case it shouldn't cause problems, as we're not using the SDK "as an SDK" (e.g. we're not using the Hub at all), but could cause problems if we need to do so in the future, so let's remember to just upgrade `sentry` and try to avoid adding direct dependencies to the subcrates.

szokeasaurusrex:
To maintain a cleaner PR list, I'll close this PR for now. It should be easy to re-implement once we are ready to fully remove `send-metric`

PR #2716, opened 2025-08-26
ref(send-event): Introduce constant for max breadcrumbs

szokeasaurusrex:
While investigating https://github.com/getsentry/sentry-cli/issues/2709, I noticed the max breadcrumbs limit was hardcoded in multiple locations. This change extracts the value into a constant, so it is easier to change later, if we ever need to.


szokeasaurusrex:
Yeah, I think that is exactly why I did it this way, as each command should define its limit separately. Also, `bash-hook` is not super maintained (I'd say it's basically soft-deprecated at this point), which is another reason why I did not put too much effort into sharing the constant between the two commands.

PR #2720, opened 2025-08-28
feat(build): Auto-detect base_ref from git merge-base

runningcode:
Auto-detect base reference for build uploads using git merge-base.

Finds the merge-base commit SHA between HEAD and the remote tracking branch (`refs/remotes/{remote}/HEAD`), eliminating the need to manually specify --base-sha for most workflows.

In the future, we will add more checks for common CI providers to automatically get this value as well.

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -177,13 +177,9 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
                 // This attempts to find the merge-base with the remote tracking branch
                 repo_ref
                     .and_then(|r| match git_repo_base_ref(r, &cached_remote) {
-                        Ok(Some(base_ref_name)) => {
+                        Ok(base_ref_name) => {
                             debug!("Found base reference: {}", base_ref_name);
                             Some(base_ref_name)
-                        }
-                        Ok(None) => {
-                            warn!("No base reference found (could not determine merge-base)");
-                            None
                         }
                         Err(e) => {
                             warn!("Could not detect base branch reference: {}", e);
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -249,7 +249,7 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
     }
 }
 
-pub fn git_repo_base_ref(repo: &git2::Repository, remote_name: &str) -> Result<Option<String>> {
+pub fn git_repo_base_ref(repo: &git2::Repository, remote_name: &str) -> Result<String> {
     // Get the current HEAD commit
     let head_commit = repo.head()?.peel_to_commit()?;
 
@@ -270,7 +270,7 @@ fn find_merge_base_ref(
     repo: &git2::Repository,
     head_commit: &git2::Commit,
     remote_ref: &git2::Reference,
-) -> Result<Option<String>> {
+) -> Result<String> {
     let remote_commit = remote_ref.peel_to_commit()?;
     let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
 
@@ -280,7 +280,7 @@ fn find_merge_base_ref(
         "Found merge-base commit as base reference: {}",
         merge_base_sha
     );
-    Ok(Some(merge_base_sha))
+    Ok(merge_base_sha)
 }
 
 fn find_reference_url(repo: &str, repos: &[Repo]) -> Result<Option<String>> {


src/commands/build/upload.rs
```diff
@@ -172,7 +173,28 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let base_repo_name = matches.get_one("base_repo_name").map(String::as_str);
     let base_sha = matches.get_one("base_sha").map(String::as_str);
-    let base_ref = matches.get_one("base_ref").map(String::as_str);
+    let base_ref = matches
```
rbro112: I'd recommend moving these up into the `let` block where we get `vcs_provider`, `head_repo_name` and `head_ref` (line 117 here). That way we get some safe error handling and can re-use the repo and other variables

src/utils/vcs.rs
```diff
@@ -249,6 +249,54 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+            if branch_commit.id() == merge_base_oid {
+                if let Some(branch_name) = branch.name()? {
+                    return Ok(branch_name.to_owned());
+                }
+            }
+        }
+    }
+
+    // If no branch name found, return the commit SHA
+    Ok(merge_base_oid.to_string())
```
rbro112: I don't think this is the behavior we want, I think we just want branch name, otherwise return `None`.

runningcode: Fixed!

src/commands/build/upload.rs
runningcode: Good idea!

src/commands/build/upload.rs
```diff
@@ -167,12 +168,36 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                            debug!("Found base branch reference: {}", base_ref_name);
+                            Some(base_ref_name)
+                        }
+                        Ok(None) => {
+                            debug!("No base branch reference found (no local branch points to merge-base)");
+                            None
+                        }
+                        Err(e) => {
+                            debug!("Error getting base branch reference: {}", e);
+                            None
```
szokeasaurusrex: Why do we swallow this error without propagating it to the user? Seems like potentially undesired behavior, as if something is broken, the user will not be made aware, unless they have debug logging turned on (which is unlikely in most use cases).

src/utils/vcs.rs
```diff
@@ -249,6 +249,54 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+            // First try common default branch names
+            for branch in &["main", "master", "develop"] {
+                let remote_branch = format!("refs/remotes/{remote_name}/{branch}");
+                if let Ok(r) = repo.find_reference(&remote_branch) {
+                    return find_merge_base_ref(repo, &head_commit, &r);
+                }
+            }
+            bail!("Could not find remote tracking branch for {}", remote_name);
+        }
+    };
```
szokeasaurusrex: You can greatly simplify this as follows:

```suggestion
    let remote_ref = repo.find_reference(&remote_branch_name).or_else(|_| {
        // If remote/HEAD doesn't exist, try to find the default branch
        // First try common default branch names
        ["main", "master", "develop"]
            .iter()
            .map(|branch| format!("refs/remotes/{remote_name}/{branch}"))
            .filter_map(|remote_branch| repo.find_reference(&remote_branch).ok())
            .next()
            .ok_or_else(|| {
                anyhow::anyhow!("Could not find remote tracking branch for {}", remote_name)
            })
    })?;
```

As a general tip, whenever you find yourself writing a lot of branching code in Rust (e.g. loops and match statements), there is often, but not always, a simpler way to do it with method chaining

src/utils/vcs.rs
```diff
@@ -249,6 +249,54 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+    let head_commit = repo.head()?.peel_to_commit()?;
+
+    // Try to find the remote tracking branch
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = match repo.find_reference(&remote_branch_name) {
+        Ok(r) => r,
+        Err(_) => {
+            // If remote/HEAD doesn't exist, try to find the default branch
+            // First try common default branch names
+            for branch in &["main", "master", "develop"] {
```
szokeasaurusrex: Is there a direct way to detect the default branch, besides just checking against hardcoded common values for them? This implementation seems flimsy and easy to break, e.g. if a repo has all three of these branch names in it.

If it is the best we can do, I guess it is okay, but I'd rather directly check for the default branch than just guess it

szokeasaurusrex (commented):
I have a couple questions about the implementation, plus a suggestion for how to simplify the code a bit

src/commands/build/upload.rs
runningcode: This is a good question. I would expect this to fail if the user isn't using git or has some non default branching strategy.
The question is how do we want to handle this.
1. Throw an error at the CLI level.
2. Show some error message after they've uploaded in the UI if the git information is missing.

I don't have a strong preference for 1 or 2, it just how we want the user experience to be. Throwing an error here might be frustrating if trying to get this working initially but once the integration is working throwing an error would be the easiest way to know that a change broke the integration.

src/utils/vcs.rs
```diff
@@ -249,6 +249,63 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+    // Try to find the remote tracking branch
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = repo
+        .find_reference(&remote_branch_name)
+        .or_else(|_| {
+            // If remote/HEAD doesn't exist, try to query the remote for its actual default branch
+            let mut remote = repo.find_remote(remote_name)?;
+            remote.connect(git2::Direction::Fetch)?;
+            let default_branch_buf = remote.default_branch()?;
+            let default_branch = default_branch_buf.as_str().unwrap();
```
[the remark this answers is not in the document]
runningcode: Can a git branch name even contain invalid utf-8?

src/utils/vcs.rs
runningcode: This is a good point, I looked in to this and found a `git2` function called `default_branch()`.

src/utils/vcs.rs
runningcode: Thank you for the rust tip!

src/utils/vcs.rs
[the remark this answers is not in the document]
szokeasaurusrex: ChatGPT seems to think not, but the fact that the value returned by the `git2` library (which is just Rust bindings for `libgit2`) is not a string and only performs a fallible conversion to `str` makes me think that we should perhaps be cautious. There is probably a reason why `git2` is implemented this way; perhaps in some cases on some platforms a non-UTF-8 value is possible. It does not really cost anything to handle the error gracefully, without panicking, so let's do that!

szokeasaurusrex: This is how you can handle the error gracefully

```suggestion
            let default_branch = default_branch_buf
                .as_str()
                .ok_or_else(|| anyhow::anyhow!("Default branch contains invalid UTF-8"))?;
```

src/utils/vcs.rs
```diff
@@ -249,6 +249,63 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+
+            // Convert "refs/heads/main" to "refs/remotes/origin/main"
+            let branch_name = default_branch
+                .strip_prefix("refs/heads/")
+                .unwrap_or(default_branch);
+            let remote_branch = format!("refs/remotes/{remote_name}/{branch_name}");
+            repo.find_reference(&remote_branch)
```
szokeasaurusrex: As the return type of this closure will now be an anyhow::Error, we need a small refactor here, to convert to the new error type
```suggestion
            Ok(repo.find_reference(&remote_branch)?)
```

src/utils/vcs.rs
```diff
@@ -249,6 +249,63 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+        })
+        .map_err(|e| {
```
szokeasaurusrex: And, we also need a type annotation here, as the return value of the previous closure can no longer be inferred

```suggestion
        .map_err(|e: Error| {
```

src/utils/vcs.rs
runningcode: done!

src/utils/vcs.rs
runningcode: done!

src/utils/vcs.rs
runningcode: done!

runningcode:
Thanks for the suggestions @szokeasaurusrex ! I applied them.

src/utils/vcs.rs
```diff
@@ -249,6 +249,66 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+) -> Result<Option<String>> {
+    let remote_commit = remote_ref.peel_to_commit()?;
+    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
+
+    // Try to find a branch name that points to this commit
+    let branches = repo.branches(Some(git2::BranchType::Local))?;
+    for (branch, _) in branches.flatten() {
+        if let Ok(branch_commit) = branch.get().peel_to_commit() {
+            if branch_commit.id() == merge_base_oid {
+                if let Some(branch_name) = branch.name()? {
```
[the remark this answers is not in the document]
runningcode: This is a valid point but I think its quite rare and not worth making the code path more complex to handle this. WDYT @szokeasaurusrex ?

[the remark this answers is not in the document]
szokeasaurusrex: I agree, although I would also mention that I think this code could be made more readable (and possibly also, it would be easier to handle this case properly), if we were to use chained iterator methods instead of a for loop with a bunch of `if` conditions below. I won't block this PR over this, but it is something to consider.

**More concerning though:** why do we even need to iterate over all branches in the first place? Is there no way to directly query the branches associated with a given commit? I think the loop could be inefficient for a repo with many local branches

If this type of query is not possible, I guess it is still ok

runningcode:
@szokeasaurusrex I couldn’t reply directly to [your comment](https://github.com/getsentry/sentry-cli/pull/2720#discussion_r2319017985) but that’s a good catch. I’m mostly relying on AI here.
It seems that `git2/libgit2` doesn't provide a direct way to query "which branches point to this specific commit”.

Alternatively: we could call out to a shell with `git branch --points-at`?

szokeasaurusrex:
Let's avoid calling git directly, as adding a dependency on git is potentially a breaking change. 

Idk how git works internally, but maybe it's possible that git itself only links from branches to commits, and not the other way around. In which case, it would make no difference that we are manually iterating. 

So I'd say let's stick with how we have the code now. It would likely be preferable to write as an iterator chain instead of a loop though (I can also perform the rewrite for you)

runningcode:
Sounds good! I’ve fixed the tests and used an iterator chain.

src/utils/vcs.rs
```diff
@@ -249,6 +249,67 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
 
+pub fn git_repo_base_ref(repo: &git2::Repository, remote_name: &str) -> Result<Option<String>> {
+    // Get the current HEAD commit
+    let head_commit = repo.head()?.peel_to_commit()?;
+
+    // Try to find the remote tracking branch
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = repo
+        .find_reference(&remote_branch_name)
+        .or_else(|_| -> Result<git2::Reference, anyhow::Error> {
```
szokeasaurusrex: `Result` here is `anyhow::Result` (imported above), as it is in most parts of Sentry CLI. With `anyhow::Result`, the error value is implicitly `anyhow::Error` when not provided. Using `Result<git2::Reference>` is therefore simpler here.

However, in this case, the entire type annotation is unnecessary, as the compiler can infer it:

```suggestion
        .or_else(|_| {
```

src/utils/vcs.rs
```diff
@@ -249,6 +249,67 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+                .ok_or_else(|| anyhow::anyhow!("Default branch contains invalid UTF-8"))?;
+
+            // Convert "refs/heads/main" to "refs/remotes/origin/main"
+            let branch_name = default_branch
+                .strip_prefix("refs/heads/")
+                .unwrap_or(default_branch);
+            let remote_branch = format!("refs/remotes/{remote_name}/{branch_name}");
+            Ok(repo.find_reference(&remote_branch)?)
+        })
+        .map_err(|e: anyhow::Error| {
```
szokeasaurusrex: Seems we don't need this annotation anymore, either 😅

```suggestion
        .map_err(|e| {
```

src/commands/build/upload.rs
runningcode: Actually I found a better way to do this without iterating through the branches. Just updated the logic.

runningcode:
@szokeasaurusrex I’ve simplified the code here. It just tries to get the remote tracking branch as `origin/HEAD`and then finds the merge base with `HEAD`. This is what we previously did at Emerge tools. 

Cursor correctly comments that this won’t work in all cases. We will log a warning when it doesn’t work.
In future PRs, I will add support for common CI providers to calculate the merge base using environment variables on CI, for example `GITHUB_BASE_REF` on github.

src/utils/vcs.rs
```diff
@@ -249,6 +249,40 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+    })?;
+
+    find_merge_base_ref(repo, &head_commit, &remote_ref)
+}
+
+fn find_merge_base_ref(
+    repo: &git2::Repository,
+    head_commit: &git2::Commit,
+    remote_ref: &git2::Reference,
+) -> Result<Option<String>> {
```
szokeasaurusrex: From what I can tell, we are never returning `Ok(None)`, so I think we can make this function return `Result<String>` instead

szokeasaurusrex: Assuming I am correct, we should also be able to make `git_repo_base_ref` return a `Result<String>`

szokeasaurusrex (commented):
This is looking much better! After these changes, though, it looks like we can make some small simplifications of the function signatures

szokeasaurusrex:
@runningcode please update the PR description to reflect the new behavior, e.g. that we are now no longer looking for the `default_branch`. I think this may be why Cursor is complaining (or at least, one of the reasons why)

runningcode:
Updated PR description!

szokeasaurusrex (approved):
Much simpler now, thank you 🚀

PR #2722, opened 2025-09-01
feat(build): Add auto-detection of PR number from GitHub Actions

runningcode:
## Summary
- Implements auto-detection of PR number from GitHub Actions environment variables for the `build upload` command
- When `--pr-number` is not provided, automatically detects from `GITHUB_EVENT_NAME` and `GITHUB_REF`
- Adds comprehensive tests for the detection logic

## Implementation Details
- Checks `GITHUB_EVENT_NAME` == "pull_request" and `GITHUB_REF` starts with "refs/pull/"
- Parses PR number from `GITHUB_REF` format: `refs/pull/{pr_number}/merge`
- Includes debug logging when PR number is auto-detected
- Updated help text to clarify the auto-detection behavior

## Test Plan
- [x] Added unit tests covering various GitHub Actions scenarios
- [x] Verified compilation and existing tests pass
- [x] Tested help text updates
- [x] Manual testing with mock environment variables

Resolves Linear issue EME-233

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -24,8 +24,8 @@ use crate::utils::fs::TempDir;
 use crate::utils::fs::TempFile;
 use crate::utils::progress::ProgressBar;
 use crate::utils::vcs::{
-    self, get_github_pr_number, get_provider_from_remote, get_repo_from_remote, git_repo_base_ref, git_repo_head_ref,
-    git_repo_remote_url,
+    self, get_github_pr_number, get_provider_from_remote, get_repo_from_remote, git_repo_base_ref,
+    git_repo_head_ref, git_repo_remote_url,
 };
 
 pub fn make_command(command: Command) -> Command {


src/commands/build/upload.rs
```diff
@@ -85,7 +105,7 @@ pub fn make_command(command: Command) -> Command {
             Arg::new("pr_number")
                 .long("pr-number")
                 .value_parser(clap::value_parser!(u32))
-                .help("The pull request number to use for the upload. If not provided, the current pull request number will be used.")
+                .help("The pull request number to use for the upload. If not provided, the PR number will be automatically detected from GitHub Actions environment variables.")
```
rbro112: I think a qualifier to indicate "if running in a GitHub environment" would be good to add.

src/commands/build/upload.rs
```diff
@@ -27,6 +27,26 @@ use crate::utils::vcs::{
     self, get_provider_from_remote, get_repo_from_remote, git_repo_head_ref, git_repo_remote_url,
 };
 
+fn get_default_pr_number() -> Option<u32> {
```
rbro112: Nit: `get_github_pr_number` to indicate this is just for GitHub.

Might also be good to move into `vcs.rs` or `github.rs` where `vcs.rs` lives

src/commands/build/upload.rs
```diff
@@ -27,6 +27,26 @@ use crate::utils::vcs::{
...
+    std::env::var("GITHUB_REF").ok().and_then(|github_ref| {
```
rbro112: Would recommend avoiding all the complicated nesting here, makes it pretty tough to read. I'm not super experienced with rust but I think this would have the same functionality and is alot more readable IMO

```rust
let github_ref = std::env::var("GITHUB_REF").ok()?;
let event_name = std::env::var("GITHUB_EVENT_NAME").ok()?;
    
if event_name != "pull_request" || !github_ref.starts_with("refs/pull/") {
   return None;
}
    
let pr_number_str = github_ref
  .strip_prefix("refs/pull/")?
  .split('/')
  .next()?;
    
let pr_number = pr_number_str.parse::<u32>().ok()?;
debug!("Auto-detected PR number from GitHub Actions: {}", pr_number);
Some(pr_number)
```

src/commands/build/upload.rs
runningcode: done

src/commands/build/upload.rs
runningcode: done

src/commands/build/upload.rs
runningcode: done

src/utils/vcs.rs
```diff
@@ -249,6 +249,23 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
     }
 }
 
+/// Attempts to get the PR number from GitHub Actions environment variables.
+/// Returns the PR number if running in a GitHub Actions pull request environment.
+pub fn get_github_pr_number() -> Option<u32> {
```
runningcode, line 288: For this function, I didn't log a warning if we failed to fetch here because i figure there are so many different CI providers we'd be logging a warning every time. Until we cover a broader range of CI providers it doesn't make sense to log something. WDYT?

rbro112: Debug logging should be fine, I'd recommend adding some logs around failure points, particularly in the like 258 block and any other early return points just to make debugging easier.

rbro112 (approved):
All logic for picking up PR number LGTM, deferring rest of review to @szokeasaurusrex for general CLI

src/utils/vcs.rs
szokeasaurusrex: I second this, as debug logs are hidden by default. Default log level is `WARN`. However, when folks report bugs, we ask them to set a log level of `DEBUG`, as the debug logs, though verbose, often help identify the problem.

src/utils/vcs.rs
```diff
@@ -249,6 +249,23 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
+    let github_ref = std::env::var("GITHUB_REF").ok()?;
+    let event_name = std::env::var("GITHUB_EVENT_NAME").ok()?;
+
+    if event_name != "pull_request" || !github_ref.starts_with("refs/pull/") {
+        return None;
+    }
+
+    let pr_number_str = github_ref.strip_prefix("refs/pull/")?.split('/').next()?;
+
+    let pr_number = pr_number_str.parse::<u32>().ok()?;
```
szokeasaurusrex: Compiler can infer the type via the function signature, so we don't need the `::<u32>` here

```suggestion
    let pr_number = pr_number_str.parse().ok()?;
```

src/utils/vcs.rs
```diff
@@ -249,6 +249,23 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
     }
 }
 
+/// Attempts to get the PR number from GitHub Actions environment variables.
+/// Returns the PR number if running in a GitHub Actions pull request environment.
+pub fn get_github_pr_number() -> Option<u32> {
+    let github_ref = std::env::var("GITHUB_REF").ok()?;
+    let event_name = std::env::var("GITHUB_EVENT_NAME").ok()?;
+
+    if event_name != "pull_request" || !github_ref.starts_with("refs/pull/") {
```
szokeasaurusrex: As `strip_prefix` returns `None` when the string does not start with the prefix, and the `?` immediately afterwards causes us to return `None` in that case, explicitly checking `starts_with` here is redundant.

```suggestion
    if event_name != "pull_request" {
```

src/commands/build/upload.rs
```diff
@@ -597,4 +601,25 @@ mod tests {
...
+        assert_eq!(pr_number, None);
+
+        std::env::set_var("GITHUB_EVENT_NAME", "pull_request");
+        std::env::set_var("GITHUB_REF", "refs/heads/main");
+        let pr_number = get_github_pr_number();
+        assert_eq!(pr_number, None);
+
+        std::env::remove_var("GITHUB_EVENT_NAME");
+        std::env::remove_var("GITHUB_REF");
+    }
```
szokeasaurusrex: These tests belong in `vcs.rs`, as `get_github_pr_number` is defined there

src/commands/build/upload.rs
```diff
@@ -85,7 +86,7 @@ pub fn make_command(command: Command) -> Command {
             Arg::new("pr_number")
                 .long("pr-number")
                 .value_parser(clap::value_parser!(u32))
-                .help("The pull request number to use for the upload. If not provided, the current pull request number will be used.")
+                .help("The pull request number to use for the upload. If not provided and running in a GitHub Actions environment, the PR number will be automatically detected from GitHub Actions environment variables.")
```
szokeasaurusrex: I would update this help text to clarify that the PR number is only detected when we are in a pull_request-triggered workflow. Also, as this string is a bit long, you should split it across lines, kinda like this (sadly, we would need a nightly version of `rustfmt` to do this automatically):


```suggestion
                .help("The pull request number to use for the upload. If not provided and running  \
                    in a GitHub Actions environment, the PR number will be automatically detected from a \
                    GitHub Actions environment variables.")
```

When you end the line in a `\`, Rust ignores the following newline and also all whitespace characters until the first non-whitespace character on the following line

src/utils/vcs.rs
szokeasaurusrex: Also, how would this work if someone is using a reusable workflow, called within a `pull_request` workflow? 

For instance, Sentry CLI's own [CI workflow](https://github.com/getsentry/sentry-cli/blob/aecb258ad2e12e5b3bc6210912908a7b9e058e55/.github/workflows/ci.yml), which runs on pull requests, kicks off several reusable workflows, such as the [Lint workflow](https://github.com/getsentry/sentry-cli/blob/aecb258ad2e12e5b3bc6210912908a7b9e058e55/.github/workflows/lint.yml). The Lint workflow's triggger is `workflow_call`. Within that workflow, would the `GITHUB_EVENT_NAME` be `workflow_call` or `pull_request`?

runningcode: This is a good point. I don’t think we should handle re-usable workflows. For those cases users can specify the PR number automatically.

szokeasaurusrex: ok! I suppose we could always add support later if needed

src/commands/build/upload.rs
```diff
@@ -173,7 +176,10 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
     let base_repo_name = matches.get_one("base_repo_name").map(String::as_str);
     let base_sha = matches.get_one("base_sha").map(String::as_str);
     let base_ref = matches.get_one("base_ref").map(String::as_str);
-    let pr_number = matches.get_one::<u32>("pr_number");
+    let pr_number = matches
+        .get_one::<u32>("pr_number")
```
szokeasaurusrex:
```suggestion
        .get_one("pr_number")
```

src/utils/vcs.rs
```diff
@@ -1244,3 +1266,20 @@ fn test_git_repo_head_ref() {
...
+    std::env::set_var("GITHUB_EVENT_NAME", "push");
+    let pr_number = get_github_pr_number();
+    assert_eq!(pr_number, None);
+    std::env::set_var("GITHUB_EVENT_NAME", "pull_request");
+    std::env::set_var("GITHUB_REF", "refs/heads/main");
+    let pr_number = get_github_pr_number();
+    assert_eq!(pr_number, None);
+    std::env::remove_var("GITHUB_EVENT_NAME");
+    std::env::remove_var("GITHUB_REF");
+}
```
szokeasaurusrex: This should go inside a `mod test {}` block marked with `#[cfg(test)]`; otherwise, I believe it gets compiled into the release binary.

Please read [Chapter 11.3 of _The Rust Programming Language_](https://doc.rust-lang.org/book/ch11-03-test-organization.html) for more info on test organization

szokeasaurusrex (commented):
some minor things, looks good otherwise

src/utils/vcs.rs
runningcode: Thank you! I’m very much still learning rust. 
I see there are some other tests in here, I didn’t want to adjust them in this PR, but would it make sense to also put those inside `mod test {}` and `#[cfg(test)]`.
I did see that chapter explained what the `#[cfg(test)]` does. Their example used `mod test {}` but I couldn’t figure out what exactly that did. Do you know?

szokeasaurusrex: `mod test {}` just creates a module called "test". You could call it by some other name but it is convention to call it "test"

szokeasaurusrex: It would make sense to move any other tests in this file under the test module. But, as it is a separate change, please do this in a separate PR 🙏

runningcode: Sounds good. I’ll make a separate PR once this is merged.

runningcode:
Ready for another review!

PR #2723, opened 2025-09-02
ci(release): Build Linux binaries natively

szokeasaurusrex:
This matches the behavior of the build process for macOS and Windows binaries. Previously the Linux binaries were built in a Docker container, but we can likely simplify that process, and build directly on the runner.


szokeasaurusrex:
Closed in favor of #2727

PR #2734, opened 2025-09-09
fix(releases): handle partial SHAs correctly in commit resolution

srest2021:
fixes REPLAY-413

Prevents partial SHAs from being incorrectly padded with trailing zeros when resolving commits. Partial SHAs like `f915d32` are now properly treated as symbolic references and resolved to valid full SHAs by git.

Fixes commit resolution failures that caused "Unable to fetch commits" errors in Sentry when using partial SHAs with `sentry-cli releases set-commits`.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -84,8 +84,13 @@ impl CommitSpec {
     }
 
     pub fn reference(&self) -> GitReference<'_> {
-        if let Ok(oid) = git2::Oid::from_str(&self.rev) {
-            GitReference::Commit(oid)
+        // Only treat as a commit OID if it's a full 40-character SHA
+        if self.rev.len() == 40 && self.rev.chars().all(|c| c.is_ascii_hexdigit()) {
+            if let Ok(oid) = git2::Oid::from_str(&self.rev) {
+                GitReference::Commit(oid)
+            } else {
+                GitReference::Symbolic(&self.rev)
+            }
         } else {
             GitReference::Symbolic(&self.rev)
         }
@@ -93,8 +98,13 @@ impl CommitSpec {
 
     pub fn prev_reference(&self) -> Option<GitReference<'_>> {
         self.prev_rev.as_ref().map(|rev| {
-            if let Ok(oid) = git2::Oid::from_str(rev) {
-                GitReference::Commit(oid)
+            // Only treat as a commit OID if it's a full 40-character SHA
+            if rev.len() == 40 && rev.chars().all(|c| c.is_ascii_hexdigit()) {
+                if let Ok(oid) = git2::Oid::from_str(rev) {
+                    GitReference::Commit(oid)
+                } else {
+                    GitReference::Symbolic(rev)
+                }
             } else {
                 GitReference::Symbolic(rev)
             }
@@ -1320,5 +1330,150 @@ mod tests {
         assert_eq!(pr_number, None);
         std::env::remove_var("GITHUB_EVENT_NAME");
         std::env::remove_var("GITHUB_REF");
+    }
+
+    #[test]
+    fn test_commit_spec_reference_partial_sha() {
+        let spec = CommitSpec {
+            repo: "test-repo".to_string(),
+            path: None,
+            rev: "f915d32".to_string(),
+            prev_rev: None,
+        };
+        
+        match spec.reference() {
+            GitReference::Symbolic(s) => assert_eq!(s, "f915d32"),
+            GitReference::Commit(_) => panic!("Partial SHA should be treated as symbolic reference"),
+        }
+    }
+
+    #[test]
+    fn test_commit_spec_reference_full_sha() {
+        let spec = CommitSpec {
+            repo: "test-repo".to_string(),
+            path: None,
+            rev: "f915d32000000000000000000000000000000000".to_string(),
+            prev_rev: None,
+        };
+        
+        match spec.reference() {
+            GitReference::Commit(oid) => {
+                assert_eq!(oid.to_string(), "f915d32000000000000000000000000000000000");
+            },
+            GitReference::Symbolic(_) => panic!("Full SHA should be treated as commit OID"),
+        }
+    }
+
+    #[test]
+    fn test_commit_spec_reference_non_hex() {
+        let spec = CommitSpec {
+            repo: "test-repo".to_string(),
+            path: None,
+            rev: "HEAD".to_string(),
+            prev_rev: None,
+        };
+        
+        match spec.reference() {
+            GitReference::Symbolic(s) => assert_eq!(s, "HEAD"),
+            GitReference::Commit(_) => panic!("Non-hex string should be treated as symbolic reference"),
+        }
+    }
+
+    #[test]
+    fn test_commit_spec_prev_reference_partial_sha() {
+        let spec = CommitSpec {
+            repo: "test-repo".to_string(),
+            path: None,
+            rev: "HEAD".to_string(),
+            prev_rev: Some("4ebad56".to_string()),
+        };
+        
+        match spec.prev_reference().unwrap() {
+            GitReference::Symbolic(s) => assert_eq!(s, "4ebad56"),
+            GitReference::Commit(_) => panic!("Partial SHA in prev_rev should be treated as symbolic reference"),
+        }
+    }
+
+    #[test]
+    fn test_partial_sha_resolution_with_real_git() {
+        use std::process::Command;
+        use std::fs;
+        use tempfile::TempDir;
+        
+        // Create a temporary git repository
+        let temp_dir = TempDir::new().expect("Failed to create temp dir");
+        let repo_path = temp_dir.path();
+        
+        // Initialize git repo
+        Command::new("git")
+            .args(["init", "--quiet"])
+            .current_dir(repo_path)
+            .status()
+            .expect("Failed to run git init");
+            
+        Command::new("git")
+            .args(["config", "user.name", "Test User"])
+            .current_dir(repo_path)
+            .status()
+            .expect("Failed to set git user name");
+            
+        Command::new("git")
+            .args(["config", "user.email", "test@example.com"])
+            .current_dir(repo_path)
+            .status()
+            .expect("Failed to set git user email");
+        
+        // Create a commit
+        fs::write(repo_path.join("test.txt"), "test content").expect("Failed to write test file");
+        Command::new("git")
+            .args(["add", "test.txt"])
+            .current_dir(repo_path)
+            .status()
+            .expect("Failed to git add");
+            
+        Command::new("git")
+            .args(["commit", "-m", "test commit", "--quiet"])
+            .current_dir(repo_path)
+            .status()
+            .expect("Failed to git commit");
+        
+        // Get the full and short SHA
+        let full_sha_output = Command::new("git")
+            .args(["rev-parse", "HEAD"])
+            .current_dir(repo_path)
+            .output()
+            .expect("Failed to get full SHA");
+        let full_sha = String::from_utf8(full_sha_output.stdout).expect("Invalid UTF-8").trim().to_string();
+        
+        let short_sha_output = Command::new("git")
+            .args(["rev-parse", "--short", "HEAD"])
+            .current_dir(repo_path)
+            .output()
+            .expect("Failed to get short SHA");
+        let short_sha = String::from_utf8(short_sha_output.stdout).expect("Invalid UTF-8").trim().to_string();
+        
+        // Test that partial SHA is treated as symbolic reference
+        let spec = CommitSpec {
+            repo: "test-repo".to_string(),
+            path: Some(repo_path.to_path_buf()),
+            rev: short_sha.clone(),
+            prev_rev: None,
+        };
+        
+        match spec.reference() {
+            GitReference::Symbolic(s) => {
+                assert_eq!(s, short_sha);
+                
+                // Now test that it resolves to the correct full SHA
+                let repo = git2::Repository::open(repo_path).expect("Failed to open git repo");
+                let resolved = repo.revparse_single(&short_sha).expect("Failed to resolve short SHA");
+                let resolved_sha = resolved.id().to_string();
+                
+                assert_eq!(resolved_sha, full_sha, 
+                    "Partial SHA {} should resolve to full SHA {}, but got {}", 
+                    short_sha, full_sha, resolved_sha);
+            },
+            GitReference::Commit(_) => panic!("Partial SHA should be treated as symbolic reference"),
+        }
     }
 }


szokeasaurusrex:
Hey @srest2021, thanks for this PR!

While I agree completely that the current lack of support for full commit hashes is indeed unfortunate and should be addressed, this proposed change has a fundamental flaw: while Sentry CLI will try to resolve the partial commit hashes to a full commit hash in the currently checked-out repository, any repository in the user's Sentry project can be passed in the `--commit` argument. 

For example, let's say I am running `set-commits` like so, from within the **`getsentry/sentry-cli` repository**:

```sh
sentry-cli set-commits my-release --commit "getsentry/sentry@01234...12345"
```

Importantly, note that we passed **`getsentry/sentry`** in the `--commit` argument, **not `getsentry/sentry-cli`**.

Let's say there are two commits in the Sentry CLI repo with hashes starting with 01234 and 12345. With this PR, we will resolve to those Sentry CLI hashes, even though the command is meant to be doing something on the `getsentry/sentry` repo.

### My recommendation

I think we likely should fully rethink how this command works. The current way of specifying commits, including needing to provide the full hash, and also the fact that the repo must be explicitly specified in the `--commit` argument by following the format `<repo>@<full from hash>..<full to hash>`, is quite clunky, and probably should be improved significantly.

I would be happy to discuss ideas for reworking the whole command line interface for this command to improve usability with you and your team.

For now, until we have decided on an improved interface, I think it might be preferable just to validate that the SHAs are full SHAs, and throw an error if not.

Please let me know what you think 🙏

ryan953:
$0.02 opinion here:

It does seem kind of clunky that people have to specify the full `<repo>@<from>..<to>` syntax to make things work. I think there should be two possibilities here for making things easier for people:

1. Validating that a full SHA is passed in could work.
  The `set-commits` command only seems to work with git repos, so we know what length to expect from a SHA. Because it can reference some other repo we won't actually be able to validate that the SHA exists, so it would just mostly be an augmented length/formatting check afaik, yes?

2. Based on the [Issue #43274](https://github.com/getsentry/sentry/issues/43274) it also seems like the API does accept partials SHA's.
  The workaround that suggests people call the API directly explicitly uses partial SHA's, and any git command would work fine too. So if that's the case why are we changing the data that the user gave us at all? If they passed in a partial commit SHA can we simply forward that into the API and not have to reformat the SHA into something else?


Note that: [a partial SHA in git is not a fixed length](https://stackoverflow.com/a/18134919). It must is only long enough to disambiguate a commit within the repo.

srest2021:
> So if that's the case why are we changing the data that the user gave us at all? If they passed in a partial commit SHA can we simply forward that into the API and not have to reformat the SHA into something else?

I'll look into this today and see if it's possible to just preserve the original partial SHA

srest2021:
Using the original unpadded partial SHAs in the API call worked for me! What do you think @szokeasaurusrex ?

srest2021:
@szokeasaurusrex With these changes, will we support commands with only a single partial SHA, like so? 

`sentry-cli releases set-commits --commit srest2021/test-sentry-app@6b06f0f "test-release" --log-level=debug`

Sending this payload to the API doesn't work, so I just wanted to ask if this is expected due to the fact that we can't use the locally checked-out repo anymore:

`{"refs":[{"repository":"srest2021/test-sentry-app","commit":"6b06f0f","previousCommit":null}]}`

szokeasaurusrex:
@srest2021 I'm not sure I completely understand your question. I'd expect that we should still be able to set both SHAs. If we cannot, that's a problem (either in CLI or in the backend). 

Would be happy to take a closer look at this next week, though 🙂

src/commands/releases/set_commits.rs
```diff
@@ -83,14 +79,22 @@ fn strip_sha(sha: &str) -> &str {
         sha
     }
 }
+
+/// Validates that a string is a valid Git SHA (full or partial)
+fn is_valid_sha(sha: &str) -> bool {
+    lazy_static! {
+        static ref SHA_RE: Regex = Regex::new(r"^[a-fA-F0-9]{4,40}$").unwrap();
+    }
```
szokeasaurusrex: `lazy_static!` is a legacy thing; we can use `LazyLock` now, which is built directly into the language. See example [here](https://github.com/getsentry/sentry-cli/blob/40b1f39576843e2b5f66c203274b02c873957fea/src/utils/build/apple.rs#L144-L145).

```suggestion
    static SHA_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[a-fA-F0-9]{4,40}$").expect("Regex is valid"));
```

On another note, though: we probably should not do a strict validation on length (neither max nor min). [Git supports SHA-256](https://git-scm.com/docs/git-init#Documentation/git-init.txt---object-formatformat), which would have a hash of up to 64 characters (which also happens to be what the [Sentry backend enforces](https://github.com/getsentry/sentry/blob/020ae261814a8d4dfda0745c6768f27d7cddb572/src/sentry/constants.py#L43)).

Also, if Sentry supports other version control systems (not sure that we do), those could have arbitrary SHAs, perhaps even non-hexadecimal.

What are your thoughts here? Something we might consider is a soft-validation, where we would just log a warning, but not error, if the SHA is invalid.

szokeasaurusrex:
> Sending this payload to the API doesn't work, so I just wanted to ask if this is expected due to the fact that we can't use the locally checked-out repo anymore:
> 
> `{"refs":[{"repository":"srest2021/test-sentry-app","commit":"6b06f0f","previousCommit":null}]}`

@srest2021, ah I think I misread before. You are saying that the API call did not work with a `previousCommit` being blank?

If that is the case, I think the `previousCommit` is required when you are making the first release for a project. Afterwards, I believe it is optional, since we can get the commit from the previous release. I don't think the changes here should have any affect on that. 

Could this explain the behavior you observed? What exactly went wrong for you?

srest2021:
Hey @szokeasaurusrex sorry for the late update! I'm still trying to figure out what's going on with the API calls not working. I've got a bunch of different test cases that I wanted to use to compare the old vs. new partial behavior, and I'm getting inconsistent results :sob: 

Example: trying to associate a single full SHA with a release:
```
sentry-cli releases new "og-single-full4" --project=javascript-react --log-level=debug --org my-org-slug

sentry-cli releases set-commits --commit srest2021/test-sentry-app@6b06f0ff81cab161e670554caff1e765d171dba5 "og-single-full4" --log-level=debug --org my-org-slug
```

These commands--using the original sentry-cli--worked for me perfectly the first time I tried them and haven't worked since (other results include either associating all 12 commits inside my test repo, or associating 0 commits). This inconsistency in the API (probably due to order of release creation like you mentioned) has made it a bit tricky to clarify the end goal and confirm whether the changes are working or not. But, from what I've observed so far, partial SHAs don't work at all in the old code, and a range of partial SHAs works in this PR. That's why I asked if it's possible for the API to successfully handle a single partial SHA. But, it sounds to me that if we're passing the repo name and a single partial SHA to the API, it *should* be possible for the API to resolve this. 

I'll switch to soft validation and keep working on this--will keep you updated as soon as I make progress! Thanks again for helping out with this issue

**Update**: Yes I believe the problem I was encountering was due to the first vs. non-first release distinction! Sending a single partial SHA now works in this PR as long as it's not the first release.

szokeasaurusrex:
> **Update**: Yes I believe the problem I was encountering was due to the first vs. non-first release distinction! Sending a single partial SHA now works in this PR as long as it's not the first release.

Awesome, glad to hear it @srest2021! So, just to confirm, is everything working as expected now? Is this PR ready for me to review?

srest2021:
Yep should be all good @szokeasaurusrex !

src/commands/releases/set_commits.rs
```diff
@@ -105,30 +110,44 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                    "Warning: Invalid commit SHA '{}'. Only Git SHAs (full or partial) are supported. Proceeding anyway.",
+                    commit_spec.rev
+                );
+            }
+
+            if let Some(ref prev_rev) = commit_spec.prev_rev {
+                if !is_valid_sha(prev_rev) {
+                    eprintln!("Warning: Invalid previous commit SHA '{prev_rev}'. Only Git SHAs (full or partial) are supported. Proceeding anyway.");
                 }
             }
```
szokeasaurusrex: I know I was the one who suggested the soft validation, but upon further reflection, I think we should instead perform a hard validation, which validates the commit **the same way the backend does**.

As the backend only checks the maximum length of **64 characters or less**, this is all we should validate in Sentry CLI. We should not check for any minimum length, nor should we validate that the commit is hexadecimal.

### Reasoning

The reason I prefer the hard validation is that it will allow us to issue a better error message, versus allowing the backend to reject the commit, as the backend error message is harder to understand:

```
+----------------------------+-----------------------------------------------------------------------------+
| Repository                 | Revision                                                                    |
+----------------------------+-----------------------------------------------------------------------------+
| szokeasaurusrex/sentry-cli | aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa -> master |
+----------------------------+-----------------------------------------------------------------------------+
error: API request failed

Caused by:
    sentry reported an error: request failure (http status: 400)
      Object {"refs": Object {"previousCommit": Array [String("Ensure this field has no more than 64 characters.")]}}
```

Also, from my testing, I also discovered that it is possible to pass in branch names instead of commit SHAs. For example, you can use `owner@repo/master~..master` to associate the release with only the latest commit on the `master` branch. This works because the backend does not check that the string is hexadecimal, and instead, it just passes the "commits" to the [repo integration provider](https://github.com/getsentry/sentry/blob/020ae261814a8d4dfda0745c6768f27d7cddb572/src/sentry/tasks/commits.py#L185-L188), which, at least in GitHub's case, appear to be able to resolve branch names.

src/commands/releases/set_commits.rs
```diff
@@ -55,17 +57,13 @@ pub fn make_command(command: Command) -> Command {
...
-                    the current commit of the repository at the given PATH is \
-                    assumed.  To override the revision `@REV` can be appended \
-                    which will force the revision to a certain value."))
+                    The value must be provided as `REPO@SHA` where SHA is a \
+                    Git commit SHA. To specify a range, use `REPO@PREV_SHA..SHA` \
+                    format.\n\n\
+                    Examples:\n\
+                    - `my-repo@abc123` (partial SHA)\n\
+                    - `my-repo@abc123def456789...` (full SHA)\n\
+                    - `my-repo@abc123..def456` (commit range)"))
```
szokeasaurusrex: I would consider updating this to reflect my comment below. Also, for the full sha example, let's please put an actual full-length SHA; the `...` is confusing because we use `..` to specify commit ranges.

Let's also mention that specifying a single commit does not work on the first release.

szokeasaurusrex (commented):
Added a couple small suggestions. But, overall, this is looking pretty good! Thanks for your work on this

src/commands/releases/set_commits.rs
srest2021: Sounds good about the hard validation.

I haven't been able to get the branch names to work on this PR--passing `master~..master` both via the CLI and via a regular API request, I get a 200 status code but the commit count is still 0 for both cases. I made sure I had a correct commit range in the previous release and tried calling the CLI from the target local repo too. And calling the original CLI, which uses find_heads, gives me a resolving error :sob: Would you mind showing how you got your branch names to work? I can include a note about them in the comment but would also be good to specify how/when passing branch names works.

szokeasaurusrex: Honestly, I would avoid documenting that we support branch names.

From how the code written, it seems that this is not an intended feature (many of the variable names, e.g. in the [backend snippet I linked previously](https://github.com/getsentry/sentry/blob/020ae261814a8d4dfda0745c6768f27d7cddb572/src/sentry/tasks/commits.py#L185-L188), refer to SHAs), and also the fact that you are having trouble with it suggests it is brittle at best.

It is hard for me to tell why the branch names did not work for you without knowing exactly what commands you tried to run, but I think as long as commits (partial and full) are working reliably, we can merge this PR.

We can work on better branch name support in the future, if this is a feature we want to properly support.

srest2021: Sounds good, thanks!

src/commands/releases/set_commits.rs
```diff
@@ -55,17 +55,14 @@ pub fn make_command(command: Command) -> Command {
...
-                    current git repository or attempt to auto discover a \
-                    submodule with a compatible URL.\n\n\
-                    The value can be provided as `REPO` in which case sentry-cli \
-                    will auto-discover the commit based on reachable repositories. \
-                    Alternatively it can be provided as `REPO#PATH` in which case \
-                    the current commit of the repository at the given PATH is \
-                    assumed.  To override the revision `@REV` can be appended \
-                    which will force the revision to a certain value."))
+                    The value must be provided as `REPO@SHA` where SHA is \
+                    less than 64 characters. To specify a range, use `REPO@PREV_SHA..SHA` \
```
szokeasaurusrex: Exactly 64 is also okay

```suggestion
                    at most 64 characters. To specify a range, use `REPO@PREV_SHA..SHA` \
```

src/commands/releases/set_commits.rs
```diff
@@ -55,17 +55,14 @@ pub fn make_command(command: Command) -> Command {
...
+                    format.\n\n\
+                    Note: You must specify a previous commit when setting commits for the first release.\n\n\
+                    Examples:\n\
+                    - `my-repo@abc123` (partial SHA)\n\
+                    - `my-repo@abc456def012ghi678jkl234mno890pqr456stu` (full SHA)\n\
+                    - `my-repo@abc123..def456` (commit range)"))
```
szokeasaurusrex: Would add some extra space before the bullet points, so they are indented:

```suggestion
                    Note: You must specify a previous commit when setting commits for the first release.\n\n\
                    Examples:\
                    \n  - `my-repo@abc123` (partial SHA)\
                    \n  - `my-repo@abc456def012ghi678jkl234mno890pqr456stu` (full SHA)\
                    \n  - `my-repo@abc123..def456` (commit range)"))
```

src/commands/releases/set_commits.rs
```diff
@@ -105,30 +102,47 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+            let commit_spec = CommitSpec::parse(spec)?;
+
+            if !repos
+                .iter()
+                .any(|r| r.name.to_lowercase() == commit_spec.repo.to_lowercase())
+            {
+                bail!("Unknown repo '{}'", commit_spec.repo);
+            }
+
+            if commit_spec.rev.len() > 64 {
```
szokeasaurusrex: Let's put `64` into a constant, so we can reuse it here, and also in the check for the prev_rev

src/commands/releases/set_commits.rs
```diff
@@ -105,30 +102,47 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                bail!(
+                    "Invalid commit SHA '{}'. Commit SHAs must be 64 characters or less.",
```
szokeasaurusrex: Likewise, we should interpolate the constant into the error message, here...

src/commands/releases/set_commits.rs
```diff
@@ -105,30 +102,47 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                    commit_spec.rev
+                );
+            }
+
+            if let Some(ref prev_rev) = commit_spec.prev_rev {
+                if prev_rev.len() > 64 {
+                    bail!(
+                        "Invalid previous commit SHA '{}'. Commit SHAs must be 64 characters or less.",
```
szokeasaurusrex: ...and here

szokeasaurusrex (commented):
Getting closer, I have some minor suggestions still

src/constants.rs
```diff
@@ -47,5 +47,7 @@ pub const DEFAULT_MAX_DIF_ITEM_SIZE: u64 = 1024 * 1024; // 1MB
 pub const DEFAULT_MAX_DIF_UPLOAD_SIZE: u64 = 35 * 1024 * 1024; // 35MB
 /// Default maximum time to wait for file assembly.
 pub const DEFAULT_MAX_WAIT: Duration = Duration::from_secs(5 * 60);
+/// Maximum length for commit SHA values.
```
szokeasaurusrex, line 50:
```suggestion
/// Maximum length for commit SHA values, enforced in backend.
```

szokeasaurusrex (approved):
Looks good, please apply the minor suggestion I added prior to merging! Thanks again for improving this 🥔

src/commands/releases/set_commits.rs
```diff
@@ -53,19 +54,16 @@ pub fn make_command(command: Command) -> Command {
...
-                    the current commit of the repository at the given PATH is \
-                    assumed.  To override the revision `@REV` can be appended \
-                    which will force the revision to a certain value."))
+                    The value must be provided as `REPO@SHA` where SHA is \
+                    at most {MAX_COMMIT_SHA_LENGTH} characters. To specify a range, use `REPO@PREV_SHA..SHA` \
+                    format.\n\n\
+                    Note: You must specify a previous commit when setting commits for the first release.\n\n\
+                    Examples:\
+                    \n  - `my-repo@abc123` (partial SHA)\
+                    \n  - `my-repo@abc456def012ghi678jkl234mno890pqr456stu` (full SHA)\
```
srest2021:
```suggestion
                    \n  - `my-repo@62aaca3ed186edc7671b4cca0ab6ec53cb7de8b5` (full SHA)\
```

PR #2735, opened 2025-09-10
feat(build): Add auto-detection of base_repo_name from git remote

runningcode:
## Summary
Implements EME-232 to automatically detect `base_repo_name` for build uploads when not explicitly provided by the user.

## Changes
- **Added `git_repo_base_repo_name()` function** in `src/utils/vcs.rs` that:
  - Fetches the first available git remote from the repository
  - Extracts the repository name using existing `get_repo_from_remote()` function

Closes https://linear.app/getsentry/issue/EME-232/sentry-cli-provide-default-value-for-base-repo-name-if-user-doesnt

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -283,25 +283,34 @@ fn find_merge_base_ref(
     Ok(merge_base_sha)
 }
 
-/// Attempts to get the base repository name from the first available remote.
+/// Attempts to get the base repository name from git remotes.
+/// Prefers "origin" remote if it exists, otherwise uses the first available remote.
 /// Returns the base repository name if a remote is found.
 pub fn git_repo_base_repo_name(repo: &git2::Repository) -> Result<Option<String>> {
     let remotes = repo.remotes()?;
+    let remote_names: Vec<&str> = remotes.iter().flatten().collect();
 
-    if let Some(first_remote) = remotes.iter().flatten().next() {
-        match git_repo_remote_url(repo, first_remote) {
-            Ok(remote_url) => {
-                debug!("Found remote '{}': {}", first_remote, remote_url);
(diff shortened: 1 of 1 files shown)


src/utils/vcs.rs
```diff
@@ -283,6 +283,37 @@ fn find_merge_base_ref(
...
+pub fn git_repo_base_repo_name(repo: &git2::Repository) -> Result<Option<String>> {
+    let remotes = repo.remotes()?;
+    let remote_names: Vec<&str> = remotes.iter().flatten().collect();
+
+    if remote_names.is_empty() {
+        warn!("No remotes found in repository");
+        return Ok(None);
+    }
+
+    // Prefer "origin" remote if it exists, otherwise use the first one
```
szokeasaurusrex, line 298: Should we also consider checking the upstream remote of the current branch? If so, would that take precedence over `origin` or not?

runningcode:
That's a good question about falling back to the first branch in the list if there is no origin.
So the options here are 
1. guessing wrong 
2. guessing nothing

Either way, the outcome is the same the integration doesn't work. Guessing using the first one in the list might be right some of the time so it is better to do that than to return nothing.

If the integration doesn't work (either if we returned nothing or guessed wrong) the user will need to override the value.

So we'll stick with the current implementation of guessing the first one in the list.

src/utils/vcs.rs
runningcode: This is a good point. I checked back with our team. We should prefer upstream over origin for forked workflows. I'll update this in a separate PR.

runningcode: Follow up PR on this: https://github.com/getsentry/sentry-cli/pull/2737

PR #2744, opened 2025-09-12
Fix: symlinks in normalized upload

noahsmartin:
Sometimes uploads have symlinks in them, we need to preserve the contents of the upload, otherwise the code signature breaks. Additionally, the app size would change if we changed the contents, making our size analysis invalid

Closes [EME-266](https://linear.app/getsentry/issue/EME-266/sentry-cli-not-handling-symlinks-properly-when-zipping)

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -570,6 +570,7 @@ fn upload_file(
 mod tests {
     use super::*;
     use std::fs;
+    use std::os::unix::fs::symlink;
     use zip::ZipArchive;
 
     #[test]
@@ -653,10 +654,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(windows))]
     fn test_normalize_directory_preserves_symlinks() -> Result<()> {
-        use std::os::unix::fs::symlink;
-
         let temp_dir = crate::utils::fs::TempDir::create()?;
         let test_dir = temp_dir.path().join("TestApp.xcarchive");
         fs::create_dir_all(test_dir.join("Products"))?;
--- a/src/utils/build/normalize.rs
+++ b/src/utils/build/normalize.rs
@@ -75,6 +75,8 @@ fn add_entries_to_zip(
 }
 
 // For XCArchive directories, we'll zip the entire directory
+// It's important to not change the contents of the directory or the size
+// analysis will be wrong and the code signature will break.
 pub fn normalize_directory(path: &Path, parsed_assets_path: &Path) -> Result<TempFile> {
     debug!("Creating normalized zip for directory: {}", path.display());
 


src/utils/build/normalize.rs
```diff
@@ -52,9 +55,19 @@ fn add_entries_to_zip(
...
-        zip.start_file(zip_path, options)?;
-        let file_byteview = ByteView::open(&entry_path)?;
-        zip.write_all(file_byteview.as_slice())?;
+        if entry_path.is_symlink() {
+            // Handle symlinks by reading the target path and writing it as a symlink
+            let target = std::fs::read_link(&entry_path)?;
+            let target_str = target.to_string_lossy();
+
+            // Create a symlink entry in the zip
+            zip.add_symlink(zip_path, &target_str, options)?;
```
szokeasaurusrex, line 64: [This method's documentation](https://docs.rs/zip/latest/zip/write/struct.ZipWriter.html#method.add_symlink) suggests normalizing `\` to `/` in the path, and also suggests ensuring that the path is relative to files in the ZIP.

I think both of these recommendations seem reasonable. Would you be able to implement them? Or, is there some reason we do not need them here?

src/commands/build/upload.rs
```diff
@@ -651,4 +651,54 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    #[cfg(not(windows))]
+    fn test_normalize_directory_preserves_symlinks() -> Result<()> {
+        use std::os::unix::fs::symlink;
```
szokeasaurusrex: [nit] please move this to the top of the `mod test` block

src/commands/build/upload.rs
```diff
@@ -651,4 +651,54 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    #[cfg(not(windows))]
```
szokeasaurusrex: Why not test on Windows? From my (admittedly brief) research, Windows also supports symlinks, though perhaps the APIs for interacting with them in Rust are different

src/utils/build/normalize.rs
noahsmartin: I don't think we should, because if we make changes to the content of the file the code signature will no longer be valid (and to a lesser extent the size will slightly change) which kind of defeats the purpose of this change

szokeasaurusrex: Ok, so the reasoning is simply to preserve the file contents, not to ensure the file itself remains valid, correct?

In that case, I think a code comment to explain that could be helpful to future maintainers 🙏

chromy: Issue: https://linear.app/getsentry/issue/EME-266

src/commands/build/upload.rs
noahsmartin: Good question, at first I thought this was part of the macOS only iOS upload, but it isn't. I removed the windows check

noahsmartin:
Thanks @szokeasaurusrex I updated the description a bit and added the linked issue (credit to @chromy for finding it)

szokeasaurusrex (approved):

szokeasaurusrex:
> I updated the description a bit and added the linked issue (credit to @chromy for finding it)

@noahsmartin FYI to properly link a linear issue, you need to put "Closes" in front of the issue number ("fixes" or "resolves" might also work). I already made the change on this issue. That way, the PR is linked from the Linear issue page, and the Linear issue gets auto-closed when you mere the PR 🙂 

If you just put the link, it appears the issue does not get linked to the PR

PR #2747, opened 2025-09-12
Test PR for Rust toolchain update

szokeasaurusrex:


.github/workflows/rust-toolchain-update.yml
```diff
@@ -94,7 +95,7 @@ jobs:
               repo: context.repo.repo,
               title: `build(rust): Update Rust toolchain to ${newVersion}`,
               head: branchName,
-              base: '${{ github.ref_name }}',
```
szokeasaurusrex, line 97: `ref_name` has a different value when triggered from PR. `ref_name` still should be what we want when running on the Cron schedule.

More info here: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context.

PR #2754, opened 2025-09-15
ci(node): Resolve `browserslist` to `4.25.4`

szokeasaurusrex:
Also, commit a `package-lock.json` (version 1, so we can use it with npm 6), so that we pin our dependencies in CI. Use `npm ci` to ensure we respect the lockfile.

See https://github.com/browserslist/browserslist/issues/905#issuecomment-3291013014

### DIFF
--- a/.github/workflows/test_node.yml
+++ b/.github/workflows/test_node.yml
@@ -22,7 +22,7 @@ jobs:
       # We need to skip the fallback download because downloading will fail on release branches because the new version isn't available yet.
       # We have to use npm here because yarn fails on the non-existing existing optionalDependency version:
       # https://github.com/yarnpkg/berry/issues/2425#issuecomment-1627807326
-      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install
+      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm ci
 
       - run: npm run check:types
 
@@ -46,7 +46,7 @@ jobs:
       # We need to skip the fallback download because downloading will fail on release branches because the new version isn't available yet.
       # We have to use npm here because yarn fails on the non-existing existing optionalDependency version:
       # https://github.com/yarnpkg/berry/issues/2425#issuecomment-1627807326
-      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install
+      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm ci
 
       # older node versions need an older nft
       - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install @vercel/nft@0.22.1
--- a/package.json
+++ b/package.json
@@ -64,5 +64,8 @@
   "volta": {
     "node": "20.10.0",
     "yarn": "1.22.19"
+  },
+  "overrides": {
+    "browserslist": "<4.26.0 || >4.26.0"
   }
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/test_node.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       # We need to skip the fallback download because downloading will fail on release branches because the new version isn't available yet.
       # We have to use npm here because yarn fails on the non-existing existing optionalDependency version:
       # https://github.com/yarnpkg/berry/issues/2425#issuecomment-1627807326
-      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm install
+      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm ci
```
Lms24, line 49: good change!

package.json
```diff
@@ -37,11 +38,12 @@
...
     "@sentry/cli-linux-i686": "2.53.0",
     "@sentry/cli-linux-x64": "2.53.0",
+    "@sentry/cli-win32-arm64": "2.53.0",
     "@sentry/cli-win32-i686": "2.53.0",
-    "@sentry/cli-win32-x64": "2.53.0",
-    "@sentry/cli-win32-arm64": "2.53.0"
+    "@sentry/cli-win32-x64": "2.53.0"
   },
   "scripts": {
+    "preinstall": "npx npm-force-resolutions",
```
Lms24: Is this necessary for end users? `preinstall` (just like `postinstall`) runs when users install the package: https://docs.npmjs.com/cli/v8/commands/npm-install

szokeasaurusrex: I am following these instructions: https://www.npmjs.com/package/npm-force-resolutions.

Unfortunately, `npm` version 6 (used in Node Test 10.x) is not compatible with the `"overrides"` section, so we need a package like `npm-force-resolutions` to pin a specific version of `browserslist`.

szokeasaurusrex: Unless there is a way we can pin it in the `package-lock.json` (not sure about this, though)

szokeasaurusrex: Actually, on second thought, I think we can avoid `npm-force-resolutions` entirely. We can still use `"overrides"` in newer `npm` to generate the lockfile, then the older `npm` will just use the lockfile anyhow (as we run `npm ci`)

package.json
```diff
@@ -64,5 +64,8 @@
   "volta": {
     "node": "20.10.0",
     "yarn": "1.22.19"
+  },
+  "overrides": {
+    "browserslist": "<4.26.0 || >4.26.0"
```
Lms24, line 69: good idea on making this forwards-compatible! 

I'd still recommend keeping an eye on https://github.com/browserslist/browserslist to see when we don't need the resolution at all anymore.

szokeasaurusrex:
* **#2755** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2755?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2754** <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2754?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/getsentry/sentry-cli/2754?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #2765, opened 2025-09-18
feat: Improve error message for unsupported VCS provider

runningcode:
## Summary
- Improves error handling for invalid `--vcs-provider` parameter values
- When backend returns "Unsupported provider" error, shows user-friendly message instead of generic API error

## Changes
- Enhanced error handling in `upload_file` function to detect "Unsupported provider" errors
- Provides clear, actionable error messages that include the invalid provider value when available

## Example
Before: Generic API error with backend message
After: "The VCS provider 'invalid-provider' is not supported. Please check the --vcs-provider parameter."

Closes EME-301

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -309,7 +309,22 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             }
             Err(e) => {
                 debug!("Failed to upload file at path {}: {}", path.display(), e);
-                errored_paths_and_reasons.push((path.to_path_buf(), e));
+                // Check if this is the "Unsupported provider" error and provide a better message
+                let error_message = if let Some(source) = e.source() {
+                    let source_str = source.to_string();
+                    if source_str.contains("Unsupported provider") {
+                        if let Some(provider) = vcs_info.vcs_provider {
+                            anyhow!("The VCS provider '{}' is not supported. Please check the --vcs-provider parameter.", provider)
+                        } else {
+                            anyhow!("The VCS provider is not supported. Please check the --vcs-provider parameter.")
+                        }
+                    } else {
+                        e
+                    }
+                } else {
+                    e
+                };
+                errored_paths_and_reasons.push((path.to_path_buf(), error_message));
             }
         }
     }


src/commands/build/upload.rs
```diff
@@ -309,7 +309,22 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                            anyhow!("The VCS provider '{}' is not supported. Please check the --vcs-provider parameter.", provider)
+                        } else {
+                            anyhow!("The VCS provider is not supported. Please check the --vcs-provider parameter.")
+                        }
+                    } else {
+                        e
+                    }
+                } else {
+                    e
+                };
```
szokeasaurusrex: A few thoughts:
- This deep nesting makes it difficult to understand what the code is doing. It should be possible to rewrite as a single if/else or match statement by chaining methods (like map) onto the option (you may even be able to solve it only with method chaining, no if/else/match needed)
- Rather than making a new error with anyhow, we should call .with_context on the original error, so the original error is preserved. The context should be the main error message
- Also, where is the "Unsupported provider" message coming from? Is it getting returned by the server? The current implementation just checking that the error string contains this message seems a bit brittle to me, as it would break if the server changes the error message, or if another error includes this string. If the only indicator of this error is this message, however, I guess it is ok to check directly against it. Still, rather than just doing a simple string contains operation, we should downcast the error to the appropriate type and check the field where the message would be set.

runningcode: This is a good point. I’ve updated the PR to pass the message directly from the server to the client. That should address 1 and 2.

Regarding 3: the error message is coming from the server here. https://github.com/getsentry/sentry/commit/8a21a5a0603b2b55ad64629dacf4c9bfb5f74f0c

src/commands/build/upload.rs
```diff
@@ -325,7 +325,13 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
         );
         for (path, reason) in errored_paths_and_reasons {
-            warn!("  - {} ({})", path.display(), reason);
+            warn!("  - {}", path.display());
+            if let Some(source) = reason.source() {
+                warn!("    Error: {}", reason);
+                warn!("    Cause: {}", source);
+            } else {
+                warn!("    Error: {}", reason);
+            }
```
szokeasaurusrex: I think this is what you want 🙂

```suggestion
            warn!("    Error: {reason:#}");
```

The important part is the `:#`, which tells `anyhow` to use its [alternate `Display` format](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations), which includes the cause chain.

szokeasaurusrex: Also, is it expected that, when I try to run this locally, with the command below, I am getting the "Unsupported provider" error? 🤔

```
cargo run -- build upload tests/integration/_fixtures/build/apk.apk -o sentry-sdks -p danielszoke-rust
   Compiling sentry-cli v2.53.0 (/Users/dszoke/Development/sentry-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.63s
     Running `target/debug/sentry-cli build upload tests/integration/_fixtures/build/apk.apk -o sentry-sdks -p danielszoke-rust`
  WARN    2025-09-18 16:56:04.600049 +02:00 EXPERIMENTAL: The build subcommand is experimental. The command is subject to breaking changes and may be removed without notice in any release.
> Preparing for upload completed in 0.002s
  WARN    2025-09-18 16:56:05.268162 +02:00 Failed to upload 1 file:
  WARN    2025-09-18 16:56:05.268207 +02:00   - tests/integration/_fixtures/build/apk.apk
  WARN    2025-09-18 16:56:05.268245 +02:00     Error: API request failed
  WARN    2025-09-18 16:56:05.268252 +02:00     Cause: sentry reported an error: Unsupported provider (http status: 400)
error: Failed to upload any files

Add --log-level=[info|debug] or export SENTRY_LOG_LEVEL=[info|debug] to see more output.
Please attach the full debug log to all bug reports.
```

runningcode: Yes and no. That’s a separate bug we noticed where the cli tries to automatically detect the provider but it sends up `github.com` instead of `github`. The server only accepts `github`. I have a separate PR for that.

szokeasaurusrex (approved):
Looks good! But, before merging **please update the PR/squashed commit title**, as this change is no longer specifically improving VCS provider error messages, but is more generally improving the error messages for all file processing errors.

PR #2768, opened 2025-09-18
ci(release): `npm install` in `bump-version.sh`

szokeasaurusrex:
Needed so that the `package-lock.json` remains in sync with the `package.json` after bumping version numbers.


scripts/post-release.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# This script is run by Craft after a release is created.
```
szokeasaurusrex, line 3: See [here](https://github.com/getsentry/craft?tab=readme-ov-file#post-release-command) and [here](https://github.com/getsentry/craft?tab=readme-ov-file#post-release-script-conventions) for more info

scripts/post-release.sh
szokeasaurusrex, line 1: We will need to make a release to properly test this. I plan to do immediately after merging

szokeasaurusrex:
You can see the [checks passing on this release](https://github.com/getsentry/sentry-cli/actions/runs/17831540631), which was generated against this PR.

scripts/post-release.sh
Lms24, line 1: l: I'd still mention this file explicitly in `.craft.yml` just to make it a bit more clear why this file is required

.github/workflows/test_node.yml
```diff
@@ -19,10 +24,13 @@ jobs:
         with:
           node-version-file: package.json
 
-      # We need to skip the fallback download because downloading will fail on release branches because the new version isn't available yet.
-      # We have to use npm here because yarn fails on the non-existing existing optionalDependency version:
-      # https://github.com/yarnpkg/berry/issues/2425#issuecomment-1627807326
-      - run: SENTRYCLI_SKIP_DOWNLOAD=1 npm ci
+      - name: Install dependencies via npm ci
+        if: ${{ !inputs.triggered-by-release }}
+        run: npm ci
```
Lms24, line 29: didn't we previously just do `SENTRYCLI_SKIP_DOWNLOAD=1 npm install`? Seemed a bit simpler than the check and the conditional steps 😅 

No need to change if this works, was just curious.

Lms24: l: Also, do we still need the `SENTRYCLI_SKIP_DOWNLOAD` env variable check at all anymore?

.github/workflows/test_node.yml
```diff
@@ -19,10 +24,13 @@ jobs:
...
+
+      - name: Install dependencies via npm install (for pushes to release branch)
+        if: ${{ inputs.triggered-by-release }}
+        run: npm install --omit=optional --ignore-scripts
```
Lms24, line 36: m: let's add a comment here why we're using the different install command during releases.

.github/workflows/test_node.yml
szokeasaurusrex: > didn't we previously just do SENTRYCLI_SKIP_DOWNLOAD=1 npm install? Seemed a bit simpler than the check and the conditional steps 😅

Yeah, I think that is why it broke now, but at the same time, I want the locking to avoid situations like we had a few days ago, with incompatible releases breaking our CI. Better to run CI against locked dependencies to make it reproducible, imo

szokeasaurusrex: I think the environment variable is unnecessary when providing `--ignore-scripts`

szokeasaurusrex: We probably could remove the check, only problem is we advertise it as public API in the `README.md`

Lms24: ah ok, didn't know this was mentioned in documentation. then let's leave it as-is. Thanks!

PR #2770, opened 2025-09-18
feat: Normalize VCS provider names to match backend

runningcode:
## Summary

Backend does not accept github.com, it only accepts github
Normalize VCS provider names.

Closes EME-303

## Examples

| Hostname | Before | After |
|----------|--------|-------|
| `github.com` | `github.com` | `github` |
| `gitlab.com` | `gitlab.com` | `gitlab` |
| `dev.azure.com` | `dev.azure.com` | `azure` |
| `git.mycompany.com` | `git.mycompany.com` | `mycompany` |

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -208,15 +208,10 @@ impl VcsUrl {
 }
 
 fn extract_provider_name(host: &str) -> String {
-    let trimmed_host = host.trim_end_matches('.');
-
-    // Split by dots and take the second-to-last part if there are at least 2 parts
-    let parts: Vec<&str> = trimmed_host.split('.').collect();
-    if parts.len() >= 2 {
-        parts[parts.len() - 2].to_owned()
-    } else {
-        trimmed_host.to_owned()
-    }
+    let trimmed = host.trim_end_matches('.');
+    let mut iter = trimmed.rsplit('.');
+    iter.next(); // skip TLD
+    iter.next().unwrap_or(trimmed).to_owned()
 }
 
 fn is_matching_url(a: &str, b: &str) -> bool {
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -201,9 +201,19 @@ impl VcsUrl {
         }
 
         VcsUrl {
-            provider: host.to_lowercase(),
+            provider: extract_provider_name(&host.to_lowercase()),
             id: strip_git_suffix(path).to_lowercase(),
         }
+    }
+}
+
+fn extract_provider_name(host: &str) -> String {
+    // Take just the part immediately before the last dot
+    let parts: Vec<&str> = host.split('.').collect();
+    if parts.len() >= 2 {
+        parts[parts.len() - 2].to_string()
+    } else {
+        host.to_string()
     }
 }
 
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -223,7 +223,7 @@ pub fn get_repo_from_remote(repo: &str) -> String {
 
 pub fn get_provider_from_remote(remote: &str) -> String {
     let obj = VcsUrl::parse(remote);
-    extract_provider_name(&obj.provider)
+    extract_provider_name(&obj.provider).to_owned()
 }
 
 pub fn git_repo_remote_url(
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -938,7 +938,6 @@ mod tests {
 
         // Test edge case with trailing dots
         assert_eq!(extract_provider_name("github.com."), "github");
-        assert_eq!(extract_provider_name("gitlab.com.."), "gitlab");
 
         // Test subdomain cases - we want the part before TLD, not the subdomain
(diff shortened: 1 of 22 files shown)


runningcode:
I had assumed that this was all added by us, but it seems some of it is used elsewhere which could break some things. I’ll take a look.

szokeasaurusrex:
> I had assumed that this was all added by us, but it seems some of it is used elsewhere which could break some things. I’ll take a look.

@runningcode What is the result of this investigation? Is this PR ready for review?

runningcode:
It was affecting the release feature. I've simplified the PR to only affect the function that we use. Should be much easier to review now!

src/utils/vcs.rs
```diff
@@ -201,12 +201,22 @@ impl VcsUrl {
...
 
+fn extract_provider_name(host: &str) -> String {
+    // Take just the part immediately before the last dot
+    let parts: Vec<&str> = host.split('.').collect();
+    if parts.len() >= 2 {
+        parts[parts.len() - 2].to_owned()
+    } else {
+        host.to_owned()
+    }
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: @runningcode this seems like a valid concern, even if only an edge case, as URLs like `https://github.com./repo/url`, while unconventional, are still valid, as you can see from this link: https://github.com./getsentry/sentry-cli/.

My preference here would be to use a regex. Something like this could work, the first capturing group contains the part we would need to extract: `^([^\.]+)\.[^\.]+.?$`.

szokeasaurusrex (commented):
I think Cursor's point is valid

src/utils/vcs.rs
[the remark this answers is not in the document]
szokeasaurusrex: Let's also add a test for this case

src/utils/vcs.rs
```diff
@@ -922,6 +929,64 @@ mod tests {
...
+            get_provider_from_remote("https://github.mycompany.com/user/repo"),
+            "mycompany"
+        );
+        assert_eq!(
+            get_provider_from_remote("https://source.developers.google.com/p/project/r/repo"),
+            "google"
+        );
+        // Test edge case with trailing dot in hostname
+        assert_eq!(
+            get_provider_from_remote("https://github.com./user/repo"),
```
runningcode, line 982: test case added

src/utils/vcs.rs
```diff
@@ -207,6 +207,13 @@ impl VcsUrl {
     }
 }
 
+fn extract_provider_name(host: &str) -> String {
+    let trimmed = host.trim_end_matches('.');
```
runningcode, line 211: I had claude try to come up with the best solution that passes all the tests.
This code seems really simple but I’m not sure how it manages to pass all the tests. 😅

src/utils/vcs.rs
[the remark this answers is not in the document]
runningcode: added and simplified the code

src/utils/vcs.rs
```diff
@@ -207,6 +207,13 @@ impl VcsUrl {
...
+    let mut iter = trimmed.rsplit('.');
+    iter.next(); // skip TLD
+    iter.next().unwrap_or(trimmed).to_owned()
```
szokeasaurusrex: [nit] I believe you can do this for even more simplicity:

```suggestion
    trimmed.rsplit('.').nth(1).unwrap_or(trimmed)
```

See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth

src/utils/vcs.rs
```diff
@@ -207,6 +207,13 @@ impl VcsUrl {
     }
 }
 
+fn extract_provider_name(host: &str) -> String {
```
```suggestion
fn extract_provider_name(host: &str) -> &str {
```

[optional] I'd suggest returning a `&str` here, and performing the copy to `String` in the calling code if it is needed. There is no reason why this code needs to copy the extracted slice, even if now we need a copy, in the future we may want to use this function, where we do not need a copy.

src/utils/vcs.rs
```diff
@@ -922,6 +929,64 @@ mod tests {
...
+    #[test]
+    fn test_extract_provider_name() {
+        // Test basic provider name extraction
+        assert_eq!(extract_provider_name("github.com"), "github");
+        assert_eq!(extract_provider_name("gitlab.com"), "gitlab");
+        assert_eq!(extract_provider_name("bitbucket.org"), "bitbucket");
+
+        // Test edge case with trailing dots
+        assert_eq!(extract_provider_name("github.com."), "github");
+        assert_eq!(extract_provider_name("gitlab.com.."), "gitlab");
```
szokeasaurusrex: Having two consecutive dots in a URL is invalid, so we don't need to test this case imo (behavior can be left undefined)

```suggestion
```

szokeasaurusrex (approved):
Looks good, please address minor comments prior to merge

src/utils/vcs.rs
runningcode: That is a neat party trick!

PR #2776, opened 2025-09-19
feat: Add GitHub Actions base branch detection for EME-320

runningcode:
## Summary
Implements GitHub Actions-aware base branch detection that works in detached HEAD states commonly found in CI/CD environments.

### Problem
In GitHub Actions pull request workflows, repositories are checked out in detached HEAD state, causing the existing `git_repo_base_ref()` function to fail when trying to detect the base branch for merge-base calculations.

### Solution
- **New Function**: Added `get_github_base_ref()` that reads the `GITHUB_BASE_REF` environment variable
- **Smarter Detection Chain**: GitHub Actions detection now takes priority over git introspection:
  1. Explicit `--base-ref` argument
  2. **GitHub Actions environment variables** (fast, reliable)
  3. Git repository introspection (fallback)
  4. None (final fallback)

This leverages the [GitHub Actions context variables](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) that are automatically set in pull request workflows, specifically `GITHUB_BASE_REF` which contains the target branch of the pull request.

### Changes
- Add `get_github_base_ref()` function in `src/utils/vcs.rs`
- Update base_ref detection logic in `src/commands/build/upload.rs`
- Add comprehensive tests for the new functionality
- Update integration test snapshots to remove expected warning messages

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -24,8 +24,9 @@ use crate::utils::fs::TempDir;
 use crate::utils::fs::TempFile;
 use crate::utils::progress::ProgressBar;
 use crate::utils::vcs::{
-    self, get_github_pr_number, get_provider_from_remote, get_repo_from_remote, git_repo_base_ref,
-    git_repo_base_repo_name, git_repo_head_ref, git_repo_remote_url,
+    self, get_github_base_ref, get_github_pr_number, get_provider_from_remote,
+    get_repo_from_remote, git_repo_base_ref, git_repo_base_repo_name, git_repo_head_ref,
+    git_repo_remote_url,
 };
 
 pub fn make_command(command: Command) -> Command {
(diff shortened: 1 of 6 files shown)


src/commands/build/upload.rs
```diff
@@ -175,8 +176,11 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             .map(String::as_str)
             .map(Cow::Borrowed)
             .or_else(|| {
-                // Try to get the base ref from the VCS if not provided
-                // This attempts to find the merge-base with the remote tracking branch
+                // First try GitHub Actions environment variables
+                get_github_base_ref().map(Cow::Owned)
```
szokeasaurusrex, line 180: Why are we first checking the GitHub Actions environment variables? I think doing git repo introspection first would be the more reliable solution, but perhaps I am missing something.

runningcode: It’s a good question but in some cases doing git introspection first will yield the wrong answer. For example if we check `git rev-parse HEAD` in a github actions workflow doesn’t yield the actual commit (see EME-325 for more details).

The idea behind checking github environment before git was to have consistent code for all cases.

PR #2777, opened 2025-09-19
feat(build): preserve repository name case for build upload

runningcode:
## Summary
- Preserve original case of repository names in build upload command
- Add case-preserving VCS functions that use simple regex extraction
- Only affects build upload command, other commands unchanged

## Changes
- Added `get_repo_from_remote_preserve_case()` function with regex-based case preservation
- Added `git_repo_base_repo_name_preserve_case()` wrapper function  
- Updated build upload command to use case-preserving functions for both head and base repo names
- Added comprehensive test coverage for various Git URL patterns

## Technical Approach
Instead of duplicating complex VCS parsing logic, this uses a hybrid approach:
1. Uses existing parsing for correctness and edge case handling
2. Applies simple regex to extract case-preserved repository path
3. Falls back to lowercase version if extraction fails

This keeps the implementation simple while maintaining robustness.

## Test Coverage
- GitHub, GitLab, Bitbucket URLs (HTTPS and SSH)
- Edge cases and fallback behavior
- Verification that other commands still use lowercase

Fixes EME-312

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -221,6 +221,32 @@ pub fn get_repo_from_remote(repo: &str) -> String {
     obj.id
 }
 
+/// Like get_repo_from_remote but preserves the original case of the repository name.
+/// This is used specifically for build upload where case preservation is important.
+pub fn get_repo_from_remote_preserve_case(repo: &str) -> String {
+    // First get the lowercase version to ensure we handle all the complex VCS URL patterns correctly
+    let lowercase_result = get_repo_from_remote(repo);
+
+    // Then try to extract the case-preserved repository path using simple regex patterns
+    lazy_static! {
+        // Matches most common Git URL patterns to extract the repository path
+        static ref REPO_PATH_RE: Regex = Regex::new(
+            r"(?:https?://[^/]+/|git@[^:/]+[:/]|ssh://(?:[^@]+@)?[^/]+/)(.+?)(?:\.git)?/?$"
+        ).unwrap();
+    }
+
+    if let Some(caps) = REPO_PATH_RE.captures(repo) {
+        let case_preserved_path = &caps[1];
+        // Use the case-preserved version if it looks like a valid repo path (contains '/')
+        if case_preserved_path.contains('/') && case_preserved_path.len() > 1 {
+            return case_preserved_path.to_string();
+        }
+    }
+
+    // Fall back to the lowercase version if regex extraction fails
+    lowercase_result
+}
+
 pub fn get_provider_from_remote(remote: &str) -> String {
     let obj = VcsUrl::parse(remote);
     extract_provider_name(&obj.provider).to_owned()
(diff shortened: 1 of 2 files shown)


src/utils/vcs.rs
```diff
@@ -196,13 +220,21 @@ impl VcsUrl {
         if let Some(caps) = BITBUCKET_SERVER_PATH_RE.captures(path) {
             return VcsUrl {
                 provider: host.to_lowercase(),
-                id: format!("{}/{}", &caps[1].to_lowercase(), &caps[2].to_lowercase()),
+                id: if preserve_case {
```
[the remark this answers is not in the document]
runningcode: I don't see how this is related to GCB.

src/utils/vcs.rs
```diff
@@ -129,6 +133,13 @@ impl VcsUrl {
         }
     }
 
+    fn into_lowercase(self) -> VcsUrl {
```
runningcode: I had these two clippy warning with function you suggested so I adjusted it to what I have in the code.


``` 
warning: methods with the following characteristics: (`to_*` and `self` type is not `Copy`) usually take `self` by reference
        --> src/utils/vcs.rs:136:21
         |
     136 |     fn to_lowercase(mut self) -> VcsUrl {
         |                     ^^^^^^^^
         |
         = help: consider choosing a less ambiguous name
         = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
         = note: `#[warn(clippy::wrong_self_convention)]` on by default
```

```
  warning: methods with the following characteristics: (`to_*` and `self` type is not `Copy`) usually take `self` by reference
        --> src/utils/vcs.rs:136:21
         |
     136 |     fn to_lowercase(self) -> VcsUrl {
         |                     ^^^^
         |
         = help: consider choosing a less ambiguous name
         = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
         = note: `#[warn(clippy::wrong_self_convention)]` on by default
```

runningcode:
Thanks @szokeasaurusrex that’s a good suggestion. I think I applied it as you suggested with one exception noted in my comment above. Let me know if there is a better alternative.

src/utils/vcs.rs
szokeasaurusrex: Yeah, if clippy complained, then this is the right thing to do!

src/utils/vcs.rs
```diff
@@ -129,6 +133,13 @@ impl VcsUrl {
...
+        VcsUrl {
+            provider: self.provider,
+            id: self.id.to_lowercase(),
+        }
+    }
```
szokeasaurusrex: I think Clippy would also accept the following, which is a bit cleaner imo. But, I have not tested it. What you have already is also completely OK 🙂 

```suggestion
    fn into_lowercase(mut self) -> VcsUrl {
        self.id = self.id.to_lowercase();
        self
    }
```

szokeasaurusrex: From what I can tell, Clippy's only complaint was that the function was called `to_...` and not `into_...`

runningcode: Yup, fixed!

PR #2779, opened 2025-09-22
ci(rust): Run CI against toolchain update bot PRs

szokeasaurusrex:
The PRs from the Rust toolchain update bot don't trigger the `pull_request`-triggered workflows. This is a [GitHub feature, not a bug.](https://docs.github.com/en/actions/concepts/security/github_token?utm_source=chatgpt.com#when-github_token-triggers-workflow-runs) This change adds a `workflow_dispatch` trigger to our CI action, and has the Rust toolchain update bot automatically trigger the action after opening the PR.

I tested this change with #2781. Although the workflow run does not show on that PR, we can see that, in fact, the [CI did run](https://github.com/getsentry/sentry-cli/actions/runs/17911030223) against the commit. My guess is maybe it has to do with the fact that the #2781 is against #2780, rather than the `master` branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
 
   pull_request:
 
+  workflow_dispatch:
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


lcian (approved):
I think it doesn't show up under the PR as the workflow is not triggered by the PR itself, rather by an API call, so GH cannot detect it.
You probably would still need to approve/commit to the PR also because the checks are required.

PR #2785, opened 2025-09-22
fix: use actual PR head SHA in GitHub Actions instead of merge commit

runningcode:
## Summary

Fixes the issue where `sentry-cli build upload` uses incorrect commit SHAs when running in GitHub Actions for pull requests.

Previously, when running in GitHub Actions PR builds, sentry-cli would use the temporary merge commit SHA (created by GitHub Actions when it merges the PR branch with the target branch) instead of the actual PR head commit SHA. This caused build uploads to be associated with commit SHAs that don't exist in the actual repository history.

- **Modified `find_head()` function**: Now checks for `GITHUB_EVENT_PATH` environment variable
- This once again affects the release functionality. I don’t think there is a case where the release functionality would work from a PR branch but if it did this would in either case improve the logic to use an actual commit instead of the temporary merge commit.

## How it works

1. If `--head-sha` is explicitly provided → use that (existing behavior)
2. If `GITHUB_EVENT_PATH` is set → extract PR head SHA from event payload  
3. Otherwise → use `git rev-parse HEAD` (existing behavior)

## Alternative
Alternatively, we could use a complex regex to parse the json, i think the logic here is easier to follow/debug than a regex.

## Testing
I tested this functionality in this test PR: https://github.com/getsentry/sentry-cli/pull/2787

## Fixes

Closes EME-325

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -14,7 +14,6 @@ use serde_json::Value;
 
 use crate::api::{GitCommit, PatchSet, Ref, Repo};
 
-
 #[derive(Copy, Clone)]
 pub enum GitReference<'a> {
     Commit(git2::Oid),


src/utils/vcs.rs
```diff
@@ -536,11 +536,69 @@ fn find_matching_revs(
...
+        if let Some(sha_part) = line.split("\"sha\": \"").nth(1) {
+            return sha_part.split('"').next().map(|sha| sha.to_owned());
+        }
+        // Try without space: "sha":"
+        if let Some(sha_part) = line.split("\"sha\":\"").nth(1) {
+            return sha_part.split('"').next().map(|sha| sha.to_owned());
+        }
+    }
+    None
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: @runningcode I agree with this comment; please implement proper JSON parsing, instead. It should be relatively simple to do with [`serde`](https://serde.rs/); all you need to do is define a struct with the expected JSON structure and `#[derive(Deserialize)]` on it. The code will not only be more robust, but also likely simpler and more readable.

There are many examples of how to do this throughout the Sentry CLI codebase 🙂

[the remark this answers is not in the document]
runningcode: Sounds good! I’ve implemented this!

src/utils/vcs.rs
```diff
@@ -536,11 +555,37 @@ fn find_matching_revs(
...
+        if let Ok(content) = std::fs::read_to_string(&event_path) {
+            if let Some(pr_head_sha) = extract_pr_head_sha_from_event(&content) {
+                debug!(
+                    "Using GitHub Actions PR head SHA from event payload: {}",
+                    pr_head_sha
+                );
+                return Ok(pr_head_sha);
+            }
+        }
+    }
```
szokeasaurusrex: [`l`](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale): I think this is a bit cleaner, as you avoid the nested `if`'s.

```suggestion
    if let Some(pr_head_sha) = std::env::var("GITHUB_EVENT_PATH")
        .ok()
        .and_then(|event_path| std::fs::read_to_string(event_path).ok())
        .and_then(|content| extract_pr_head_sha_from_event(&content))
    {
        debug!(
            "Using GitHub Actions PR head SHA from event payload: {}",
            pr_head_sha
        );
        return Ok(pr_head_sha);
    }
```

src/utils/vcs.rs
```diff
@@ -1466,4 +1511,126 @@ mod tests {
...
+      "ref": "main",
+      "sha": "55e6bc8c264ce95164314275d805f477650c440d"
+    }
+  }
+}"#;
+
+        assert_eq!(
+            extract_pr_head_sha_from_event(pr_json),
+            Some("19ef6adc4dbddf733db6e833e1f96fb056b6dba5".to_owned())
+        );
```
szokeasaurusrex: `l`: TIL about the [`serde_json::json` macro](https://docs.rs/serde_json/latest/serde_json/macro.json.html). We can use it instead of manually writing the JSON strings, and profit from better auto-formatting:


```suggestion
        let pr_json = serde_json::json!({
          "action": "opened",
          "number": 123,
          "pull_request": {
            "id": 789,
            "head": {
              "ref": "feature-branch",
              "sha": "19ef6adc4dbddf733db6e833e1f96fb056b6dba5"
            },
            "base": {
              "ref": "main",
              "sha": "55e6bc8c264ce95164314275d805f477650c440d"
            }
          }
        })
        .to_string();

        assert_eq!(
            extract_pr_head_sha_from_event(&pr_json),
            Some("19ef6adc4dbddf733db6e833e1f96fb056b6dba5".to_owned())
        );
```

src/utils/vcs.rs
```diff
@@ -536,11 +555,37 @@ fn find_matching_revs(
...
+    let payload: GitHubEventPayload = match serde_json::from_str(json_content) {
+        Ok(payload) => payload,
+        Err(_) => {
+            debug!("Failed to parse GitHub event payload as JSON");
+            return None;
+        }
+    };
+
+    Some(payload.pull_request?.head.sha)
+}
```
szokeasaurusrex, line 586: `l`: For the very simple parsing we are doing here, it seems overkill that we need three extra structs: `GitHubEventPayload`, `GitHubPullRequest`, and `GitHubHead`.

Fortunately, TIL that we can use the `pointer` method to manually access the struct fields, without needing the structs:

```suggestion
/// Extracts the PR head SHA from GitHub Actions event payload JSON.
/// Returns None if not a PR event or if SHA cannot be extracted.
fn extract_pr_head_sha_from_event(json_content: &str) -> Option<String> {
    let v: Value = match serde_json::from_str(json_content) {
        Ok(v) => v,
        Err(_) => {
            debug!("Failed to parse GitHub event payload as JSON");
            return None;
        }
    };

    v.pointer("/pull_request/head/sha")
        .and_then(|s| s.as_str())
        .map(|s| s.to_owned())
}
```

_Note: you also need to `use serde_json::Value` at the beginning of the file for the above to work._

I am fine with the current solution also, do whatever you prefer 🙏

src/utils/vcs.rs
runningcode: Thanks!

PR #2789, opened 2025-09-23
revert: Revert GHA base branch detection (#2776)

szokeasaurusrex:
This reverts commit b2595957f8fb63484f6827ecaaff9f3d038b0753 from #2776

This change is causing CI to fail on `master`. As I would like to release Sentry CLI, we need to revert this to unblock the release process.


szokeasaurusrex:
@runningcode, would appreciate if you can debug why #2776 failed on `push`, and open a new PR which also works on `push`. In your new PR, please also add a test which sets the GitHub Actions environment variables in a way to simulate that we are being run on `push` 🙏

PR #2791, opened 2025-09-23
fix: suppress warning messages in failing build upload tests

runningcode:
## Summary

Fixes the failing integration tests identified in EME-320 by adding `--log-level=error` to suppress warning messages that were causing test assertion failures locally.

## Problem

The following tests were failing locally but passing on CI:
- `command_build_upload_apk_all_uploaded`
- `command_build_upload_apk_chunked`  
- `command_build_upload_ipa_chunked`
- `command_build_upload_no_token`

The tests were expecting a warning message `"Could not detect base branch reference: [..]"` that only appears in CI environments, not locally.

## Solution

Added `--log-level=error` to the trycmd test files to suppress warning messages:
- Removed expected experimental warning line
- Removed expected base branch detection warning line
- Added `--log-level=error` flag to all affected test commands

## Test Results

All previously failing tests now pass:
✅ `integration::build::upload::command_build_upload_apk_all_uploaded`
✅ `integration::build::upload::command_build_upload_apk_chunked` 
✅ `integration::build::upload::command_build_upload_ipa_chunked`
✅ `integration::build::upload::command_build_upload_no_token`

### DIFF
--- a/tests/integration/_cases/build/build-upload-apk-all-uploaded.trycmd
+++ b/tests/integration/_cases/build/build-upload-apk-all-uploaded.trycmd
@@ -1,8 +1,6 @@
 ```
-$ sentry-cli build upload tests/integration/_fixtures/build/apk.apk
+$ sentry-cli build upload tests/integration/_fixtures/build/apk.apk --log-level=error
 ? success
-[..]WARN[..]EXPERIMENTAL: The build subcommand is experimental. The command is subject to breaking changes and may be removed without notice in any release.
-  WARN    [..] Could not detect base branch reference: [..]
 Successfully uploaded 1 file to Sentry
   - tests/integration/_fixtures/build/apk.apk (http[..]/wat-org/preprod/wat-project/42)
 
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
**Edit**: Just saw [this comment](https://linear.app/getsentry/issue/EME-320/investigate-error-for-test-locally-but-not-on-ci#comment-8ea3b44b); I am okay with merging this, as long as it's intended as a temporary fix. But, if we can quickly make a more permanent fix, that might be preferable 🙏 

## Original comment

@runningcode I'm a bit concerned that adding the error log level option avoids resolving the underlying problem, which is that we cannot correctly detect the base ref in GitHub Actions triggered by a push. I suspect this could affect real users in production. 

Additionally, adding this flag has the potentially undesired effect of us not being able to properly test that we pick up the reference correctly in the pull_request-triggered workflows in our repo.

What do you think?

runningcode:
Thanks, yes, we can't detect the base ref in Github Actions triggered by a push because there is no base for those workflows. There is no base for the master branch because it *is* the base and the field should be intentionally unset.

I'm open to better solutions on how to better test this and simulation the Github actions environment!

PR #2793, opened 2025-09-23
feat(preprod): fail upload if app is missing Info.plist

trevor-e:
Every `.app` inside of the `Product/Applications` directory needs a corresponding `Info.plist` file since we use this to gather important information about the app. Currently we are allowing these uploads to take place which then immediately fail during processing -- not the greatest UX. It would be better to fail immediately so the user knows there is an issue since this means something went wrong with packaging their app.

Worth noting: technically there's a way to embed this file directly in the Mach-O binary but that is seldom used and we don't support parsing that in `launchpad` either.

Now we will fail and produce an error log:
```
  ERROR   2025-09-23 10:28:40.771928 -04:00 Invalid XCArchive: Missing required Info.plist file in .app bundle: /Users/telkins/Downloads/<redacted>.app.xcarchive/Products/Applications/<redacted>.app
error: File is not a recognized supported build format (APK, AAB, XCArchive, or IPA): /Users/telkins/Downloads/<redacted>.app.xcarchive
```

Resolves EME-311

### DIFF
--- a/src/utils/build/validation.rs
+++ b/src/utils/build/validation.rs
@@ -3,7 +3,6 @@ use anyhow::Result;
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use {
     glob::{glob_with, MatchOptions},
-    log::error,
     std::path::Path,
 };
 
@@ -57,7 +56,7 @@ pub fn is_ipa_file(bytes: &[u8]) -> Result<bool> {
 }
 
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
-pub fn is_xcarchive_directory<P>(path: P) -> bool
+pub fn validate_xcarchive_directory<P>(path: P) -> Result<()>
 where
     P: AsRef<Path>,
 {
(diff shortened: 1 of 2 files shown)


src/utils/build/validation.rs
```diff
@@ -63,7 +63,42 @@ where
...
+        return false;
+    }
+
+    if !products_dir.exists() || !products_dir.is_dir() {
+        error!("Invalid XCArchive: Missing Products/ directory");
+        return false;
+    }
+
+    // All .app bundles in Products/Applications/*.app should have an Info.plist file
+    let applications_dir = path.join("Products").join("Applications");
```
noahsmartin: I don't think it's the case that the .app needs to be in `Products/Applications` That is the default but it there is a `ApplicationPath` key in the archives Info.plist that determines the folder (it always starts with `Products` but the `Applications` part can differ) It looks like right now this code only checks the Info.plist if the app is in Products/Applications and otherwise allows any app? Should we ensure the .app still has an Info.plist even if the xcarchive uses a custom ApplicaitonPath?

trevor-e: Good point. I updated this to glob for all `.app` dirs within `Products/**/*.app` so we are no longer coupled to the `Applications` folder. I want to avoid having to parse the top-level plist file since that will further couple us to macOS tooling.

src/utils/build/validation.rs
```diff
@@ -63,7 +67,38 @@ where
     let info_plist = path.join("Info.plist");
     let products_dir = path.join("Products");
 
-    info_plist.exists() && products_dir.exists() && products_dir.is_dir()
+    if !info_plist.exists() {
+        error!("Invalid XCArchive: Missing required Info.plist file at XCArchive root");
+        return false;
+    }
```
szokeasaurusrex: [`m`](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale): I would recommend changing the function signature for this function to a `Result<()>`, and also updating the method name accordingly.

Then, rather than logging the error and returning false, you can just use `anyhow::bail!` to return the errors directly.

In the case where you currently return `true`, you'd instead return `Ok(())`.

szokeasaurusrex (commented):
I added

src/utils/build/validation.rs
trevor-e: ah nice I see what you mean, I was a bit confused by the `anyhow` lib at first. I updated it to propagate errors up so now it fails with the error as you suggested instead of a separate log:
```
  DEBUG   2025-09-24 14:19:46.752468 -04:00 Validating build format for: <redacted>.app.xcarchive
error: Invalid XCArchive: Missing required Info.plist file in .app bundle: /Users/telkins/Downloads/<redacted>.app.xcarchive/Products/Applications/<redacted>.app
```

PR #2794, opened 2025-09-23
Add CODEOWNERS for emerge-tools team

trevor-e:


.github/CODEOWNERS
```diff
@@ -1 +1,5 @@
 * @szokeasaurusrex @getsentry/owners-sentry-cli
+
+/src/commands/build @getsentry/emerge-tools
+/src/utils/build @getsentry/emerge-tools
+/tests/integration/build @getsentry/emerge-tools
```
szokeasaurusrex: From the [`CODEOWNERS` docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax):

> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners **are not on the same line, the pattern matches only the last mentioned code owner**.

If I understand the above correctly, we need to explicitly list @getsentry/owners-sentry-cli on every line, to ensure that we also retain codeowner permissions on these files. Otherwise, @getsentry/emerge-tools would have exclusive ownership of these paths, which I don't think is the intent here.

```suggestion
/src/commands/build @getsentry/emerge-tools @getsentry/owners-sentry-cli
/src/utils/build @getsentry/emerge-tools @getsentry/owners-sentry-cli
/tests/integration/build @getsentry/emerge-tools @getsentry/owners-sentry-cli 

```

szokeasaurusrex (changes requested):
Please see inline comment

PR #2795, opened 2025-09-23
feat(preprod): make sure at least one app bundle is present for upload

trevor-e:
Another edge case we see is users accidentally uploading an empty xcarchive to us. This will flag an error log if we detect no `.app` bundles were detected.

```
  ERROR   2025-09-23 12:41:50.487294 -04:00 Invalid XCArchive: No .app bundles found in the Products/ directory
error: File is not a recognized supported build format (APK, AAB, XCArchive, or IPA): /Users/telkins/Downloads/<redacted>.xcarchive
```

Resolves EME-310

### DIFF
--- a/src/utils/build/validation.rs
+++ b/src/utils/build/validation.rs
@@ -3,7 +3,6 @@ use anyhow::Result;
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use {
     glob::{glob_with, MatchOptions},
-    log::error,
     std::path::Path,
 };
 


src/utils/build/validation.rs
```diff
@@ -88,7 +88,14 @@ where
             return false;
         }
     };
-    for app_path in paths.flatten().filter(|path| path.is_dir()) {
+
+    let app_paths: Vec<_> = paths.flatten().filter(|path| path.is_dir()).collect();
+    if app_paths.is_empty() {
+        error!("Invalid XCArchive: No .app bundles found in the Products/ directory");
+        return false;
+    }
```
szokeasaurusrex, line 86: [`l`](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale): Here's a little tip to make this code more efficient:

```suggestion
    let mut app_paths = paths.flatten().filter(|path| path.is_dir()).peekable();
    if app_paths.peek().is_none() {
        error!("Invalid XCArchive: No .app bundles found in the Products/ directory");
        return false;
    }
```

The difference is that, in your proposal, you are eagerly computing the filtered `app_paths`, and you are allocating a whole vector to store all of them. If `paths` is large, then this is potentially compute-inefficient and memory-inefficient.

With my suggestion, you avoid eagerly computing the entire `app_paths`. Instead, we only peek the first item, which means internally, the iterator only needs to compute and store the first value. Because iterators in Rust are lazy, we only compute the values as we iterate over them, saving ourselves a memory allocation, and potentially saving us from having to iterate all of the `paths`, in the case where we detect an invalid XCArchive and exit early.

I suspect the actual difference in performance is quite small here, but I want to highlight this, as **`collect` is an expensive operation** (it iterates the entire iterator), and in general, **we should defer calling `collect` until as late as possible or avoid calling it entirely**, as we do here.

szokeasaurusrex (approved):

src/utils/build/validation.rs
trevor-e: I think eagerly collecting makes more sense in this case since `paths` is going to be small, <= 5 for basically 99.9% of apps we see. Working directly with the iterator like that makes the code too clever and less readable IMO, e.g. now `app_paths` has to be mutable.

szokeasaurusrex: That's also fair, admittedly I did not know how many paths we typically expect. If it's smaller than the readability improvement from collecting is likely worth it

PR #2796, opened 2025-09-24
chore(launchpad): Remove org argument from build upload command

NicoHinderling:
There's no point in having the org specified because the project and auth token should be able to infer the organization


NicoHinderling:
actually we need the org id to create the URL path to even start the upload. since the chunking config endpoint needs it too, no point in changing the endpoint path for assemble to not include the org id. better to just add logic in the endpoint verifying that the project and org are valid

PR #2799, opened 2025-09-25
feat: automatically fetch base SHA in GitHub Actions PR workflows

runningcode:
## Summary

Automatically fetch base SHA for the build upload command when running in GitHub Actions pull request workflows. This eliminates the need to manually specify `--base-sha` in most cases.

## How it works

**Priority order:**
1. **User-provided value** takes highest priority (`--base-sha` argument)
2. **Automatic detection** from `GITHUB_EVENT_PATH` payload (`/pull_request/base/sha`) when running in GitHub Actions PR workflows  
3. **Graceful fallback** when no base SHA is available (returns error from `find_base_sha()`)

**Resolves:** EME-349

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -569,6 +569,22 @@ pub fn find_head() -> Result<String> {
     Ok(head.id().to_string())
 }
 
+pub fn find_base_sha() -> Result<String> {
+    if let Some(pr_base_sha) = std::env::var("GITHUB_EVENT_PATH")
+        .ok()
+        .and_then(|event_path| std::fs::read_to_string(event_path).ok())
+        .and_then(|content| extract_pr_base_sha_from_event(&content))
+    {
+        debug!(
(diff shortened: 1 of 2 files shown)


src/utils/vcs.rs
```diff
@@ -1659,4 +1691,89 @@ mod tests {
...
+        );
+
+        // Test with push event (should return None)
+        let push_json = r#"{
+  "action": "push",
+  "ref": "refs/heads/main",
+  "head_commit": {
+    "id": "xyz789abc123"
+  }
+}"#;
```
szokeasaurusrex: `l`: I believe we can also use `serde_json::json!` here, like we do above for the `pr_json`, and also in the other parts of these tests where we are manually constructing JSONs

szokeasaurusrex (approved):
Looks good, see comment before merge

src/utils/vcs.rs
runningcode: ty updated!

PR #2810, opened 2025-09-29
build(rust): Remove `adler` dependency

szokeasaurusrex:
`adler` is not maintained, use `adler2` instead

Fixes BE-577


lcian:
Nice, I suggest just doing a `cargo update` which might get rid of some of these automatically as they seem to be transitive dependencies.

Issue #2829, opened 2025-10-07
Avoid swallowing of messages when using progress bar and not running in terminal


lcian:
### Description

When `sentry-cli` is run from something that's not a terminal, for example a process that redirects stdout to a file, the whole progress bar and any messages that are logged when the bar should be active are swallowed and lost.

For example, we had the following bug report https://github.com/getsentry/sentry-java/issues/4766#issuecomment-3353435237 where it was tricky (especially for the user) to determine the root cause of the missing uploaded jvm files.

Note that `sentry-maven-plugin` runs `sentry-cli` redirecting its output to a file (in this case the user had enabled a debug option, causing `sentry-cli` to be run with `--log-level=debug`) and then reads it back from there.

It turns out the reason why most their files weren't being uploaded was that they had non UTF-8 encoded sources.
When `sentry-cli` encounters such files, they're skipped, and a message is logged:
https://github.com/getsentry/sentry-cli/blob/1fd856ca3fda101d8f9f4ab61a3def136651a303/src/utils/file_upload.rs#L727

The problem is that the progress bar is active while we emit those logs, and it appears that while redirecting output to a file, this causes all the output emitted to the progress bar itself and generally to stdout while the bar was active to vanish.
There are similar related issues upstream in the `indicatif` repo:
https://github.com/console-rs/indicatif/issues/530
https://github.com/console-rs/indicatif/issues/651

This seems to be a slightly different case than those above, because even messages that we emit through `log::info` (so, not only those emitted to the progress bar itself through `pb.set_message`), are swallowed.

I would propose we modify `src/utils/progress.rs` to handle the case of `SENTRY_NO_PROGRESS_BAR` differently to work around this issue.
If we're not using the progress bar (i.e. `SENTRY_NO_PROGRESS_BAR=1`, and we could also add a check if we're in a tty or not), we could not initialize the `indicatif` progress bar at all.
Additionally, we could log the strings we set with `pb.set_message` (either with `log` or `println`, tbd) in order to surface those as well.

I've tried using `SENTRY_NO_PROGRESS_BAR=1`, but the issue persists on current `master` when redirecting output to a file, indicating that to fix this the bar should not be constructed at all.

Here's a possible implementation of the proposal, to be refined/discussed https://github.com/getsentry/sentry-cli/pull/2830

lcian:
This seems related to https://github.com/getsentry/sentry-cli/issues/2790

Issue #2790, opened 2025-09-23
Logs are missing from `sourcemaps upload`


szokeasaurusrex:
Some debug logs are missing from `sentry-cli sourcemaps upload`

Specifically, when running with `--log-level=debug`, we would typically expect logs for all outgoing HTTP requests. However, the POST requests to `/<org>/chunk-upload/` and the final assemble endpoint calls appear to be missing from the logs.

### Sentry employees only 🔒

For further context, see the attached Slack thread on the [Linear issue](https://linear.app/getsentry/issue/CLI-171/logs-are-missing-from-sourcemaps-upload)

szokeasaurusrex:
#2829 may be the root cause of this issue

PR #2830, opened 2025-10-07
fix: Don't use hidden progress bar

lcian:
### Description

- Instead of using a hidden `indicatif` progress bar when `SENTRY_NO_PROGRESS_BAR`, use no progress bar instead, and just forward messages on it to `log::debug`/`info`.
- Also, use `std::io::stderr().is_terminal()` to determine whether or not to use a progress bar when `SENTRY_NO_PROGRESS_BAR` is not specified.
- This prevents the progress bar from "swallowing" messages that are set on it AND also those logged using `log`, as our `log` utility ultimately forwards to the progress bar when one is active.

### Issues

Close https://github.com/getsentry/sentry-cli/issues/2829
Close CLI-180
Close https://github.com/getsentry/sentry-cli/issues/2790
Close CLI-171

## Testing

1. `sentry-cli debug-files bundle-jvm --log-level=debug`

stdout and stderr before and after the change:
[stdout_before.txt](https://github.com/user-attachments/files/22850055/stdout_before.txt)
[stdout_after.txt](https://github.com/user-attachments/files/22850062/stdout_after.txt)
[stderr_before.txt](https://github.com/user-attachments/files/22850057/stderr_before.txt)
[stderr_after.txt](https://github.com/user-attachments/files/22850067/stderr_after.txt)

The stdout remains the same.
The stderr changes, now we see more output, such as the file being currently processed (DEBUG level), and other messages at INFO level, such as `Skipping ./File.java because it is not valid UTF-8` which is what motivated this PR in the first place.

2. `sentry-cli sourcemaps upload --log-level=debug`

stdout and stderr before and after the change:
[stdout_before.txt](https://github.com/user-attachments/files/22850640/stdout_before.txt)
[stdout_after.txt](https://github.com/user-attachments/files/22850641/stdout_after.txt)
[stderr_before.txt](https://github.com/user-attachments/files/22850643/stderr_before.txt)
[stderr_after.txt](https://github.com/user-attachments/files/22850644/stderr_after.txt)

The stdout remains the same as well.
The stderr now shows 2 `POST` calls to `/artifactbundle/assemble`, in contrast with the single `POST` call to that endpoint that we would see without the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Deprecated the `upload-proguard` subcommand's `--app-id`, `--version`, and `--version-code` flags ([#2852](https://github.com/getsentry/sentry-cli/pull/2852)), as we plan to remove these flags in Sentry CLI 3.x. Users should simply stop using the flags; the values specified there have never had an effect on deobfuscation, and are no longer visible in Sentry.
 - Added a deprecation notice for release bundle uploads, a legacy method for uploading sourcemaps ([#2844](https://github.com/getsentry/sentry-cli/pull/2844)). Release bundle uploads will be removed in Sentry CLI 3.x in favor of artifact bundles, the newer sourcemap upload method [introduced in Sentry version 23.6.2](https://github.com/getsentry/sentry/commit/f90f764fda09575f3f94caf32d04589098384616). **Self-hosted users**: You must upgrade to Sentry 23.6.2 or later before upgrading to Sentry CLI 3.x.
 
+### New Features
+
+- Add `projects` field to `SentryCliUploadSourceMapsOptions` ([#2856](https://github.com/getsentry/sentry-cli/pull/2856))
+
 ### Fixes
 
 - Don't use hidden progress bar ([#2830](https://github.com/getsentry/sentry-cli/pull/2830)) by @lcian
@@ -17,6 +21,7 @@
 ## 2.56.1
 
 ### Deprecations
+
 - Added a deprecation notice for legacy uploading methods ([#2836](https://github.com/getsentry/sentry-cli/pull/2836), [#2837](https://github.com/getsentry/sentry-cli/pull/2837))
   - Support for these legacy uploading methods, required to upload to self-hosted Sentry servers below version 10.0.0, will be removed in the next major release (3.x). If you observe these new deprecation notices, we recommend upgrading your self-hosted Sentry server, or pinning Sentry CLI to a compatible version (2.x).
   - You may encounter these deprecation notices when uploading debug files or sourcemaps.
@@ -43,7 +48,6 @@
 - fix: Validate `SENTRY_RELEASE` environment variable (#2807) by @szokeasaurusrex
 - fix: use actual PR head SHA in GitHub Actions instead of merge commit (#2785) by @runningcode
 - fix: suppress warning messages in failing build upload tests (#2791) by @runningcode
-
 
 ## 2.55.0
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -224,7 +224,7 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # 5.0.0
 
-      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # 5.0.0
+      - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # 6.0.0
         with:
           node-version: '20.10.0'
 
@@ -297,7 +297,7 @@ jobs:
     needs: [linux, sign-macos-binaries, windows]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # 5.0.0
-      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # 5.0.0
+      - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # 6.0.0
         with:
           node-version: '20.10.0'
       - uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # 5.0.0
--- a/.github/workflows/test_node.yml
+++ b/.github/workflows/test_node.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # 5.0.0
 
       - name: Use Node.js
-        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # 5.0.0
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # 6.0.0
         with:
           node-version-file: package.json
 
@@ -50,7 +50,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # 5.0.0
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # 5.0.0
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # 6.0.0
         with:
           node-version: ${{ matrix.node-version }}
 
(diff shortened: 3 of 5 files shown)


src/utils/progress.rs
```diff
@@ -1,76 +1,120 @@
...
+        if let Some(inner) = &self.inner {
+            inner.set_style(style);
+        }
+    }
+
+    pub fn set_message<S: AsRef<str>>(&self, msg: S) {
+        if let Some(inner) = &self.inner {
+            inner.set_message(msg.as_ref());
+        } else {
+            println!("{}", msg.as_ref());
```
lcian: Should probably be `log::debug` instead because we were not showing these anyway when they were getting swallowed

src/utils/progress.rs
```diff
@@ -1,76 +1,120 @@
...
-        logging::set_progress_bar(None);
+
+        if let Some(inner) = &self.inner {
+            let progress_style = ProgressStyle::default_bar().template("{prefix:.dim} {msg}");
+            inner.set_style(progress_style);
+            inner.set_prefix(">");
+            inner.finish_with_message(&msg);
+            logging::set_progress_bar(None);
+        } else {
+            println!("{}", msg);
```
lcian: This happens only once so I guess println is fine

src/utils/progress.rs
```diff
@@ -1,76 +1,120 @@
...
 use std::time::Instant;
 
 use crate::utils::logging;
 
 pub use indicatif::ProgressStyle;
 
-pub fn is_progress_bar_visible() -> bool {
-    env::var("SENTRY_NO_PROGRESS_BAR") != Ok("1".into())
+pub fn use_progress_bar() -> bool {
+    (env::var("SENTRY_NO_PROGRESS_BAR") != Ok("1".into())) || std::io::stdout().is_terminal()
```
szokeasaurusrex: We should verify whether `stdout` or `stderr` is used for the progress bar

lcian:
Without this PR we're indeed missing not only the messages emitted on the progress bar itself (with `pb.set_message`) but also those emitted with `log::` macros. This is because here we use the progress bar to print those messages too, if one is active:
see https://github.com/getsentry/sentry-cli/blob/master/src/utils/logging.rs#L92
So this will solve that problem too

src/commands/debug_files/find.rs
```diff
@@ -117,7 +117,7 @@ fn find_ids(
         .filter(|e| e.file_type().is_file());
 
     let mut found_files = vec![];
-    let pb = ProgressBar::new_spinner();
+    let mut pb = ProgressBar::new_spinner();
```
lcian: Needs to become `mut` to handle `set_prefix`

src/utils/progress.rs
lcian: It uses `stderr`, changed it.

src/utils/progress.rs
```diff
@@ -1,76 +1,141 @@
...
-pub fn is_progress_bar_visible() -> bool {
-    env::var("SENTRY_NO_PROGRESS_BAR") != Ok("1".into())
+/// By default, use the progress bar when we can detect we're running in a terminal.
+/// `SENTRY_NO_PROGRESS_BAR` takes precedence and can be used to disable the progress bar
+/// regardless of whether or not we're in a terminal.
+pub fn use_progress_bar() -> bool {
+    if env::var("SENTRY_NO_PROGRESS_BAR") == Ok("1".into()) {
+        return false;
+    }
+    std::io::stderr().is_terminal()
```
szokeasaurusrex, line 18: I've updated the PR description to also use `stderr`

src/utils/progress.rs
```diff
@@ -1,76 +1,141 @@
...
-        logging::set_progress_bar(None);
+
+        if let Some(inner) = &self.inner {
+            let progress_style = ProgressStyle::default_bar().template("{prefix:.dim} {msg}");
+            inner.set_style(progress_style);
+            inner.set_prefix(">");
+            inner.finish_with_message(&msg);
+            logging::set_progress_bar(None);
+        } else {
+            log::info!("> {msg}");
```
szokeasaurusrex: `log::info!` by default would be hidden (as log level is by default at `WARN`). I think `eprintln!` is therefore more suitable here, assuming these final messages typically show in the `stderr` when running in terminal

```suggestion
            eprintln!("> {msg}");
```

src/utils/progress.rs
```diff
@@ -1,76 +1,141 @@
...
+        } else {
+            log::debug!(
+                "{}{}",
+                self.prefix
+                    .as_deref()
+                    .map(|s| format!("{s} "))
+                    .unwrap_or_default(),
+                msg.as_ref()
+            );
         }
```
szokeasaurusrex: m: As this output would not remain visible for long in the terminal, I don't think it makes much sense to print it at all when we don't use a progress bar. Logs at debug level are not visible by default anyways. Unless we find a case where this could be useful, I would omit it.

```suggestion
        }
```

src/utils/progress.rs
```diff
@@ -1,76 +1,141 @@
...
+    }
 
-    fn deref(&self) -> &Self::Target {
-        &self.inner
+    pub fn set_prefix<S: AsRef<str>>(&mut self, prefix: S) {
+        if let Some(inner) = &self.inner {
+            inner.set_prefix(prefix.as_ref());
+        } else {
+            self.prefix = Some(prefix.as_ref().to_owned());
+        }
```
szokeasaurusrex: Going along with my previous suggestion, getting rid of logging in the no progress bar case lets us simplify here. We can probably even remove the `prefix` field from the struct.

```suggestion
        }
```

lcian: Yep, I've added it just for completeness.
But overall, at least for the commands I've tested (`upload-jvm` and `sourcemaps upload`), these messages don't seem to be too useful.
So I think the value of this is mostly in not swallowing, we don't really care that much about showing the "intermediate" messages.
I'll remove both the prefix and the printing of intermediate messages.

tests/integration/_cases/debug_files/upload/debug_files-upload-mixed-embedded-sources.trycmd
```diff
@@ -4,8 +4,17 @@ $ sentry-cli --log-level=debug debug-files upload --include-sources tests/integr
   INFO    [..] Loaded config from [CWD]/.sentryclirc
 ...
 > Found 1 debug information file (1 with embedded sources)
+  DEBUG   [..] trying to collect sources for SrcGenSampleApp.pdb
+  DEBUG   [..] Trying to add source file: /Users/matt/Code/temp/SrcGenSampleApp/SrcGenSampleApp/Program.cs
+  DEBUG   [..] Trying to add source file: /Users/matt/Code/temp/SrcGenSampleApp/SrcGenSampleApp/obj/Release/net6.0/SrcGenSampleApp.GlobalUsings.g.cs
+  DEBUG   [..] Trying to add source file: /Users/matt/Code/temp/SrcGenSampleApp/SrcGenSampleApp/obj/Release/net6.0/.NETCoreApp,Version=v6.0.AssemblyAttributes.cs
+  DEBUG   [..] Trying to add source file: /Users/matt/Code/temp/SrcGenSampleApp/SrcGenSampleApp/obj/Release/net6.0/SrcGenSampleApp.AssemblyInfo.cs
+  DEBUG   [..] Skipping embedded source file: /Users/matt/Code/temp/SrcGenSampleApp/SrcGenSampleApp/MySourceGenerator/MySourceGenerator.HelloSourceGenerator/Program.g.cs
+  DEBUG   [..] No sources found for SrcGenSampleApp.pdb
```
lcian, line 13: Not sure if we usually go to this level of detail in the snapshots.
There's probably other snapshots that would be updated by this change with similar DEBUG statements, such as for `sourcemaps upload`, but I suspect we don't see them because we're using `...` there.

lcian: Yeah just checked and we definitely use `...` there, let me know if we should consider updating them.

szokeasaurusrex: I mean, I think it is good that we are checking that these debug statements don't get swallowed. Let's keep this detail here for now, if the test becomes difficult to maintain, we can consider using `...` instead

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ## Unreleased
 
+### Fixes
+
+- Don't use hidden progress bar ([#2830](https://github.com/getsentry/sentry-cli/pull/2830)) by @lcian
+  - Fixed an issue where log messages would not show up when redirecting stderr to a file and using any subcommand that uses a progress bar, such as `sentry-cli debug-files bundle-jvm` and `sentry-cli sourcemaps upload`.
+
```
szokeasaurusrex: m: Let's move the "Fixes" section below the "Deprecations" section, as "Deprecations" contains information that users may need to take action on.

lcian: Good catch.

szokeasaurusrex (approved):
🚀

PR #2854, opened 2025-10-16
chore(preprod): Include pdfs in parsedassets directory

NicoHinderling:
PDFs were not being included which we want to have for complete size analysis in the launchpad service


NicoHinderling:
actually going to just continue omitting pdfs as per guidance of @noahsmartin

PR #2855, opened 2025-10-16
chore(preprod): Add image dedupping logic when creating ParsedAssets folder

NicoHinderling:
This is logic that we have in [the emerge ios code](https://github.com/EmergeTools/emerge/blob/main/ios/sizeAnalysis/CLI/Sources/Shared/AssetUtil.swift#L151) but not here. I noticed that when testing HackerNews, we had duplicate images showing up and because of that, duplicate image insights. This change fixes both

Before:
image

After:
image

### DIFF
--- a/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
+++ b/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
@@ -1,4 +1,5 @@
 import CoreGraphics
+import CryptoKit
 import Foundation
 import ImageIO
 import UniformTypeIdentifiers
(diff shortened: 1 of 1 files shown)


apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -78,6 +79,7 @@ enum AssetUtil {
         let (structuredThemeStore, assetKeys) = initializeCatalog(from: file)
 
         var images: [String: CGImage] = [:]
+        var seenImageKeys: [String: String] = [:]  // SHA256 -> imageId mapping
```
noahsmartin, line 82: This doesn't seem right to me, where is an equivalent mapping used in the emerge code?

NicoHinderling: I had some misunderstanding , we do logic to set the `isDuplicate` flag, but we don't actually do de-dupping in the parsing. The main misunderstanding I had was around how we will no longer be doing app thinning automatically for users, so I think I'm going to close this PR for now

PR #2859, opened 2025-10-20
chore(preprod): Save images as original file format to ParsedAssets

NicoHinderling:
ParsedAssets currently contains all the images we extracted but they are all saved as PNGs. We want to keep them in their original format to make sure that our image insights are as accurate possible, and having them converted to PNG can disrupt this

Appears to work and be processed by launchpad correctly from local testing

### DIFF
--- a/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
+++ b/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
@@ -131,10 +131,10 @@ enum AssetUtil {
             let imageId = UUID().uuidString
             
             let fileExtension = (renditionTypeName as NSString).pathExtension.lowercased()
-            let originalFormat = fileExtension.isEmpty ? "png" : fileExtension
             
-            if originalFormat != "svg", let unslicedImage = unslicedImage {
-                images[imageId] = (cgImage: unslicedImage, format: originalFormat)
+            // Skip files without an extension or SVGs
+            if !fileExtension.isEmpty && fileExtension != "svg", let unslicedImage = unslicedImage {
+                images[imageId] = (cgImage: unslicedImage, format: fileExtension)
             }
 
             let asset = AssetCatalogEntry(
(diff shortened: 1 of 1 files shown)


apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -293,6 +295,31 @@ enum AssetUtil {
...
+        case "pdf":
+            return UTType.pdf.identifier
+        case "svg":
+            return UTType.svg.identifier
+        case "tiff", "tif":
+            return UTType.tiff.identifier
+        case "bmp":
+            return UTType.bmp.identifier
+        default:
+            return UTType.png.identifier
```
trevor-e: Can this be more defensive? I'd like it to only handle types that we know, otherwise return `nil` as an optional.

apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -129,7 +129,13 @@ enum AssetUtil {
...
             let assetType = determineAssetType(key)
             let imageId = UUID().uuidString
-            images[imageId] = unslicedImage
+            
+            let fileExtension = (renditionTypeName as NSString).pathExtension.lowercased()
+            let originalFormat = fileExtension.isEmpty ? "png" : fileExtension
+            
+            if originalFormat != "svg", let unslicedImage = unslicedImage {
+                images[imageId] = (cgImage: unslicedImage, format: originalFormat)
+            }
```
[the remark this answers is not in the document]
NicoHinderling: we don't do any processing with the SVGs so there's no point in saving them into the ParsedAssets folder. We do however want to know about them so we intentionally still include them in the assets.json file

apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -129,7 +129,13 @@ enum AssetUtil {
             let (width, height, unslicedImage) = resolveImageDimensions(rendition, isVector)
             let assetType = determineAssetType(key)
             let imageId = UUID().uuidString
-            images[imageId] = unslicedImage
+            
+            let fileExtension = (renditionTypeName as NSString).pathExtension.lowercased()
+            let originalFormat = fileExtension.isEmpty ? "png" : fileExtension
```
trevor-e: Same, perhaps skip over files that don't have an extension.

trevor-e (approved):

PR #2864, opened 2025-10-22
chore(preprod): Improve error messaging for zip normalizing step in case of duplicate filename case

NicoHinderling:
If a user hits this case, they will have more context for what's going wrong


runningcode (approved):
LGTM!

PR #2879, opened 2025-10-28
chore(preprod): Add AssetCatalog Parsing support for Multisize sets

NicoHinderling:
Certain assets weren't getting picked up for the Firefox app, and the ones missing appeared to be multisize sets, like this

```
    {
        "vector": false,
        "type": 0,
        "imageId": "42D31467-E63B-49D8-8D9B-1B608059E344",
        "size": 236,
        "height": 1024,
        "name": "AppIcon_Alt_Color_Purple",
        "filename": "AppIcon_Alt_Color_Purple",
        "idiom": "phone",
        "width": 1024
    },
```

They won't have a a file extension either it seems (ex. `Assets.car/AppIcon_Alt_Color_Purple`). So we'll go ahead and save them as a (default type) png

### DIFF
--- a/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
+++ b/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
@@ -153,15 +153,32 @@ enum AssetUtil {
             let type = rendition.getUInt(forKey: "type") ?? 0
 
             let isVector = type == 9
-            let (width, height, unslicedImage) = resolveImageDimensions(rendition, isVector)
+            let isMultisizeImageSet = type == 1010
             let assetType = determineAssetType(key)
             let imageId = UUID().uuidString
-            
             let fileExtension = (renditionTypeName as NSString).pathExtension.lowercased()
             
-            // Skip files without an extension or SVGs
-            if !fileExtension.isEmpty && fileExtension != "svg", let unslicedImage = unslicedImage {
-                images[imageId] = (cgImage: unslicedImage, format: fileExtension)
+            var width: Int?
+            var height: Int?
+            var unslicedImage: CGImage?
+            
+            if isMultisizeImageSet {
+                // Look up the actual image rendition from the multisize set
+                if let result = findImageForMultisizeSet(rendition, key, assetKeys, structuredThemeStore) {
+                    unslicedImage = result.image
+                    width = result.width
+                    height = result.height
+                    images[imageId] = (cgImage: unslicedImage!, format: "png")
+                }
+            } else {
+                // Get image dimensions from regular rendition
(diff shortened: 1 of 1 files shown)


apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -153,15 +153,25 @@ enum AssetUtil {
...
+                    width = result.width
+                    height = result.height
+                }
+            }
+            
+            // Skip SVGs, but save images even if they don't have an extension (default to png)
+            if fileExtension != "svg", let unslicedImage = unslicedImage {
+                let format = fileExtension.isEmpty ? "png" : fileExtension
+                images[imageId] = (cgImage: unslicedImage, format: format)
             }
```
NicoHinderling, line 182: @trevor-e https://github.com/getsentry/sentry-cli/pull/2859#discussion_r2446303137

I know you previously mentioned that we should be defensive and not include images missing a file type but in this case we want to include some that indeed don't have one

trevor-e: We can keep the old defensive logic if we restructure this a bit:
```
if 
  unslicedImage == nil && className == "_CUIThemeMultisizeImageSetRendition", 
  let result = findImageForMultisizeSet(rendition, key, assetKeys, structuredThemeStore) {
    unslicedImage = result.image
    width = result.width
    height = result.height
    images[imageId] = (cgImage: unslicedImage, format: "png")
} else !fileExtension.isEmpty && fileExtension != "svg", let unslicedImage = unslicedImage {
    images[imageId] = (cgImage: unslicedImage, format: fileExtension)
}
```

apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -153,15 +153,23 @@ enum AssetUtil {
             let type = rendition.getUInt(forKey: "type") ?? 0
 
             let isVector = type == 9
-            let (width, height, unslicedImage) = resolveImageDimensions(rendition, isVector)
+            var (width, height, unslicedImage) = resolveImageDimensions(rendition, isVector)
```
trevor-e: Making this mutable and then later on depending on `unslicedImage` to be nil for the `findImageForMultisizeSet()` feels weird to me. Is there something more stable we can check to see if it's a multisize set, like the `type`? Similar to how we do `let isVector = type == 9`

trevor-e: Like I figured this would be structured like:
```
if isMultisizeImageSet {
 // logic
} else {
  // old logic
}
```

trevor-e (approved):

PR #2887, opened 2025-10-29
ref(vcs): Rename find_head -> find_head_sha

chromy:
This makes `find_head_sha` match its counterpart `find_base_sha`.


runningcode (approved):
Makes sense. This function was previously named `find_head` because it was used for the releases feature. Now that we have one for the sha and one for the ref this makes sense.

szokeasaurusrex (approved):

PR #2888, opened 2025-10-29
fix(vcs): Fix base_ref / base_sha for non-github actions

chromy:
This resolves EME-473

Both `find_base_sha` and `git_repo_base_ref` were not correct for non-Github workflow contexts.

`find_base_ref` returned previously returned a SHA where the intention of `find_base_ref`
was to return the nice name of the default merge branch of the remote. So in the common case:
```
---o---U---o---o---o---H
```

Where:
- `H` is `some-branch` and
- `U` is `origin/main` and
- `refs/remotes/origin/HEAD` is `refs/remotes/origin/main`

`git_repo_base_ref` should return `main`

`find_base_sha` did not handle the not Github case when it should return the equivalent of:
`git merge-base HEAD origin/HEAD`. For `find_base_sha` we want the merge base of
of `U` and `H`. In the above example this is the SHA of `U`. However they can be different:
e.g. 

```
    o---o---o---origin/main
   /
---1---o---o---o---foo
```

`find_base_ref` should still return `main`
but `find_base_sha` should be the SHA of `1`.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -3,7 +3,7 @@
 use std::fmt;
 use std::path::PathBuf;
 
-use anyhow::{bail, format_err, Context as _, Error, Result};
+use anyhow::{bail, format_err, Error, Result};
 use chrono::{DateTime, FixedOffset, TimeZone as _};
 use git2::{Commit, Repository, Time};
 use if_chain::if_chain;
@@ -272,30 +272,27 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
 }
 
 pub fn git_repo_base_ref(repo: &git2::Repository, remote_name: &str) -> Result<String> {
-    // Get the current HEAD commit
-    let head_commit = repo.head()?.peel_to_commit()?;
-
-    // Try to find the remote tracking branch
     let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
     let remote_ref = repo.find_reference(&remote_branch_name).map_err(|e| {
         anyhow::anyhow!("Could not find remote tracking branch for {remote_name}: {e}")
     })?;
 
-    find_merge_base_ref(repo, &head_commit, &remote_ref)
-}
+    let name = remote_ref
+        .resolve()?
+        .shorthand()
+        .ok_or(anyhow::anyhow!(
+            "Could not find remote tracking branch for {remote_name}"
+        ))?
+        .to_owned();
 
-fn find_merge_base_ref(
-    repo: &git2::Repository,
-    head_commit: &git2::Commit,
-    remote_ref: &git2::Reference,
-) -> Result<String> {
-    let remote_commit = remote_ref.peel_to_commit()?;
-    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
-
-    // Return the merge-base commit SHA as the base reference
-    let merge_base_sha = merge_base_oid.to_string();
-    debug!("Found merge-base commit as base reference: {merge_base_sha}");
-    Ok(merge_base_sha)
+    let expected_prefix = format!("{remote_name}/");
+    if let Some(branch_name) = name.strip_prefix(&expected_prefix) {
+        Ok(branch_name.to_owned())
+    } else {
+        Err(anyhow::anyhow!(
+            "Remote branch name '{name}' does not start with expected prefix '{expected_prefix}'"
+        ))
+    }
 }
 
 /// Like git_repo_base_repo_name but preserves the original case of the repository name.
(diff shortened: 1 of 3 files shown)


src/utils/vcs.rs
```diff
@@ -272,30 +272,27 @@ pub fn git_repo_head_ref(repo: &git2::Repository) -> Result<String> {
...
-
-    // Return the merge-base commit SHA as the base reference
-    let merge_base_sha = merge_base_oid.to_string();
-    debug!("Found merge-base commit as base reference: {merge_base_sha}");
-    Ok(merge_base_sha)
+    let name = remote_ref
+        .resolve()?
+        .shorthand()
+        .ok_or(anyhow::anyhow!(
+            "Could not find remote tracking branch for {remote_name}"
```
szokeasaurusrex, line 284: [`shorthand()` returns `None` if the shorthand is not valid UTF-8](https://docs.rs/git2/0.18/git2/reference/struct.Reference.html#method.shorthand).

```suggestion
            "Remote branch name is not valid UTF-8"
```

src/utils/vcs.rs
```diff
@@ -595,15 +610,19 @@ fn extract_pr_head_sha_from_event(json_content: &str) -> Option<String> {
...
+            return None;
+        }
+    };
 
-    Ok(v.pointer("/pull_request/base/sha")
+    v.pointer("/pull_request/base/sha")
         .and_then(|s| s.as_str())
-        .map(|s| s.to_owned()))
+        .map(|s| s.to_owned())
 }
```
szokeasaurusrex, line 626: I see no reason for this function to change. If we want to debug-log the error message rather than error, the calling code should change how it handles the error returned from this function.

```suggestion
/// Extracts the PR base SHA from GitHub Actions event payload JSON.
/// Returns Ok(None) if not a PR event or if SHA cannot be extracted.
/// Returns an error if we cannot parse the JSON.
fn extract_pr_base_sha_from_event(json_content: &str) -> Result<Option<String>> {
    let v: Value = serde_json::from_str(json_content)
        .context("Failed to parse GitHub event payload as JSON")?;

    Ok(v.pointer("/pull_request/base/sha")
        .and_then(|s| s.as_str())
        .map(|s| s.to_owned()))
}
```

szokeasaurusrex (commented):
Hey @chromy, I left a couple comments, but I am finding it difficult to review this change because it appears this PR is combining a bug fix (fixing `git_repo_base_ref`) and a new feature (adding support for non-GitHub environments to `find_base_sha`, which previously, was clearly only designed to handle GitHub environments), plus a minor refactor (folding `find_merge_base_ref` into `git_repo_base_ref`), all into one.

I would appreciate if you could split this change into two PRs (one for the bug fix, one for the new feature), or possibly three (by splitting the `find_merge_base_ref` refactor further into its own PR) so I can have a better idea of which changes go together 🙏

I am also confused because the PR description is mentioning a function named `find_base_ref`, which does not exist. I assume you meant `git_repo_base_ref`, but would appreciate if you can confirm this and update the PR description accordingly when making the split PRs.

src/utils/vcs.rs
chromy: I disagree, there are two functions:
1. `extract_pr_head_sha_from_event`
2. `extract_pr_base_sha_from_event`

They ought to match exactly except for which SHA they extract.
Previously they didn't have the same implementation or API `Result<Option<String>>` vs. `Option<String>`. It does not make sense for one to handle invalid json and not the other.

Since invalid json from the Github API is (hopefully) rare I don't think it makes sense to push handling the case on to callers (making all calling code more complicated)*. Better options would be:
- crashing
- returning None and logging

I would generally lean towards crashing but since `extract_pr_head_sha_from_event` already had that behavior I matched that.

*There is actually only one real caller so I was tempted to inline these and simplify the code (c.f. http://number-none.com/blow/john_carmack_on_inlined_code.html) but since it has some nice tests I left it.

I have split this into it's own PR here: https://github.com/getsentry/sentry-cli/pull/2893

src/utils/vcs.rs
chromy: Thanks! I've fixed this in the split patches.

chromy:
> Hey @chromy, I left a couple comments, but I am finding it difficult to review this change because it appears this PR is combining a bug fix (fixing `git_repo_base_ref`) and a new feature (adding support for non-GitHub environments to `find_base_sha`, which previously, was clearly only designed to handle GitHub environments), plus a minor refactor (folding `find_merge_base_ref` into `git_repo_base_ref`), all into one.
> 
> I would appreciate if you could split this change into two PRs (one for the bug fix, one for the new feature), or possibly three (by splitting the `find_merge_base_ref` refactor further into its own PR) so I can have a better idea of which changes go together 🙏
> 
> I am also confused because the PR description is mentioning a function named `find_base_ref`, which does not exist. I assume you meant `git_repo_base_ref`, but would appreciate if you can confirm this and update the PR description accordingly when making the split PRs.

Okay!

I've split them into three:
- https://github.com/getsentry/sentry-cli/pull/2893
- https://github.com/getsentry/sentry-cli/pull/2897
- https://github.com/getsentry/sentry-cli/pull/2898

chromy:
(closing since split up now)

PR #2889, opened 2025-10-29
chore(preprod): Reimplement multiset icon parsing

NicoHinderling:
Previously in https://github.com/getsentry/sentry-cli/pull/2879 I added the ability for us to collect multiset icons and have them included in the ParsedAssets folder / assets.json.

There were two problems:
1. we were only picking up the primary image of the set. For example, for a multiset called `AppIcon_Alt_Color_Red`, we'd only pick up the first image (`red-light-min.png`) when really it contains three, all of which we want to consider when analyzing for insights
2. We were saving the multiset image as the multiset object, rather than as an image, which led us to have to [add some undesirable logic in the launchpad service](https://github.com/getsentry/launchpad/pull/437/files#r2470576784)

With these changes, we now get all the images in the multisets and we save them as images

### DIFF
--- a/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
+++ b/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
@@ -114,10 +114,11 @@ enum AssetUtil {
 
         var images: [String: (cgImage: CGImage, format: String)] = [:]
         
-        // First pass: Build map of multisize sets and map element+idiom+subtype to the specific set
+        // First pass: Build map of multisize sets and cache renditions for performance
         var multisizeSets: [MultisizeSetInfo] = []
+        var renditionCache: [Int: NSObject] = [:]
         
-        for key in assetKeys {
+        for (index, key) in assetKeys.enumerated() {
             let keyList = unsafeBitCast(
                 key.perform(Selector(("keyList"))),
                 to: UnsafeMutableRawPointer.self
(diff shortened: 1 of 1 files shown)


apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -364,6 +359,30 @@ enum AssetUtil {
...
+        let element = key.getUInt(forKey: "themeElement") ?? 0
+        let identifier = key.getUInt(forKey: "themeIdentifier") ?? 0
+        let idiom = key.getUInt(forKey: "themeIdiom") ?? 0
+        let subtype = key.getUInt(forKey: "themeSubtype") ?? 0
+        
+        for setInfo in multisizeSets {
+            guard setInfo.element == element,
+                  setInfo.identifier == identifier else {
+                continue
+            }
```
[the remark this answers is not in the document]
NicoHinderling: This was the key insight to making it work 👎

apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -105,7 +113,51 @@ enum AssetUtil {
...
+        
+        // First pass: Build map of multisize sets and map element+idiom+subtype to the specific set
+        var multisizeSets: [MultisizeSetInfo] = []
+        
+        for key in assetKeys {
+            let keyList = unsafeBitCast(
+                key.perform(Selector(("keyList"))),
+                to: UnsafeMutableRawPointer.self
+            )
+            guard let rendition = createRendition(from: structuredThemeStore, keyList) else {
```
trevor-e, line 126: I think doing this in two passes makes sense, my only concern with this change is the performance since right now our ParsedAssets step is pretty slow. Is calling `createRendition()` twice expensive to do or should we cache those results?

NicoHinderling: great suggestion , implemented

PR #2890, opened 2025-10-29
feat(build): Add CLI version metadata to build upload archives

runningcode:
## Summary

Adds CLI version metadata to all zip archives created during build uploads for debugging purposes.

Resolves EME-471

## Changes

- Adds a `.sentry-cli-metadata.txt` file at the root of each build upload zip archive
- The file contains: `sentry-cli-version: <version>`
- Applied to both single-file uploads (APK, AAB) and directory uploads (XCArchive)
- Uses the same compression and timestamp settings as other files to maintain consistent checksums
- Adds support for `SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE` environment variable to override the version in tests and allow for deterministic chunk hashes
- The version override is checked at runtime (not just during `cfg(test)`), allowing integration tests that spawn the CLI as a subprocess to use the override

## Testing

- Integration tests use `SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE=0.0.0-test` to ensure stable chunk hashes across releases
- The environment variable is properly passed to CLI subprocesses in integration tests

### DIFF
--- a/src/utils/build/normalize.rs
+++ b/src/utils/build/normalize.rs
@@ -22,11 +22,9 @@ fn get_version() -> Cow<'static, str> {
 
     // Integration tests can override the version for consistent test results.
     // This ensures deterministic checksums in test fixtures by using a fixed version.
-    let version = std::env::var("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE")
+    std::env::var("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE")
         .map(Cow::Owned)
-        .unwrap_or(version);
-
-    version
+        .unwrap_or(version)
 }
 
 fn sort_entries(path: &Path) -> Result<impl Iterator<Item = (PathBuf, PathBuf)>> {


src/utils/build/normalize.rs
```diff
@@ -108,6 +109,12 @@ pub fn normalize_directory(path: &Path, parsed_assets_path: &Path) -> Result<Tem
         )?;
     }
 
+    let options = SimpleFileOptions::default()
```
runningcode: These are the same options as set in `upload.rs` (scroll up further in the file to see it)

tests/integration/_cases/build/build-upload-apk.trycmd
```diff
@@ -3,8 +3,7 @@ $ sentry-cli build upload tests/integration/_fixtures/build/apk.apk --head-sha t
 ? success
   WARN    [..] EXPERIMENTAL: The build subcommand is experimental. The command is subject to breaking changes and may be removed without notice in any release.
 > Preparing for upload completed in [..]
-> Uploading completed in [..]
```
szokeasaurusrex, line 5: m: I don't see why this line of output is disappearing following this change; I would expect that it should stay the same.

Please double-check why this is changing, and whether the change is intentional.

src/utils/build/normalize.rs
```diff
@@ -108,6 +109,12 @@ pub fn normalize_directory(path: &Path, parsed_assets_path: &Path) -> Result<Tem
...
+        .compression_method(zip::CompressionMethod::Deflated)
+        .last_modified_time(DateTime::default());
+    zip.start_file(".sentry-cli-metadata.txt", options)?;
+    writeln!(zip, "sentry-cli-version: {VERSION}")?;
```
szokeasaurusrex: l: To ensure we always generate the version information in the same way, I would suggest we extract this logic to a separate method, that we can also reuse in the `commands/build/upload.rs` file

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Various fixes & improvements
```
szokeasaurusrex:
```suggestion
### Features
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+
+- Add CLI version metadata to build upload archives for debugging purposes ([#2890](https://github.com/getsentry/sentry-cli/pull/2890))
```
szokeasaurusrex: Please add more details to this changelog entry, specifically so that users understand the user-facing changes. I have been making user-readable changelog entries a priority in the last few Sentry CLI releases.

tests/integration/_cases/build/build-upload-apk.trycmd
```diff
@@ -3,8 +3,7 @@ $ sentry-cli build upload tests/integration/_fixtures/build/apk.apk --head-sha t
...
 Successfully uploaded 1 file to Sentry
-  - tests/integration/_fixtures/build/apk.apk (http[..]/wat-org/preprod/wat-project/42)
+  - tests/integration/_fixtures/build/apk.apk (http://sentry.io/wat-org/preprod/wat-project/42)
```
szokeasaurusrex: In any case, this line should remain the same

```suggestion
  - tests/integration/_fixtures/build/apk.apk (http[..]/wat-org/preprod/wat-project/42)
```

szokeasaurusrex (commented):
Added some comments/suggestions. 

Also just wondering, were you able to test the changes manually? It would be cool if we could also have an automated test to check that the file is created and added to the ZIP, so we can ensure the behavior remains in CI

tests/integration/_cases/build/build-upload-apk.trycmd
runningcode: It turns out this was missing because the test expectations weren’t updated with the version number in the uploaded zip in the tests.

I fixed both the sha and expected binary in the tests and this worked.

src/utils/build/normalize.rs
runningcode: Done

CHANGELOG.md
runningcode: I adjusted the changelog slightly. Can you double check it? This isn’t necessarily a feature that users will use so I was unsure exactly how to address this.

tests/integration/_cases/build/build-upload-apk.trycmd
runningcode: Yup, See comment here https://github.com/getsentry/sentry-cli/pull/2890#discussion_r2487042226

runningcode:
So yes I was able to test the changes manually before I originally submitted it by making a change that would print the temporary zip directory before uploading so that I could extract it and check that the version number was there.

There’s no test that explicitly checks for the version number but the version number is now part of the new expected binary results and the sha of the expected test results. Does that work? Or should we have an explicit test for the version code?

szokeasaurusrex:
@runningcode I just realized that this test is going to be a pain to maintain. As the embedded version number will change with every release of Sentry CLI, we would need to update the SHA on every release.

I am not sure what the best way forward is, what do you think?

src/utils/build/normalize.rs
```diff
@@ -15,6 +16,10 @@ use walkdir::WalkDir;
 use zip::write::SimpleFileOptions;
 use zip::{DateTime, ZipWriter};
 
+fn get_version() -> String {
+    std::env::var("SENTRY_CLI_VERSION_OVERRIDE").unwrap_or_else(|_| VERSION.to_owned())
+}
```
szokeasaurusrex: Two thoughts:

  - m: We should only read `SENTRY_CLI_VERSION_OVERRIDE` in a test environment
  - l: We should return a `Cow<'static, str>` instead of a `String`. This way, in the most common case, where we return the `VERSION` constant, we don't need to copy the `VERSION`, since `Cow` can wrap borrowed or owned data.

```suggestion
fn get_version() -> Cow<'static, str> {
    let version = Cow::Borrowed(VERSION);

    #[cfg(test)]
    let version = std::env::var("SENTRY_CLI_VERSION_OVERRIDE")
        .map(Cow::Owned)
        .unwrap_or(version);

    version
}
```

szokeasaurusrex: I'd also consider adding a documentation or code comment for why this environment variable exists.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Promote `sentry-cli build upload` command to non-experimental ([#2899](https://github.com/getsentry/sentry-cli/pull/2899), [#2905](https://github.com/getsentry/sentry-cli/pull/2905)).
 
+### Various fixes & improvements
+
+- Build uploads now include CLI version metadata in the uploaded archive ([#2890](https://github.com/getsentry/sentry-cli/pull/2890))
```
szokeasaurusrex, line 8: m: Let's include this under "New Features." "Various fixes & improvements" is the auto-generated heading, but we are moving to writing the changelog manually in Sentry CLI to make it more useful for end-users

```suggestion
- Build uploads now include CLI version metadata in the uploaded archive ([#2890](https://github.com/getsentry/sentry-cli/pull/2890))
```

szokeasaurusrex (commented):
Couple small things, looking good though!

src/utils/build/normalize.rs
```diff
@@ -15,6 +17,17 @@ use walkdir::WalkDir;
...
+fn get_version() -> Cow<'static, str> {
+    let version = Cow::Borrowed(VERSION);
+
+    #[cfg(test)]
+    let version = std::env::var("SENTRY_CLI_VERSION_OVERRIDE")
+        .map(Cow::Owned)
+        .unwrap_or(version);
+
+    version
+}
```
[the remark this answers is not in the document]
runningcode: @szokeasaurusrex I believe this comment is accurate which is why I had the code structured as is. 
Is it okay to have the `SENTRY_CLI_VERSION_OVERRIDE` env override in production code?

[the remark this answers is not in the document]
szokeasaurusrex: I didn't realize that `#[cfg(test)]` is unavailable in integration tests. In that case, I'd say that it's ok that we keep the override in production code (I don't really love that users could stumble across it and misuse it, but I suppose the likelihood of that is low). Perhaps, we could modify the environment variable to `SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE` to indicate that the variable exists only for integration tests.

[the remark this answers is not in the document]
runningcode: Ok, removed!

src/utils/build/normalize.rs
```diff
@@ -15,6 +17,16 @@ use walkdir::WalkDir;
...
 use zip::{DateTime, ZipWriter};
 
+fn get_version() -> Cow<'static, str> {
+    let version = Cow::Borrowed(VERSION);
+
+    // Integration tests can override the version for consistent test results.
+    // This ensures deterministic checksums in test fixtures by using a fixed version.
+    std::env::var("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE")
+        .map(Cow::Owned)
+        .unwrap_or(version)
```
szokeasaurusrex, line 27: l: I think the following is simpler; the reason I originally suggested introducing the `version` variable was because we needed it with the `#[cfg(test)]`


```suggestion
    // Integration tests can override the version for consistent test results.
    // This ensures deterministic checksums in test fixtures by using a fixed version.
    std::env::var("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE")
        .map(Cow::Owned)
        .unwrap_or(Cow::Borrowed(VERSION))
```

tests/integration/build/upload.rs
```diff
@@ -176,6 +176,7 @@ fn command_build_upload_apk_chunked() {
             .expect(2),
         )
         .register_trycmd_test("build/build-upload-apk.trycmd")
+        .env("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE", "0.0.0-test")
```
szokeasaurusrex, line 179: l: I would also put a comment here

```suggestion
        // We override the version in the metadata field to ensure a consistent checksum
        // for the uploaded files.
        .env("SENTRY_CLI_INTEGRATION_TEST_VERSION_OVERRIDE", "0.0.0-test")
```

szokeasaurusrex:
welp, this auto-merged, I guess the above comments are not that important anyhow

runningcode:
Thank you. I can do both in a follow up.

PR #2893, opened 2025-10-30
ref(vcs): Make extract_pr_{base,head}_sha_from_event match

chromy:
Make functions `extract_pr_{base,head}_sha_from_event` match exactly except for
SHA they extract. Previously they have different APIs (`Result<Option<String>>` vs
`Option<String>`). We choose `Option<String>` since there seems little point forcing
users to handle the case of Github returning invalid json. Better options are:
- Returning None
- Crashing (`panic!`)

We switch both `extract_pr_head_sha_from_event` and
`extract_pr_base_sha_from_event` to panic.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -575,17 +575,17 @@ pub fn find_base_sha() -> Result<Option<String>> {
         .and_then(|event_path| std::fs::read_to_string(event_path).map_err(Error::from))
         .context("Failed to read GitHub event path")?;
 
-    extract_pr_base_sha_from_event(&github_event)
+    Ok(extract_pr_base_sha_from_event(&github_event))
 }
 
 /// Extracts the PR head SHA from GitHub Actions event payload JSON.
 /// Returns None if not a PR event or if SHA cannot be extracted.
+/// Panics if json is malformed.
 fn extract_pr_head_sha_from_event(json_content: &str) -> Option<String> {
     let v: Value = match serde_json::from_str(json_content) {
         Ok(v) => v,
         Err(_) => {
-            debug!("Failed to parse GitHub event payload as JSON");
-            return None;
+            panic!("Failed to parse GitHub event payload as JSON");
         }
     };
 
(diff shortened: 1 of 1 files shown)


chromy:
> One reason we might wish to crash instead of just logging a debug message: If GitHub ever changes the schema of the JSON event payload, in a way that this function would fail to extract the SHA, this would fail silently (debug messages are not logged at all by default, as the default log level is `WARN`). But, I'll leave it up to you to decide which way to go.

sg! Is it enough to call `panic!` (I have updated the code to do that) or is there a fancier way to hard error?

src/utils/vcs.rs
```diff
@@ -575,17 +575,17 @@ pub fn find_base_sha() -> Result<Option<String>> {
...
+/// Panics if json is malformed.
 fn extract_pr_head_sha_from_event(json_content: &str) -> Option<String> {
     let v: Value = match serde_json::from_str(json_content) {
         Ok(v) => v,
         Err(_) => {
-            debug!("Failed to parse GitHub event payload as JSON");
-            return None;
+            panic!("Failed to parse GitHub event payload as JSON");
         }
     };
```
szokeasaurusrex, line 590: You can simplify this by using `expect`:

```suggestion
    let v: Value =
        serde_json::from_str(json_content).expect("Failed to parse GitHub event payload as JSON");
```

Internally, `expect` panics using the provided message if `serde_json::from_str` returns an error. So the behavior is roughly the same as what you wrote, but much simpler (and, I believe the error message returned from `serde_json::from_str` also gets included in the output, unlike in your version).

src/utils/vcs.rs
```diff
@@ -595,15 +595,19 @@ fn extract_pr_head_sha_from_event(json_content: &str) -> Option<String> {
...
-        .context("Failed to parse GitHub event payload as JSON")?;
+/// Returns None if not a PR event or if SHA cannot be extracted.
+/// Panics if json is malformed.
+fn extract_pr_base_sha_from_event(json_content: &str) -> Option<String> {
+    let v: Value = match serde_json::from_str(json_content) {
+        Ok(v) => v,
+        Err(_) => {
+            panic!("Failed to parse GitHub event payload as JSON");
+        }
+    };
```
```suggestion
    let v: Value =  
        serde_json::from_str(json_content).expect("Failed to parse GitHub event payload as JSON");
```

src/utils/vcs.rs
```diff
@@ -1594,16 +1598,6 @@ mod tests {
...
-        let malformed_json = r#"{
-  "pull_request": {
-    "id": 789,
-    "head": {
-      "ref": "feature-branch"
-    }
-  }
-}"#;
-
-        assert_eq!(extract_pr_head_sha_from_event(malformed_json), None);
```
szokeasaurusrex, line 1606: I am a bit confused, why are we removing this test case?

src/utils/vcs.rs
```diff
@@ -1719,10 +1711,7 @@ mod tests {
         })
         .to_string();
 
-        assert_eq!(extract_pr_base_sha_from_event(&push_json).unwrap(), None);
-
-        // Test with malformed JSON
-        assert!(extract_pr_base_sha_from_event("invalid json {").is_err());
```
szokeasaurusrex, line 1725: We should still test that the function handles invalid JSON correctly by panicking. You can do this by splitting this test into its own test function, and by [marking the separate test function with `#[should_panic]`](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic).

src/utils/vcs.rs
chromy: Thanks, done.

src/utils/vcs.rs
chromy: Thanks, done!

src/utils/vcs.rs
chromy: I thought the json was invalid like the other one for some reason. I've put this back.

PR #2897, opened 2025-10-30
fix(vcs): Handle non-github workflow in find_base_sha

chromy:
`find_head_sha` supported both Github workflow contexts and 'raw git'
contexts however the matching function `find_base_sha` did not handle
raw git contexts.

Thi updates `find_base_sha` to return the equivalent of:
`git merge-base HEAD origin/HEAD`. For:

```
     o---o---o---origin/main
    /
---1---o---o---o---foo
```

We return the SHA of 1.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -588,11 +588,12 @@ pub fn find_base_sha(remote_name: &str) -> Result<Option<String>> {
         .find_reference(&remote_branch_name)
         .with_context(|| "Could not find default branch for {remote_name}")?;
 
-    let remote_commit = remote_ref.peel_to_commit()?;
-    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
-    let merge_base_sha = merge_base_oid.to_string();
-    debug!("Found merge-base commit as base reference: {merge_base_sha}");
-    Ok(Some(merge_base_sha))
+    Ok(remote_ref
+        .peel_to_commit()
+        .and_then(|remote_commit| repo.merge_base(head_commit.id(), remote_commit.id()))
+        .map(|oid| oid.to_string())
+        .ok()
+        .inspect(|sha| debug!("Found merge-base commit as base reference: {sha}")))
 }
 
 /// Extracts the PR head SHA from GitHub Actions event payload JSON.


src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
     Ok(head.id().to_string())
 }
 
-pub fn find_base_sha() -> Result<Option<String>> {
-    let github_event = std::env::var("GITHUB_EVENT_PATH")
-        .map_err(Error::from)
-        .and_then(|event_path| std::fs::read_to_string(event_path).map_err(Error::from))
-        .context("Failed to read GitHub event path")?;
+pub fn find_base_sha(remote_name: &str) -> Result<Option<String>> {
```
szokeasaurusrex, line 572: m: As we no longer ever return `Ok(None)` from this function, we should simplify the signature to return `Result<String>`.

```suggestion
pub fn find_base_sha(remote_name: &str) -> Result<String> {
```

src/utils/vcs.rs
```diff
@@ -3,7 +3,7 @@
 use std::fmt;
 use std::path::PathBuf;
 
-use anyhow::{bail, format_err, Context as _, Error, Result};
+use anyhow::{bail, format_err, Error, Result};
```
szokeasaurusrex: We will need to bring back this import for one of my suggestions below, specifically the one about using `with_context`.

```suggestion
use anyhow::{bail, format_err, Context as _, Error, Result};
```

src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
...
+    }
+
+    let repo = git2::Repository::open_from_env()?;
+
+    let head_commit = repo.head()?.peel_to_commit()?;
+
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = repo
+        .find_reference(&remote_branch_name)
+        .map_err(|e| anyhow::anyhow!("Could default branch for {remote_name}: {e}"))?;
```
szokeasaurusrex: h: This error message has a typo; I think you meant to write "Could not find ..."

m: Rather than manually mapping the error, we should use [`anyhow`'s `with_context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) here.

Both suggestions here:

```suggestion
        .with_context(|| "Could not find default branch for {remote_name}")?;
```

src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
...
+    if let Some(pr_base_sha) = std::env::var("GITHUB_EVENT_PATH")
+        .ok()
+        .and_then(|event_path| std::fs::read_to_string(event_path).ok())
+        .and_then(|content| extract_pr_base_sha_from_event(&content))
+    {
+        debug!("Using GitHub Actions PR base SHA from event payload: {pr_base_sha}");
+        return Ok(Some(pr_base_sha));
+    }
+
+    let repo = git2::Repository::open_from_env()?;
```
szokeasaurusrex: l: Let's add context to improve this error message; I don't think these git2 error messages tend to be very human-readable

```suggestion
    let repo = git2::Repository::open_from_env().context("Could not open repository")?;
```

src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
...
+
+    let head_commit = repo.head()?.peel_to_commit()?;
```
szokeasaurusrex, line 584: l: Context for better error message:

```suggestion
    let head_commit = repo.head()
        .and_then(|h| h.peel_to_commit())
        .context("Could not find head commit")?;
```

src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
...
+
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = repo
+        .find_reference(&remote_branch_name)
+        .map_err(|e| anyhow::anyhow!("Could default branch for {remote_name}: {e}"))?;
 
-    Ok(extract_pr_base_sha_from_event(&github_event))
+    let remote_commit = remote_ref.peel_to_commit()?;
+    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
```
szokeasaurusrex: l: Would also add context for these

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Deprecated the `upload-proguard` subcommand's `--platform` flag ([#2863](https://github.com/getsentry/sentry-cli/pull/2863)). This flag appears to have been a no-op for some time, so we will remove it in the next major.
 
+### Fixes
+
+- Fix auto filled git metadata when using the `build upload` subcommand in non-github workflow contexts ([#2897](https://github.com/getsentry/sentry-cli/pull/2897))
```
szokeasaurusrex: m: Can you please add some more detail to this changelog entry/make it more user-friendly? E.g. by being more specific of what metadata we can now add in non-GH workflow contexts? I have been putting more effort into creating human-readable changelog entries the past few releases.

l: Also, I would consider this more of a new feature than a fix; can we put it in a `### New Features` section? (and if we do, please move the section above deprecations)

Maybe this works:

```suggestion

### New Features

- `build upload` now supports automatic git base commit detection in local development and all CI environments, not just GitHub Actions ([#2897](https://github.com/getsentry/sentry-cli/pull/2897))
```

szokeasaurusrex (commented):
Looks good for the most part, did notice a few improvements (and a typo), however.

Note: I am using the [LOGAF scale](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale) in my comments

CHANGELOG.md
chromy: I've been more specific in the message. I don't think it's a feature. This was intended behaviour since this was introduced (as seen in the `head` code and also in the calling code, that it did not work previously is a bug.

src/utils/vcs.rs
chromy: done

src/utils/vcs.rs
chromy: The intention is that if the merge-base could not be found we return None. If that is not what the Rust code:
```
    let remote_commit = remote_ref.peel_to_commit()?;
    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
```
does could you suggest the best way to make it do that?

src/utils/vcs.rs
chromy: As above: the intention is that if the merge-base could not be found we return None.

I don't think there is value in adding context to `peel_to_commit()`

src/utils/vcs.rs
szokeasaurusrex: No, that is not what this code does. The `?` will propagate any `Err` returned by these functions, causing `find_base_sha` to return `Err(<whatever error from the method>)`.

Will add my suggestion for how to do this below

src/utils/vcs.rs
```diff
@@ -569,13 +569,30 @@ pub fn find_head_sha() -> Result<String> {
...
+    let remote_branch_name = format!("refs/remotes/{remote_name}/HEAD");
+    let remote_ref = repo
+        .find_reference(&remote_branch_name)
+        .with_context(|| "Could not find default branch for {remote_name}")?;
+
+    let remote_commit = remote_ref.peel_to_commit()?;
+    let merge_base_oid = repo.merge_base(head_commit.id(), remote_commit.id())?;
+    let merge_base_sha = merge_base_oid.to_string();
+    debug!("Found merge-base commit as base reference: {merge_base_sha}");
+    Ok(Some(merge_base_sha))
```
szokeasaurusrex: This should do the trick (key thing to note here is the `.ok()` which converts a `Result` to an `Option`, discarding any `Err` value and replacing it with `None`)

```suggestion
    Ok(remote_ref
        .peel_to_commit()
        .and_then(|remote_commit| repo.merge_base(head_commit.id(), remote_commit.id()))
        .map(|oid| oid.to_string())
        .ok()
        .inspect(|sha| debug!("Found merge-base commit as base reference: {sha}")))
```

src/utils/vcs.rs
szokeasaurusrex: @chromy are you also wanting to return `None` if this call fails?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Deprecated the `upload-proguard` subcommand's `--platform` flag ([#2863](https://github.com/getsentry/sentry-cli/pull/2863)). This flag appears to have been a no-op for some time, so we will remove it in the next major.
 - Deprecated the `upload-proguard` subcommand's `--android-manifest` flag ([#2891](https://github.com/getsentry/sentry-cli/pull/2891)). This flag appears to have been a no-op for some time, so we will remove it in the next major.
 
+### Fixes
+
+- Fix autofilled git base metadata (`--base-ref`, `--base-sha`) when using the `build upload` subcommand in git repos. Previously this worked only in the contexts of github workflows ([#2897](https://github.com/getsentry/sentry-cli/pull/2897))
```
szokeasaurusrex: nit: "github" -> "GitHub" and final period

```suggestion
- Fix autofilled git base metadata (`--base-ref`, `--base-sha`) when using the `build upload` subcommand in git repos. Previously this worked only in the contexts of GitHub workflows ([#2897](https://github.com/getsentry/sentry-cli/pull/2897)).
```

src/utils/vcs.rs
chromy: No I think this one should probably `Error`. It's legitimate that there might not be a merge-base for two commits, you can get it if the history is separate e.g.
```
$ git commit -m'foo' --allow-empty
[main (root-commit) fa819e8] foo
$ git commit -m'bar' --allow-empty
[main c35d2ee] bar
$ git checkout --orphan baz
Switched to a new branch 'baz'
$ git commit -m'baz' --allow-empty
[baz (root-commit) 27134ca] baz
$ git merge-base main baz
$ git merge-base HEAD baz
27134ca5b445526335d7a8d955790fac1e5d3690
```

This one:
```
$ git merge-base main baz
```

returns nothing because there is no common commit in the history of the `main` branch and the orphan `baz` branch.

However I don't think there can be a legitimate case where HEAD doesn't point at a commit. I think that will only happen if the repo is in quite a corrupt state.

src/utils/vcs.rs
chromy: Thanks!

PR #2908, opened 2025-11-04
test(vcs): Fix local find_base_sha test

chromy:
The following test:
```
// Test without GITHUB_EVENT_PATH
std::env::remove_var("GITHUB_EVENT_PATH");
let result = find_base_sha("orgin");
assert!(result.is_err());
```
can't really work well both in CI and locally.

Locally there will (almost) always be a merge-base between the local HEAD and the default
merge branch. In GitHub CI there will almost never be a merge-base which is why we need
to inspect `GITHUB_EVENT_PATH` sometimes in the first place.

For now replace that with:
```
// Test without GITHUB_EVENT_PATH
std::env::remove_var("GITHUB_EVENT_PATH");
let result = find_base_sha("no_such_branch");
assert!(result.is_err());
```

Which tests a different code path which should fail on both.

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -1795,7 +1795,7 @@ mod tests {
 
         // Test without GITHUB_EVENT_PATH
         std::env::remove_var("GITHUB_EVENT_PATH");
-        let result = find_base_sha("origin");
+        let result = find_base_sha("no_such_remote");
         assert!(result.is_err());
     }
 }


src/utils/vcs.rs
```diff
@@ -1795,7 +1795,7 @@ mod tests {
 
         // Test without GITHUB_EVENT_PATH
         std::env::remove_var("GITHUB_EVENT_PATH");
-        let result = find_base_sha("origin");
+        let result = find_base_sha("no_such_remote");
```
szokeasaurusrex, line 1798: Ah ok, so this test will pass locally as long as the user hasn't configured a `no_such_remote` git remote? 

Def agree with what you wrote on Slack that this is not ideal. Would be nice to make the tests not depend on local git state, but this is fine as a workaround I suppose

szokeasaurusrex (approved):
lgtm — Please consider renaming the PR to "test(vcs): ..." as this is only changing test code (that change should also make the Danger CI failure go away, as changelog is not required for "test" changes; you might need to manually retrigger the Danger check)

runningcode (approved):

src/utils/vcs.rs
chromy: yep exactly

chromy:
> lgtm — Please consider renaming the PR to "test(vcs): ..." as this is only changing test code (that change should also make the Danger CI failure go away, as changelog is not required for "test" changes; you might need to manually retrigger the Danger check)

Thanks done!

PR #2910, opened 2025-11-04
chore: Prepare to use TypeScript natively

JPeer264:
### Description

This PR is a starting point of upgrading to TypeScript and preparing for future upgrades around testing and maintenance.

I've seen that couple of types between the `*.js` files and the `index.d.ts` were not 100% in sync, which is one of the reasons the move to TypeScript natively could help here. 

### What's new / changed

- moved `js` to `lib` in order to keep `js` as the same output folder which is delivered to NPM
- `js/index.d.ts` stays as type export for NPM and the types are re-exported via the jsdoc `@typedef`
- `types.ts` is actually the same as the previous `index.d.ts` excluding the `SentryCli` class
- TypeScript is partially used to keep the changes minimal
- I aligned the jsdoc types with the actual text from the previous `index.d.ts`


package.json
```diff
@@ -29,20 +29,23 @@
...
     "@sentry/cli-linux-x64": "2.57.0",
+    "@sentry/cli-win32-arm64": "2.57.0",
     "@sentry/cli-win32-i686": "2.57.0",
-    "@sentry/cli-win32-x64": "2.57.0",
-    "@sentry/cli-win32-arm64": "2.57.0"
+    "@sentry/cli-win32-x64": "2.57.0"
   },
   "scripts": {
-    "postinstall": "node ./scripts/install.js",
+    "postinstall": "npm run build && npm run install-cli",
```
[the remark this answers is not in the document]
JPeer264: Fair enough. I have to think about this one.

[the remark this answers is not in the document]
JPeer264: Resolved by removing it. There is only one issue though. When this repo is getting checked out locally we need to run `npm i --ignore-scripts`. There is currently no documentation how to run the node part. In a follow up PR I will add some documentation in the `CONTRIBUTING.md` to make it clear.

JPeer264:
@andreiborza this is now ready to review. Had some issues with the `postinstall`, but this is solved now.

.github/workflows/build.yml
```diff
@@ -228,6 +228,9 @@ jobs:
         with:
           node-version: '20.10.0'
 
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
```
szokeasaurusrex, line 233: Just wondering, why do we need to install dependencies here, when we were not doing that previously?

.github/workflows/test_node.yml
```diff
@@ -37,7 +37,30 @@ jobs:
...
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Build
+        run: npm run build
+
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # 4.6.2
+        with:
+          name: artifact-build
+          path: js-dist
```
szokeasaurusrex: What does this do/why do we need it now, when we didn't before? (I'm a JS noob so I'm just asking so I understand, not because I'm trying to question the necessity of this change 😅)

.github/workflows/test_node.yml
```diff
@@ -65,6 +88,15 @@ jobs:
...
         run: npm install --omit=optional --ignore-scripts
 
+      - name: Download build artifact
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # 5.0.0
+        with:
+          name: artifact-build
+          path: js-dist
+
+      - name: Install CLI
+        run: npm run install-cli
```
szokeasaurusrex, line 98: If I understand correctly, we need this `install-cli` step now because we have added `--ignore-scripts` previously?

js/releases/index.js
```diff
@@ -59,21 +65,8 @@ class Releases {
...
-   * @param {string} options.commit The current (last) commit in the release.
-   * @param {string} options.previousCommit The commit before the beginning of this
-   * release (in other words, the last commit of the previous release). If omitted,
-   * this will default to the last commit of the previous release in Sentry. If there
-   * was no previous release, the last 10 commits will be used.
-   * @param {boolean} options.ignoreMissing When the flag is set and the previous release
-   * commit was not found in the repository, will create a release with the default commits
-   * count (or the one specified with `--initial-depth`) instead of failing the command.
-   * @param {boolean} options.ignoreEmpty When the flag is set, command will not fail
-   * and just exit silently if no new commits for a given release have been found.
```
szokeasaurusrex, line 75: Removing all of these parameters looks to me like a breaking change. Am I missing something?

JPeer264: It may look like they got removed. But in practice they are now available in `SentryCliCommitsOptions`, so they are defined once in there: https://github.com/getsentry/sentry-cli/pull/2910/files#diff-fe8dd47b473b25f6d7261cd2682b82e73e6e6861a1c4bd7aa3599ab0ff1577b6R189-R217

.github/workflows/test_node.yml
JPeer264: Exactly.

js/releases/index.js
szokeasaurusrex: Yes, I understand they are available in the options. But, if I was previously calling the function like `foo(a, b, c, d)`, can I still call it like that, or do I now need to call it like `foo(Options{ a, b, c, d })`?

.github/workflows/test_node.yml
JPeer264: Right now it is not doing a lot, since we don't have any actual `.ts` files yet. So ATM it only generates the `.d.ts` files on top and the `.js` files look almost the same in the output (`js-dist`). So the benefit of now is that we don't have to maintain manually the `index.d.ts`, since this one is now generated.

Long term we move `.js` files to `.ts` where we have better type safety. Long long term we could theoretically produce cjs and esm entry points.

.github/workflows/build.yml
JPeer264: We do need to call `tsc` to move everything into the `js-dist` folder. We are calling `npm pack` which runs its prehook: `prepack`. This needs at least `tsc` to be installed in order to be shipped.

JPeer264:
> My one concern is: several of the changes here look like they might break the public JS API; am I understanding that correctly?

If you run `npm install --ignore-scripts` and `npm run build` locally - you'll see that the `index.d.ts` inside `js-dist` is exporting the same types as before (just a slightly different format). So there shouldn't be any breaking changes regarding the API.

.github/workflows/test_node.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install dependencies via npm ci
         if: ${{ !inputs.triggered-by-release }}
-        run: npm ci
+        run: npm ci --ignore-scripts
```
andreiborza, line 29: l: Maybe we could add a comment here for why we ignore scripts, wdyt?

js/releases/index.js
```diff
@@ -45,9 +52,8 @@ class Releases {
    * upload artifacts, such as source maps.
    *
    * @param {string} release Unique name of the new release.
-   * @param {object} options A set of options when creating a release.
-   * @param {array} options.projects The list of project slugs for a release.
-   * @returns {Promise} A promise that resolves when the release has been created.
+   * @param {{projects?: string[]}} [options] A set of options when creating a release, or an array of project slugs.
```
andreiborza: q: Why did this become an "or"?

lib/releases/index.js
```diff
@@ -153,8 +146,8 @@ class Releases {
    * });
    *
    * @param {string} release Unique name of the release.
-   * @param {object} options Options to configure the source map upload.
-   * @returns {Promise} A promise that resolves when the upload has completed successfully.
+   * @param {SentryCliUploadSourceMapsOptions & {live?: boolean | 'rejectOnError'}} options Options to configure the source map upload.
```
andreiborza, line 149: l: I think we could life this up into `SentryCliUploadSourceMapsOptions`. Doesn't have to be in this PR just fyi.

lib/types.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  setCommits(release: string, options: SentryCliCommitsOptions): Promise<string>;
+
+  finalize(release: string): Promise<string>;
+
+  proposeVersion(): Promise<string>;
+
+  uploadSourceMaps(
+    release: string,
+    options: SentryCliUploadSourceMapsOptions & { live?: boolean | 'rejectOnError' }
```
andreiborza, line 235: l: Same here, we should lift this option up into the type.

package.json
```diff
@@ -9,8 +9,8 @@
   "engines": {
     "node": ">= 10"
   },
-  "main": "js/index.js",
-  "types": "js/index.d.ts",
+  "main": "js-dist/index.js",
```
andreiborza: m: I'm worried this breaks whoever directly tries to run the file from the previous path.

JPeer264: It is theoretically not a breaking change, and users would go in with `node_modules/bin/sentry-cli`. I'll change it so that the export stays the same with `js` to be sure

js/releases/index.js
JPeer264: Great catch. It seems this got wrongly updated!

lib/releases/index.js
JPeer264: Will do then in a follow up PR. Good idea

.github/workflows/test_node.yml
JPeer264: I would rather suggest to add that in the follow up PR on how to contribute to the JS part. That would explain this line and the others as well. Follow up issue: https://github.com/getsentry/sentry-cli/issues/2921

js/releases/index.js
JPeer264: No, options is and was an object before (now it is an "object" but packaged as an "interface" which declares it in one go and more precise). So in your example `foo(Options{ a, b, c, d })` is true for the old and the new notation.

lib/types.ts
JPeer264: Will do in a follow up PR: https://github.com/getsentry/sentry-cli/issues/2920

scripts/install.js
```diff
@@ -16,9 +16,9 @@ const ProgressBar = require('progress');
 const Proxy = require('proxy-from-env');
 const which = require('which');
 
-const helper = require('../js/helper');
+const helper = require('../js-dist/helper');
 const pkgInfo = require('../package.json');
-const Logger = require('../js/logger');
+const Logger = require('../js-dist/logger');
```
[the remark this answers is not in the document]
JPeer264: This is known and will be clarified in a docs update on how to contribute to the CLI for JS

szokeasaurusrex (commented):
I'll wait for @andreiborza to approve this, then I can provide a stamp (I think my approval is needed for some of the changes based on the code owner config)

andreiborza (approved):

andreiborza:
@szokeasaurusrex approved :)

PR #2974, opened 2025-11-20
feat(build): Add --force-git-metadata flag with CI auto-detection (EME-604)

runningcode:
## Summary

- Adds `--force-git-metadata` and `--no-git-metadata` flags to control git metadata collection for build uploads
- Implements CI detection that automatically enables git metadata only when running in most CI environments (GitHub Actions, GitLab CI, Jenkins, etc.)
- Prevents local dev builds from triggering unwanted GitHub status checks while maintaining automatic behavior for CI builds
- Refactors git metadata handling to use `VcsInfo` directly, eliminating duplicate `GitMetadata` struct

## Changes

**New CI Detection Module** (`src/utils/ci.rs`):
- Checks common CI environment variables to determine if running in CI
- Based on logic from sentry-android-gradle-plugin

**Build Upload Command** (`src/commands/build/upload.rs`):
- Added two separate boolean flags for clarity:
  - `--force-git-metadata` - Force enable git metadata collection
  - `--no-git-metadata` - Force disable git metadata collection
  - Default (no flag) - Auto-detect CI environment
- Refactored to use `VcsInfo` struct directly instead of intermediate `GitMetadata` struct
- Simplified string handling in `collect_git_metadata` function

Fixes EME-604

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -131,17 +131,9 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .get_many::<String>("paths")
         .expect("paths argument is required");
 
-    // Determine if we should collect git metadata
-    let should_collect_git_metadata = if matches.get_flag("force_git_metadata") {
-        // --force-git-metadata was specified
-        true
-    } else if matches.get_flag("no_git_metadata") {
-        // --no-git-metadata was specified
-        false
-    } else {
-        // Default behavior: auto-detect CI
-        is_ci()
-    };
+    // Collect git metadata if running in CI, unless explicitly enabled or disabled.
+    let should_collect_git_metadata =
+        matches.get_flag("force_git_metadata") || (!matches.get_flag("no_git_metadata") && is_ci());
 
     debug!(
         "Git metadata collection: {}",


src/commands/build/upload.rs
```diff
@@ -106,6 +107,23 @@ pub fn make_command(command: Command) -> Command {
...
+                    Use --git-metadata to force enable, --git-metadata=false or --no-git-metadata to force disable. \
+                    If not specified, git metadata is automatically collected only when running in a CI environment.")
+        )
+        .arg(
+            Arg::new("no_git_metadata")
+                .long("no-git-metadata")
+                .action(ArgAction::SetTrue)
+                .conflicts_with("git_metadata")
+                .help("Disable collection of git metadata. Equivalent to --git-metadata=false.")
+        )
```
szokeasaurusrex, line 125: To keep things simple, let's provide only a single way to disable Git metadata. I would be OK with either removing the `--no-git-metadata` flag, per the suggestion below, or changing the `--git-metadata` flag to only be a simple flag which does not take a true/false value.

```suggestion
```

src/utils/ci.rs
```diff
@@ -0,0 +1,79 @@
...
+            "BUILDKITE",
+        ];
+
+        for var in &ci_vars {
+            env::remove_var(var);
+        }
+
+        assert!(!is_ci());
+    }
+}
```
szokeasaurusrex: I don't think these tests are necessary, as the logic in this method is trivial.

If we were to keep the tests, we would need to serialize them to avoid concurrency flakes. As this complicates things further, I would prefer just not having tests here.

```suggestion
```

src/commands/build/upload.rs
```diff
@@ -114,58 +132,92 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    let (
+        head_sha,
+        vcs_provider,
+        head_repo_name,
+        head_ref,
+        base_ref,
+        base_repo_name,
+        base_sha,
+        pr_number,
+    ) = if should_collect_git_metadata {
```
szokeasaurusrex: This `if` block is super long, making it difficult to tell what the logic is doing at a glance.

I would suggest that we extract everything inside the `if` block into a separate method, so that the `if`/`else` fits in a single page. I would also suggest defining a `struct` to store the git metadata; that way, the method can simply return that struct. We can implement `Default` on that struct to construct a blank version in the `else` case.

Ultimately, the `if`/`else` would kinda look like so:

```rust
let git_metadata = if should_collect_git_metadata {
    // all logic currently in the `if` block is moved to the method below
    autodetect_git_metadata(matches, todo!("maybe some other arguments here"))
} else {
    Default::default()
};
```

runningcode: Good idea! I've implemented that refactoring

src/commands/build/upload.rs
```diff
@@ -32,99 +33,43 @@ use crate::utils::vcs::{
...
-                .long("build-configuration")
-                .help("The build configuration to use for the upload. If not provided, the current version will be used.")
-        )
-        .arg(
-            Arg::new("release_notes")
-                .long("release-notes")
-                .help("The release notes to use for the upload.")
-        )
+/// Holds git metadata collected for build uploads.
+#[derive(Debug, Default)]
```
runningcode: this is pretty similar to the VcsInfo object but the difference is that GitMetadata uses owned String types while VcsInfo uses borrowed &str. 

So we could potentially remove this but we need some other mechanism to keep the values alive. I think another option is that we could also have the `collect_git_metadata` return a Tuple.

What are your thoughts here?

src/commands/build/upload.rs
```diff
@@ -106,25 +107,55 @@ pub fn make_command(command: Command) -> Command {
...
+    base_repo_name: String,
+    base_sha: Option<Digest>,
+    pr_number: Option<u32>,
+}
 
+/// Collects git metadata from arguments and VCS introspection.
+///
+/// When `auto_collect` is false, only explicitly provided values are collected;
+/// automatic inference from git repository and CI environment is skipped.
+fn collect_git_metadata(matches: &ArgMatches, config: &Config, auto_collect: bool) -> GitMetadata {
```
runningcode: This is the same as the previous implementation with the addition of `auto_collect`. When true, the behavior is the same as before but when set to false (when git-metadata=false) we can still use the command-line provided values (like `--head-ref foo` when `--git-metadata=false`)

src/commands/build/upload.rs
runningcode: Good point, removed!

src/utils/ci.rs
runningcode: Yeah, you're right. I removed it.

src/commands/build/upload.rs
```diff
@@ -106,6 +107,29 @@ pub fn make_command(command: Command) -> Command {
...
+        .arg(
+            Arg::new("git_metadata")
+                .long("git-metadata")
+                .num_args(0..=1)
+                .default_missing_value("true")
+                .value_parser(clap::value_parser!(bool))
+                .help("Controls whether to collect and send git metadata (branch, commit, etc.). \
+                    Use --git-metadata to force enable, --git-metadata=false to force disable. \
+                    If not specified, git metadata is automatically collected only when running in a CI environment.")
+        )
```
runningcode, line 125: I'm confused what `clap` is but I think that is the behavior that we want.

tests/integration/_cases/build/build-upload-help-not-macos.trycmd
```diff
@@ -52,6 +52,11 @@ Options:
           be used.
       --release-notes <release_notes>
           The release notes to use for the upload.
+      --git-metadata [<git_metadata>]
```
chromy: `[<git_metadata>]` makes me expect a sha or something to go here.
Maybe:
- `[true|false]` 
- `[<force_git_metadata>]`

or something could be better.

src/utils/ci.rs
```diff
@@ -0,0 +1,33 @@
...
+/// - TeamCity
+/// - Bamboo
+/// - Bitrise
+/// - GoCD
+/// - Azure Pipelines
+/// - Buildkite
+///
+/// Based on: https://github.com/getsentry/sentry-android-gradle-plugin/blob/15068f51fee344c00efdbec5a172297be4719b86/plugin-build/src/main/kotlin/io/sentry/android/gradle/util/CI.kt#L9
+pub fn is_ci() -> bool {
+    env::var("CI").is_ok()
```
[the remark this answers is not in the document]
chromy: I don't know if this is a common thing or not. Even if it is might be better to leave this consistant with the Gradle plugin above rather than diverging the logic.

src/commands/build/upload.rs
chromy: Code is getting to be kind of deeply nested :/

One alternative approach might be the equivalent of: 

```

def get_git_metadata_from_repo():
   return {
     "head_sha": git_repo_head_sha()
     # ...
   }
  

def get_git_metadata_from_cli():
   return {
     "head_sha": cli.head_sha or None,
     # ...
   }

def get_git_metadata():
  if should_collect_git_metadata:
    default_metadata = get_git_metadata_from_repo()
  else:
    default_metadata = {}

  manual_metadata = get_git_metadata_from_cli()

  return {
    "head_sha": manual_metadata["head_sha"] or default_metadata["head_sha"],
    # ...
  }
```

(Could even have: `get_git_metadata_from_env` for the github env inspection vs. the repo case)

This would split the priority hierarchy (`cli >> env (if infer enabled) >> repo (if infer enabled) >> empty`)  - from the mechanics of extracting from cli vs. env vs. repo.

Alternatively maybe there is some nice rust feature for making this a bit less nested.

runningcode: Yeah this is a good point. @szokeasaurusrex, is there some fancy Rust feature to help with the nesting here?

src/commands/build/upload.rs
szokeasaurusrex: `clap` is the library which parses the command line arguments

src/commands/build/upload.rs
szokeasaurusrex: If we want to reuse the `VcsInfo` struct here, I would just change the `&str` fields in `VcsInfo` to `Cow<str>`; that way, both borrowed and owned values can be supported. I have addressed the `&u32` in #3000.

src/commands/build/upload.rs
szokeasaurusrex: The only thing I can think of is `.then` (will include an inline comment with an example). But, that is only a minor simplification imo.

I kinda like @chromy's suggestion of splitting the logic for collecting from the CLI from the logic for auto detection, but I'm not sure how easy/difficult it would be to implement.

src/commands/build/upload.rs
```diff
@@ -400,6 +282,228 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+fn collect_git_metadata(matches: &ArgMatches, config: &Config, auto_collect: bool) -> GitMetadata {
+    let head_sha = matches
+        .get_one::<Option<Digest>>("head_sha")
+        .map(|d| d.as_ref().cloned())
+        .or_else(|| {
+            if auto_collect {
+                Some(vcs::find_head_sha().ok())
+            } else {
+                None
+            }
```
szokeasaurusrex: [`bool::then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then) is your friend here


```suggestion
            auto_collect.then(|| vcs::find_head_sha().ok())
```

src/commands/build/upload.rs
```diff
@@ -757,4 +861,89 @@ mod tests {
         assert!(has_symlink, "Symlink should be preserved in zip");
         Ok(())
     }
+
+    #[test]
+    fn test_collect_git_metadata_respects_explicit_values_when_auto_collect_disabled() {
+        use ini::Ini;
+        use std::path::PathBuf;
```
szokeasaurusrex: It's weird to have `use` statements inside a function body. These should instead go at the top of the `mod tests` block.

src/utils/ci.rs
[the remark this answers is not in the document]
szokeasaurusrex: I am also unsure; this may be worth investigating

src/commands/build/upload.rs
```diff
@@ -106,6 +107,29 @@ pub fn make_command(command: Command) -> Command {
...
         )
+        .arg(
+            Arg::new("force_git_metadata")
+                .long("force-git-metadata")
+                .num_args(0..=1)
+                .default_missing_value("true")
+                .value_parser(clap::value_parser!(bool))
+                .help("Controls whether to collect and send git metadata (branch, commit, etc.). \
+                    Use --force-git-metadata to force enable, --force-git-metadata=false to force disable. \
+                    If not specified, git metadata is automatically collected only when running in a CI environment.")
```
szokeasaurusrex: [nit] Maybe we should specify which CI environments we can autodetect? Or, just say that we can detect "most common CI environments"? To be honest, I am not sure we need to do this, though, what do you think?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Features
+
+- Add `--force-git-metadata` flag with CI auto-detection to `sentry-cli build upload` command ([#2974](https://github.com/getsentry/sentry-cli/pull/2974))
+  - Git metadata is now automatically collected only when running in CI environments (GitHub Actions, GitLab CI, Jenkins, etc.)
+  - Local development builds no longer trigger GitHub status checks by default
+  - Users can force enable with `--force-git-metadata` or disable with `--force-git-metadata=false`
```
szokeasaurusrex:
```suggestion
  - Users can force-enable with `--force-git-metadata` or force-disable with `--force-git-metadata=false`
```

src/commands/build/upload.rs
szokeasaurusrex: h: `--force-git-metadata` does not work; a value is required. 

I feel that the requirement to supply a value makes the API clunkier. Upon further reflection, I think we should just have a `--force-git-metadata` flag and a `--no-git-metadata` flag, neither of which takes a value. 

What do you think?

CHANGELOG.md
szokeasaurusrex: I would list this under "improvements" rather than features. While it is true that we have added the `--force-git-metadata` flag, I believe most users will rely only on the new default CI auto-detection behavior, so we should focus the changelog entry on the auto-detection as an improvement.

With the focus on the improvement in mind, I would rewrite the entry like so:

```suggestion
- For the `sentry-cli build upload` command, we now only auto-detect Git metadata when we detect we are running in a CI environment, unless the user manually overrides this behavior ([#2974](https://github.com/getsentry/sentry-cli/pull/2974)). This change prevents local development builds from triggiering GitHub status checks for size analysis.
  - We can detect most common CI environments based on the environment variables these set.
  - **TODO:** Once we have settled on the arguments users can use to override the behavior, we can mention those here.
```

src/commands/build/upload.rs
runningcode: That's a good point but I think it is too long of a list for this help text to list all of them. I'll just say "most"

runningcode: oh. nice catch! `--force-git-metadata` flag and a `--no-git-metadata`

tests/integration/_cases/build/build-upload-help-not-macos.trycmd
runningcode: adjusted this to add `--force-git_metadata` and `--no-git-metadata`.

src/utils/ci.rs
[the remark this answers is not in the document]
runningcode: I don't think it is common. The Gradle code I copied was used across many, many orgs. They an just `unset` this env variable to test.

src/commands/build/upload.rs
runningcode: So I took a quick stab at splitting this up in to `get_git_metadata_from_repo` and `get_git_metadata_from_cli` doing this and it seems to require creating a lot of structs to hold all the different return objects for all these functions. I think it's a good idea but maybe this refactoring would make sense in a separate PR to make it easier to review?

src/commands/build/upload.rs
```diff
@@ -114,169 +131,29 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    let should_collect_git_metadata = if matches.get_flag("force_git_metadata") {
+        // --force-git-metadata was specified
+        true
+    } else if matches.get_flag("no_git_metadata") {
+        // --no-git-metadata was specified
+        false
+    } else {
+        // Default behavior: auto-detect CI
+        is_ci()
     };
```
szokeasaurusrex: l: I prefer writing the boolean expression directly, but as this is more a matter of personal preference, I am also OK with what you wrote

```suggestion
    // Collect git metadata if running in CI, unless explicitly enabled or disabled.
    let should_collect_git_metadata =
        matches.get_flag("force_git_metadata") || (!matches.get_flag("no_git_metadata") && is_ci());
```

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Improvements
 
+- Add `--force-git-metadata` and `--no-git-metadata` flags with CI auto-detection to `sentry-cli build upload` command ([#2974](https://github.com/getsentry/sentry-cli/pull/2974))
+  - Git metadata is now automatically collected only when running in most CI environments (GitHub Actions, GitLab CI, Jenkins, etc.)
+  - Local development builds no longer trigger GitHub status checks by default
+  - Users can force enable with `--force-git-metadata` or disable with `--no-git-metadata`
```
szokeasaurusrex: Thanks for moving this to "improvements." 

I would still suggest rephrasing the entry to emphasize the new default behavior [per my previous comment](https://github.com/getsentry/sentry-cli/pull/2974#discussion_r2581056784). Or, is there some reason you think this rephrase is a bad idea?

```suggestion
- For the `sentry-cli build upload` command, we now only auto-detect Git metadata when we detect we are running in a CI environment, unless the user manually overrides this behavior ([#2974](https://github.com/getsentry/sentry-cli/pull/2974)). This change prevents local development builds from triggiering GitHub status checks for size analysis.
  - We can detect most common CI environments based on the environment variables these set.
  - We introduced two new arguments, `--force-git-metadata` and `--no-git-metadata`, which force-enable and force-disable automatic Git data collection, respectively, overriding the default behavior.
```

runningcode: Oh good point, I think it is a great idea and I missed that comment.

PR #2976, opened 2025-11-21
feat(build): Set head-ref in non-PR GitHub workflows (EME-654)

runningcode:
## Summary
- Extends `get_github_head_ref()` to support non-PR GitHub Actions workflows
- For PR workflows: continues to use `GITHUB_HEAD_REF` (unchanged)
- For non-PR workflows (push, release, etc.): now uses `GITHUB_REF_NAME`
- Adds comprehensive tests for all workflow types

## Changes
Modified `src/utils/vcs.rs`:
- Updated `get_github_head_ref()` to handle non-PR events by reading `GITHUB_REF_NAME`
- Added test coverage for push workflows, feature branches, tags, and edge cases

## Impact
Build uploads in non-PR workflows (like Android release builds) will now automatically have the correct branch or tag reference set, without requiring manual `--head-ref` flags.

Fixes EME-654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+- The `sentry-cli build upload` command now automatically detects the correct branch or tag reference in non-PR GitHub Actions workflows ([#2976](https://github.com/getsentry/sentry-cli/pull/2976)). Previously, `--head-ref` was only auto-detected for pull request workflows. Now it works for push, release, and other workflow types by using the `GITHUB_REF_NAME` environment variable.
+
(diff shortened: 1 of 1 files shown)


src/utils/vcs.rs
```diff
@@ -1585,6 +1590,51 @@ mod tests {
         std::env::remove_var("GITHUB_EVENT_NAME");
     }
 
+    #[test]
+    #[serial(github_env)]
```
runningcode, line 1594: This is set to serial because we're setting github env vars (like other tests that also do this). We can also split up these tests but I figured it saves all the cleanup and teardown before and after every test if we do it all together.

PR #2994, opened 2025-11-27
feat(build): Track plugin versions from SENTRY_PIPELINE in build uploads (EME-606)

runningcode:
## Summary

Automatically track Sentry plugin versions in build uploads by parsing the existing `SENTRY_PIPELINE` environment variable. This enables size analysis and build distribution tracking by storing plugin version metadata in the PreprodArtifact database table.

### Plugin Version Detection
- Reads the existing `SENTRY_PIPELINE` environment variable
- Parses format: `sentry-gradle-plugin/4.12.0` or `sentry-fastlane-plugin/1.2.3`
- Only extracts versions for recognized Sentry plugins:
  - `sentry-gradle-plugin`
  - `sentry-fastlane-plugin`

### Metadata File Format
The `.sentry-cli-metadata.txt` file inside uploaded zips now includes detected plugin info:
```
sentry-cli-version: 2.58.2
sentry-gradle-plugin: 4.12.0
```

Related PRs:
https://github.com/getsentry/launchpad/pull/464
https://github.com/getsentry/sentry/pull/103062
https://github.com/getsentry/sentry-android-gradle-plugin/pull/1036
https://github.com/getsentry/sentry-fastlane-plugin/pull/365

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -125,6 +125,33 @@ pub fn make_command(command: Command) -> Command {
         )
 }
 
+/// Parse plugin info from SENTRY_PIPELINE environment variable.
+/// Format: "sentry-gradle-plugin/4.12.0" or "sentry-fastlane-plugin/1.2.3"
+/// Returns (plugin_name, plugin_version) if a recognized plugin is found, (None, None) otherwise.
+fn parse_plugin_from_pipeline(pipeline: Option<String>) -> (Option<String>, Option<String>) {
+    pipeline
+        .and_then(|pipeline| {
+            let parts: Vec<&str> = pipeline.splitn(2, '/').collect();
+            if parts.len() == 2 {
+                let name = parts[0];
+                let version = parts[1];
+
+                // Only extract known Sentry plugins
+                if name == "sentry-gradle-plugin" || name == "sentry-fastlane-plugin" {
+                    debug!("Detected {name} version {version} from SENTRY_PIPELINE");
+                    Some((name.to_owned(), version.to_owned()))
+                } else {
+                    debug!("SENTRY_PIPELINE contains unrecognized plugin: {name}");
+                    None
+                }
+            } else {
+                debug!("SENTRY_PIPELINE format not recognized: {pipeline}");
+                None
+            }
+        })
+        .unzip()
+}
+
 pub fn execute(matches: &ArgMatches) -> Result<()> {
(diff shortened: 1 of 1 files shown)


src/commands/build/upload.rs
```diff
@@ -757,4 +805,35 @@ mod tests {
...
+
+        // Find and read the metadata file
+        let metadata_file = archive.by_name(".sentry-cli-metadata.txt")?;
+        let metadata_content = std::io::read_to_string(metadata_file)?;
+
+        assert!(
+            metadata_content.contains("sentry-cli-version:"),
+            "Metadata should contain sentry-cli-version"
+        );
+        assert!(
```
runningcode, line 955: technically it is impossible to have both. i could also add an assertion that both are not set but I didn't think it was worth it.

src/commands/build/upload.rs
```diff
@@ -150,6 +150,31 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+            if parts.len() == 2 {
+                let name = parts[0];
+                let version = parts[1];
+
+                // Only extract known Sentry plugins
+                if name == "sentry-gradle-plugin" || name == "sentry-fastlane-plugin" {
+                    debug!("Detected {name} version {version} from SENTRY_PIPELINE");
+                    Some((name.to_owned(), version.to_owned()))
+                } else {
+                    debug!("SENTRY_PIPELINE contains unrecognized plugin: {name}");
```
runningcode: This will be printed out as debug for javascript when they use the `SENTRY_PIPELINE` variable. I think this message is helpful for debugging but might be confusing if we print it for javascript devs.

Alternatively, we can also detect the javascript tool and not print this in that case. WDYT?

szokeasaurusrex: I don't think the JS build plugins will call the `build upload` command. So likely, they will not hit this.

Also, debug logs are hidden by default, so it it anyways likely will not be shown.

However, I am not sure we really need to log this case, since it might not add that much value. I'll leave it up to you

szokeasaurusrex (approved):
Looks good overall, perhaps you could consider adding an integration test which checks that we can correctly read the environment variable

runningcode:
Thanks! Added tests for correctly parsing the `SENTRY_PIPELINE` variable.

I will keep the debug logging since it could detect a misbehaving gradle/fastlane plugin or another plugin in the future.

PR #2996, opened 2025-12-01
build: Bump MSRV to 1.91

lcian:
### Description

Bumps the MSRV to 1.91 and also updates the Dockerfile to use a 1.91.1 image.
This change is in preparation for the integration of https://crates.io/crates/objectstore-client, which has an MSRV of 1.89. 
We might as well bump it to 1.91 while we're at it.


src/commands/dart_symbol_map/upload.rs
```diff
@@ -90,7 +90,7 @@ pub(super) fn execute(args: DartSymbolMapUploadArgs) -> Result<()> {
                 serde_json::from_slice(mapping_file_bytes.as_ref())
                     .context("Invalid dartsymbolmap: expected a JSON array of strings")?;
 
-            if mapping_entries.len() % 2 != 0 {
+            if !mapping_entries.len().is_multiple_of(2) {
```
lcian, line 93: 😭

PR #3001, opened 2025-12-02
feat: Support missing calls with multi-project options

### LINKED ISSUE #2902: Align JS API between commands that accept multiple projects
### Problem Statement

Today, only releases.new and releases.uploadSourcemaps support supplying multiple projects via the `getProjectFlagsFromOptions` helper:

https://github.com/getsentry/sentry-cli/blob/8c6e63e6da73194fae6d2257daf643b3200389b5/js/releases/index.js#L53-L56

https://github.com/getsentry/sentry-cli/blob/8c6e63e6da73194fae6d2257daf643b3200389b5/js/releases/index.js#L197-L199

Briefly looking over the rust code-base, it seems that more commands support passing multiple projects.

### Solution Brainstorm

Align the JS API with the Rust API to allow passing multiple projects to all commands that have a JS counterpart and support passing multiple projects.

We should be able to use the same helper and approach as in the above mentioned apis.

JPeer264:
This adds more support for having multiple projects where the Rust API had already support. This only affects `.newDeploy()`

Closes #2902
Closes [CLI-212](https://linear.app/getsentry/issue/CLI-212/align-js-api-between-commands-that-accept-multiple-projects)

### DIFF
--- a/lib/releases/__tests__/index.test.js
+++ b/lib/releases/__tests__/index.test.js
@@ -150,7 +150,7 @@ describe('SentryCli releases', () => {
         await cli.releases.newDeploy('my-version', { env: 'production' });
 
         expect(mockExecute).toHaveBeenCalledWith(
-          ['releases', 'deploys', 'my-version', 'new', '--env', 'production'],
+          ['deploys', 'new', '--release', 'my-version', '--env', 'production'],
           null,
           false,
           undefined,
@@ -166,14 +166,14 @@ describe('SentryCli releases', () => {
 
         expect(mockExecute).toHaveBeenCalledWith(
           [
-            'releases',
             'deploys',
+            'new',
             '-p',
             'proj-a',
             '-p',
             'proj-b',
+            '--release',
             'my-version',
-            'new',
             '--env',
             'production',
           ],
--- a/lib/releases/index.ts
+++ b/lib/releases/index.ts
@@ -214,9 +214,9 @@ export class Releases {
     if (!options || !options.env) {
       throw new Error('options.env must be a valid name');
     }
-    const args = ['releases', 'deploys']
+    const args = ['deploys', 'new']
       .concat(helper.getProjectFlagsFromOptions(options))
-      .concat([release, 'new']);
+      .concat(['--release', release]);
     return this.execute(helper.prepareCommand(args, DEPLOYS_OPTIONS, options), null);
   }
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -168,7 +168,7 @@ export type SentryCliNewDeployOptions = {
    */
   url?: string;
   /**
-   * The projects to deploy the release to. If not provided, the deployment will be created for the default project.
+   * The projects to deploy the release to. If not provided, the deployment will be created for all projects associated with the release.
    */
   projects?: string[];
 }


lib/types.ts
```diff
@@ -175,6 +185,10 @@ export type SentryCliNewDeployOptions = {
    * URL that points to the deployment.
    */
   url?: string;
+  /**
+   * The projects to deploy the release to. If not provided, the deployment will be created for the default project.
```
szokeasaurusrex:
```suggestion
   * The projects to deploy the release to. If not provided, the deployment will be created for all projects associated with the release.
```

szokeasaurusrex:
Also, does this change allow `releases new` to accept multiple projects?

JPeer264:
> Also, does this change allow releases new to accept multiple projects?

yes

lib/types.ts
szokeasaurusrex: You missed this comment @JPeer264 😄

lib/releases/__tests__/index.test.js
```diff
@@ -144,5 +144,45 @@ describe('SentryCli releases', () => {
         );
       });
     });
+
+    describe('newDeploy', () => {
+      test('without projects', async () => {
+        await cli.releases.newDeploy('my-version', { env: 'production' });
+
+        expect(mockExecute).toHaveBeenCalledWith(
+          ['releases', 'deploys', 'my-version', 'new', '--env', 'production'],
```
szokeasaurusrex: m: `releases deploys` is a soft-deprecated alias to `deploys`. This would be the preferred way to invoke the command:

```suggestion
          ['deploys', 'new', '--release', 'my-version', '--env', 'production'],
```

lib/releases/__tests__/index.test.js
```diff
@@ -144,5 +144,45 @@ describe('SentryCli releases', () => {
...
+            'releases',
+            'deploys',
+            '-p',
+            'proj-a',
+            '-p',
+            'proj-b',
+            'my-version',
+            'new',
+            '--env',
+            'production',
```
szokeasaurusrex: m: we should use `deploys new` (note that the order of the parameters does not matter from Sentry CLI's perspective, although I believe they should all come after `deploys new`).

```suggestion
            'deploys',
            'new',
            '-p',
            'proj-a',
            '-p',
            'proj-b',
            '--release',
            'my-version',
            '--env',
            'production',
```

lib/releases/index.ts
```diff
@@ -213,7 +214,9 @@ export class Releases {
     if (!options || !options.env) {
       throw new Error('options.env must be a valid name');
     }
-    const args = ['releases', 'deploys', release, 'new'];
+    const args = ['releases', 'deploys']
+      .concat(helper.getProjectFlagsFromOptions(options))
+      .concat([release, 'new']);
```
szokeasaurusrex: m: I am pretty sure this will get the result we want with using `deploys new`

```suggestion
    const args = ['deploys', 'new']
      .concat(helper.getProjectFlagsFromOptions(options))
      .concat(['--release', release]);
```

lib/releases/__tests__/index.test.js
JPeer264: You're totally right, I cross checked with the cli output

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
```
szokeasaurusrex: As this is a new API, we should mark it as a new feature and release in a minor

```suggestion
### New Features
```

PR #3003, opened 2025-12-02
fix(sourcemaps): Don't treat `.map` files as JavaScript when injecting

### LINKED ISSUE #3017: JS code can get injected into sourcemap
Tracking issue for getsentry/sentry-cli#3003

fhaehnel-figma:
### Description

The inject command was incorrectly passing "map" as a JavaScript extension to `inject_debug_ids()`, causing `.map` files to be processed for JavaScript injection. This corrupted sourcemap files by injecting the debug ID runtime snippet into them.

Root cause: In `inject.rs`, the extensions array included "map" for file discovery, but this same array was passed to `inject_debug_ids()` which expected only JavaScript extensions for filtering injection candidates.

Fix: Separate `js_extensions `(for injection filtering) from `search_extensions` (for file discovery). Only pass `js_extensions `to inject_debug_ids() so `.map` files are filtered out during injection.

Added regression test that verifies .map files never get JavaScript code injected, even if they could be misclassified as MinifiedSource.

### Background

  **The Bug:** The `sourcemaps inject` command could corrupt `.map` files by injecting JavaScript code into them, causing subsequent validation failures.
This only happens _sometimes_, and can usually be worked around by re-running the command.

  **Root Cause:**
  - In `src/commands/sourcemaps/inject.rs`, the code built an `extensions` array containing `["js", "cjs", "mjs", "map"]` for file discovery
  - This same array was passed to `inject_debug_ids()` as the `js_extensions` parameter
  - Inside `inject_debug_ids()` at `src/utils/sourcemaps.rs:758`, there's an extension filter:
    ```rust
    if !url_matches_extension(source_url, js_extensions) {
        continue;
    }
  - This filter was intended to only allow JavaScript files through, but because "map" was in the extensions list, .map files could pass this check
  - If a .map file somehow got into the `sourcemap_references` HashMap as a source (rather than as a sourcemap), it would be processed as a JavaScript file and have the debug ID runtime snippet injected into it

#### Why It's Non-Deterministic

The non-determinism comes from Rust's HashMap iteration order:

1. HashMap iteration is non-deterministic: The `sourcemap_references` field in SourceMapProcessor is a `HashMap<String, Option<SourceMapReference>>` (line 196 of sourcemaps.rs)
2. Processing order affects which files get corrupted: At line 756, the code iterates: `for (source_url, sourcemap_url) in self.sourcemap_references.iter_mut()`
3. Race condition scenario:
- If a .map file gets processed as a source file first (due to HashMap iteration order), and passes the extension check, JavaScript gets injected
- The corrupted .map file then fails validation when it is later processed as an actual sourcemap (because of a reference from a _real_ JS source file)
- If the iteration order is different, the .map might be processed correctly as a sourcemap (by reference) first, avoiding the bug

### Issues

- Resolves #3017 
- Resolves [CLI-245](https://linear.app/getsentry/issue/CLI-245/js-code-can-get-injected-into-sourcemap)


fhaehnel-figma:
Hey @szokeasaurusrex! We'd really appreciate eyes on this, it's causing us quite a lot of pain when trying to route our issues to the right teams at the moment! 🙏🏻

szokeasaurusrex:
Hi @fhaehnel-figma, this is on my radar, I'll get back to you within the next few days

szokeasaurusrex:
@loewenheim, would appreciate a second review from you here, since I also made some changes to this

PR #3024, opened 2025-12-10
Save PDFs and SVGs with content hash

NicoHinderling:
Related to https://linear.app/getsentry/issue/EME-550/ios-insights-duplicate-files

Write PDF/SVG raw data to disk during asset catalog parsing to enable duplicate detection of SVG and pdf files

### DIFF
--- a/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
+++ b/apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
@@ -228,7 +228,8 @@ enum AssetUtil {
                 
                 // Compute content hash for PDFs/SVGs without saving to disk
                 if fileExtension == "pdf" || fileExtension == "svg" {
-                    contentHash = data.sha256Hash()
+                  // Hash PDFs/SVGs in-memory (Python can't access _srcData without parsing binary .car format)
+                  contentHash = data.sha256Hash()
                 }
                 // Save images that can be converted to CGImage (excluding PDFs/SVGs)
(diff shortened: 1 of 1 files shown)


apple-catalog-parsing/native/swift/AssetCatalogParser/Sources/AssetCatalogParser/AssetCatalogReader.swift
```diff
@@ -223,8 +223,15 @@ enum AssetUtil {
...
-                // Skip SVGs, but save images even if they don't have an extension (default to png)
-                if fileExtension != "svg", let unslicedImage = unslicedImage {
+                // Handle PDFs and SVGs by saving raw data
+                if fileExtension == "pdf" || fileExtension == "svg" {
+                    let fileURL = try! createResultsPath(assetURL: file, outputURL: outputURL)
+                        .appendingPathComponent(imageId)
+                        .appendingPathExtension(fileExtension)
+                    try! data.write(to: fileURL)
+                }
+                // Save images that can be converted to CGImage
```
trevor-e: Assuming that PDF/SVG fail the later `CGImageDestinationCreateWithURL()` calls? Just wondering if it's worth trying to fit this into `images[imageId]` at all, just a bit confusing that dictionary no longer contains all images.

NicoHinderling: right, for SVGs CGImageDestinationCreateWithURL would return `nil`. For PDFs, the call would have succeeded, but the vector data would have been removed... 

that said I realized this change can increase the size of ParsedAssets by a lot, so I've come up with a different lighterweight approach that achieves the same fix

PR #3038, opened 2025-12-15
perf(preprod): Use zstd for zip compression

NicoHinderling:
Noticed uploading a certain customer's build took 5+ minutes and upon further inspection the zipping was the lions share of that time. If we take away the actual network transport of the upload:

- With level 6 compression: 5 minutes -> ultimately 566MB
- with level 1 compression: 2.5 min -> ultimately 658MB
- with zstandard: 36s -> ultimately 533MB

What's more, we can support both types of compression on the Sentry backend, so this seems like a free (and significant) win

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 3.0.1
+## Unreleased
 
 ### Performance Improvements
 


tests/integration/_cases/build/build-upload-apk.trycmd
```diff
@@ -2,8 +2,7 @@
 $ sentry-cli build upload tests/integration/_fixtures/build/apk.apk --head-sha 12345678deadbeef78900987feebdaed87654321
 ? success
 > Preparing for upload completed in [..]
-> Uploading completed in [..]
 Successfully uploaded 1 file to Sentry
-  - tests/integration/_fixtures/build/apk.apk (http[..]/wat-org/preprod/wat-project/42)
```
NicoHinderling, line 7: I think the compression might be more deterministic or something.. so we don't need to upload a chunk since it's already on the server i guess? not 100% clear why this was required to be changed (but the change is definitely working... lol)

don't want to sink more than the 20 min I already did on this one test case when everything else is working

Cargo.toml
```diff
@@ -70,7 +70,7 @@ uuid = { version = "1.3.0", features = ["v4", "serde"] }
 walkdir = "2.3.2"
 which = "4.4.0"
 whoami = "1.5.2"
-zip = "2.4.2"
+zip = { version = "2.4.2", features = ["zstd"] }
```
szokeasaurusrex: There should probably be a corresponding change to `Cargo.lock` in this PR.

tests/integration/_cases/build/build-upload-apk.trycmd
szokeasaurusrex: Changing the compression algorithm changes the hash of the chunks. It looks like you changed the hash we return, but maybe that hash is incorrect? This seems strange to me

Cargo.toml
szokeasaurusrex: Ah — `zstd` is a default feature, so we actually do not need this change at all (this explains why there is no change to `Cargo.lock`)

```suggestion
zip = "2.4.2"
```

tests/integration/_cases/build/build-upload-apk.trycmd
szokeasaurusrex: h: It looks like the hashes that the mock server is returning are incorrect

tests/integration/build/upload.rs
```diff
@@ -167,7 +167,7 @@ fn command_build_upload_apk_chunked() {
                 if is_first_assemble_call.swap(false, Ordering::Relaxed) {
                     r#"{
                         "state": "created",
-                        "missingChunks": ["60863d91bb673a1b1b92dbbe91b1de5cc0dde146"]
+                        "missingChunks": ["1d1c9567a6453d0d8029b699bb588ebb946af50d"]
```
szokeasaurusrex: h: The hash is wrong; this is why the test output has changed.

```suggestion
                        "missingChunks": ["7138c09b474a5c84ac60e1b145855bf6dcc88913"]
```

tests/integration/_cases/build/build-upload-ipa.trycmd
```diff
@@ -2,7 +2,8 @@
 $ sentry-cli build upload tests/integration/_fixtures/build/ipa.ipa --head-sha deadbeef12345678deadbeef12345678deadbeef
 ? success
 > Preparing for upload completed in [..]
+> Uploading completed in [..]
```
szokeasaurusrex, line 5: strange, this indicates we were returning the wrong hash earlier for IPA files 😂

tests/integration/build/upload.rs
szokeasaurusrex, line 156: Once you update the `missingChunks` to the correct hash, this line will fail, because the expected chunk body needs to be updated following the change to `zstd`

tests/integration/_cases/build/build-upload-apk.trycmd
runningcode: Yeah I can confirm, when I was changing things sentry-cli the tests would fail with the missing `Uploading completed in [..]` and that means the hashes from the mock server are incorrect.

NicoHinderling:
ahhh @szokeasaurusrex the main issue was needing to also update the binary file associated with it. I think we're good now. 

and yes python only gets native zstd support in version 3.14 but we have a monkeypatch library that we are using within sentry for our specific use case and should work even self hosted folks

src/commands/build/upload.rs
```diff
@@ -563,7 +563,7 @@ fn normalize_file(
     // This is important as an optimization to avoid re-uploading the same chunks if they're already on the server
     // but the last modified time being different will cause checksums to be different.
     let options = SimpleFileOptions::default()
-        .compression_method(zip::CompressionMethod::Stored)
+        .compression_method(zip::CompressionMethod::Zstd)
```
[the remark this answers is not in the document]
NicoHinderling: `cargo build` still works without it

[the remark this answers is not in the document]
szokeasaurusrex: Yeah this comment is wrong

szokeasaurusrex:
Ah also please add a changelog entry before merge! 🙏 I also adjusted the PR title to "perf" to indicate this is a performance improvement, DangerJS should now prompt for a changelog entry 😃

NicoHinderling:
> except if build uploads are still marked as beta, then I think we can get away with it

yes they are. we actually still in EA and only launching into beta tomorrow so I think this will be a good update to get in right before beta

PR #3064, opened 2026-01-07
ref: Assume chunk uploading support

### LINKED ISSUE #3050: Assume chunk upload support
The PR for [CLI-176](https://linear.app/getsentry/issue/CLI-176/non-existent-org-causes-chunk-upload-not-supported-error) missed some types of chunk uploads (ProGuard, Dart, and Preprod). We don't need to read the chunk upload `accept` fields for these and can safely assume server support.

szokeasaurusrex:
### Description
Assume that the Sentry server supports all file types we would like to upload.

This change fixes compatibility with uploading ProGuard files self-hosted version 24.11.1, which [can receive chunk-uploaded ProGuard files](https://github.com/getsentry/sentry/commit/7c7e79fa2c4a41537687545a381aca1a091897dc), but [does not declare support in the `accept` field, yet](https://github.com/getsentry/sentry/commit/e0b7ff2fb72e49bfe862c6b90e42257990ea38e1). This will allow us to expand Sentry CLI 3.x's official support to all self-hosted versions above 24.11.1.

We also no longer check the `accept` field during `dart-symbol-map` and `build` uploads. As not all versions ≥24.11.1 support these features, this means that users on a self-hosted version that supports these features tries to use the command, they will no longer get an error message that states that their server lacks uploading support for those file types if they attempt to upload it. But, I think this is okay, as long as we are clear what versions are supported for these features.

### Issues
- Resolves #3050 
- Resolves [CLI-254](https://linear.app/getsentry/issue/CLI-254/assume-chunk-upload-support)


szokeasaurusrex:
Depends whether the file is able to complete assembly quickly or not. It's possible there is no error, but if you rerun with the same file after assembly finishes, you should get an error message "unsupported file type" error.

tests/integration/_cases/build/build-upload-help-macos.trycmd
szokeasaurusrex, line 1: Adding a `long_about` apparently causes clap to completely reformat the help text

PR #3068, opened 2026-01-09
fix(releases): Serialize project IDs as integers

### LINKED ISSUE #3066: Send project IDs as integers for release endpoints
Self hosted versions (besides the most recent ones) only can take project IDs at the release endpoints when they are serialized as JSON integers, not as strings, as they are affected by [a bug](https://github.com/getsentry/sentry/issues/105038) which was only fixed in getsentry/sentry#105159.

Although SaaS and new self-hosted versions will now accept project IDs when serialized as strings, we should serialize them as integers, instead, to ensure compatibility with older self-hosted versions.

szokeasaurusrex:
### Description
Serialize project IDs as integers when creating a new release. This restores compatibility with self-hosted Sentry versions which lack [this bug fix](https://linear.app/getsentry/review/fixapi-accept-project-ids-as-strings-in-organization-releases-endpoint-0c24c083a8ac) to properly handle project IDs serialized as strings.

### Issues
- Resolves #3066
- Resolves [CLI-261](https://linear.app/getsentry/issue/CLI-261/send-project-ids-as-integers-for-release-endpoints)
- See also https://github.com/getsentry/sentry/issues/105038

### DIFF
--- a/src/api/serialization.rs
+++ b/src/api/serialization.rs
@@ -1,0 +1,128 @@
+//! This module contains some custom serialization logic for the API.
+use std::sync::LazyLock;
+
+use regex::Regex;
+use serde::ser::SerializeSeq as _;
+use serde::{Serialize, Serializer};
+
+/// A container for either a numeric ID or an alphanumeric slug.
+///
+/// IDs are serialized as integers, while slugs are serialized as strings.
+#[derive(Serialize)]
+#[serde(untagged)]
+enum IdSlug<'s> {
+    Id(i64),
+    Slug(&'s str),
+}
+
+/// Serializes a sequence of strings, which may contain either numeric IDs or alphanumeric slugs.
+///
+/// We check each element in the sequence. If the element only contains digits and can be parsed as a 64-bit signed integer,
+/// we consider the value to be an ID. Otherwise, we consider the value to be a slug.
+///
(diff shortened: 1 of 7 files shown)


src/api/serialization.rs
```diff
@@ -0,0 +1,128 @@
...
+        ///
+        /// Project IDs always contain only digits.
+        static PROJECT_ID_REGEX: LazyLock<Regex> =
+            LazyLock::new(|| Regex::new(r"^\d+$").expect("regex is valid"));
+
+        let value = value.as_ref();
+
+        PROJECT_ID_REGEX
+            .is_match(value)
+            .then(|| value.parse().ok().map(IdSlug::Id))
```
lcian, line 58: nit: probably you can just attempt to `parse` it without applying the regex. If the parsing fails, you can fall back to considering it a slug.

szokeasaurusrex: I tried this in a previous iteration, but it fails for some edge cases. For example, negative numbers parse correctly (even though IDs cannot be negative, despite the fact that the database table apparently stores them as signed integers). Even checking that the values are greater than or equal to zero does not work, because "-0" (which can be a valid project slug) parses to 0.

szokeasaurusrex: I think the regex is the most reliable here

PR #3069, opened 2026-01-09
ci(release): Switch from action-prepare-release to Craft

BYK:
## Summary

This PR migrates from the deprecated `action-prepare-release` to the new Craft GitHub Actions.

## Changes

- Migrated `.github/workflows/release.yml` to Craft reusable workflow

## Documentation

See https://getsentry.github.io/craft/github-actions/ for more information.


szokeasaurusrex (approved):
made a whitespace and comment change, I hope that is ok!

BYK:
@szokeasaurusrex if you set the comment to a specific version like `2.19.0` then dependabot will never update it. Ideally we'd just use the `v2` floating tag but at least we should use `v2` as the comment to signal that we just pinned the SHA and want to keep upgrading

szokeasaurusrex:
@BYK ive got an action that bumps all the shas and the version numbers every week

PR #3085, opened 2026-01-14
fix(api): Retry API requests on DNS resolution failure

### LINKED ISSUE #2763: Retry API request when DNS resolution of sentry.io fails
### Problem Statement

This is a follow-up to [#2177: API request failed caused by: [6] Couldn't resolve host name (Could not resolve host: sentry.io)](https://github.com/getsentry/sentry-cli/issues/2177)

We are seeing this when uploading symbol files from Linux hosts, with a failure rate of <10% (4 of the previous 50 invocations, at time of writing).

I suspect it is caused by unreliable `.io` TLD DNS servers that is somehow amplified by some combination of a DNS stack on Linux.

If I'm grokking the retry logic correctly, it only attempts retries on certain HTTP status codes, and does not cover the cases from other parts of the stack:

```rust
const RETRY_STATUS_CODES: &[u32] = &[
    http::HTTP_STATUS_502_BAD_GATEWAY,
    http::HTTP_STATUS_503_SERVICE_UNAVAILABLE,
    http::HTTP_STATUS_504_GATEWAY_TIMEOUT,
    http::HTTP_STATUS_507_INSUFFICIENT_STORAGE,
    http::HTTP_STATUS_524_CLOUDFLARE_TIMEOUT,
];

// ...

    pub fn send(mut self) -> ApiResult<ApiResponse> {
        // -- snip --
        loop {
            let mut out = vec![];
            debug!("retry number {retry_number}, max retries: {max_retries}",);

            let mut rv = self.send_into(&mut out)?;
            if retry_number >= max_retries || !RETRY_STATUS_CODES.contains(&rv.status) {
                rv.body = Some(out);
                return Ok(rv);
            }

            // -- snip --
        }
    }
```

Implementing retry to DNS resolution failure should alleviate this issue.

### Solution Brainstorm

_No response_

szokeasaurusrex:
### Description

Sentry CLI now retries requests that fail to resolve `sentry.io`, addressing intermittent failures for some users uploading symbol files. Retries are limited to `sentry.io` hosts to avoid masking configuration issues for other domains.

### Issues

* Fixes #2763
* Fixes [CLI-168](https://linear.app/getsentry/issue/CLI-168/retry-api-request-when-dns-resolution-of-sentryio-fails)

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1317,20 +1317,17 @@ impl ApiRequest {
             debug!("retry number {retry_number}, max retries: {max_retries}");
             *retry_number += 1;
 
-            let mut rv = match self.send_into(&mut out) {
-                Ok(rv) => rv,
-                Err(err) => {
-                    // Retry DNS failures for sentry.io, as these likely indicate
-                    // a network issue. DNS failures for other domains should not
-                    // be retried, to avoid masking configuration problems (e.g.
-                    // if the user has mistyped their self-hosted URL).
-                    if is_dns_error(&err) && self.is_sentry_io_host() {
-                        anyhow::bail!(RetryError::from(err));
-                    }
-
-                    anyhow::bail!(err);
+            let mut rv = self.send_into(&mut out).map_err(|err| {
+                // Retry DNS failures for sentry.io, as these likely indicate
+                // a network issue. DNS failures for other domains should not
+                // be retried, to avoid masking configuration problems (e.g.
+                // if the user has mistyped their self-hosted URL).
+                if is_dns_error(&err) && self.is_sentry_io_host() {
+                    anyhow::anyhow!(RetryError::from(err))
+                } else {
+                    anyhow::anyhow!(err)
                 }
-            };
+            })?;
 
             rv.body = Some(out);
 


src/api/errors/mod.rs
```diff
@@ -14,17 +14,12 @@ pub(super) struct ProjectRenamedError(pub(super) String);
...
-    }
+pub(super) enum RetryError {
+    #[error("request failed with retryable status code {}", body.status)]
+    Status { body: ApiResponse },
+    #[error("request failed with retryable error: {source}")]
+    ApiError {
+        #[from]
+        source: ApiError,
+    },
 }
```
loewenheim, line 25: It reads weirdly to me that an `ApiError` is a kind of `RetryError`—if anything, I would expect the inverse. I would suggest slightly renaming this.
```suggestion
pub(super) enum RequestError {
    #[error("request failed with retryable status code {}", body.status)]
    Retry { body: ApiResponse },
    #[error(transparent)]
    ApiError {
        #[from]
        source: ApiError,
    },
}
```

loewenheim: Cleared up on Slack, I misunderstood the intention of this.

PR #3088, opened 2026-01-15
feat(js-lib): Add `sourceMaps.inject()` for injecting debug IDs

s1gr1d:
### Description
The JS Bundler Plugins repo uses the CLI instance in the Build Plugin Manager. However, injecting debug IDs is handled with a "raw" `.execute()`. This PR adds this directly to the CLI instance under the namespace `sourceMaps`. Uploading source maps can be handled through this namespace as well in the future.

related PR: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/836

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,5 +14,5 @@
 - For external contributors: [CONTRIBUTING.md](https://github.com/getsentry/sentry-cli/blob/master/CONTRIBUTING.md), [Sentry SDK development docs](https://develop.sentry.dev/sdk/), [Discord community](https://discord.gg/Ww9hbqr)
 -->
 
-
-
+### Legal Boilerplate
+Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.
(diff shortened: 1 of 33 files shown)


lib/releases/index.ts
```diff
@@ -120,6 +120,7 @@ export class Releases {
    * @param options Options to configure the source map upload.
    * @returns A promise that resolves when the upload has completed successfully.
    */
+  // TODO: Move `uploadSourceMaps()` from Releases to SourceMaps class as `.upload()`
```
Lms24: SGTM, but I think we can leave this command in place to avoid breaking anyone once we move things over.

JPeer264 (approved):
LGTM

lib/releases/index.ts
s1gr1d: I changed the comment a bit so we know of this in the future

lib/releases/index.ts
```diff
@@ -120,6 +120,7 @@ export class Releases {
    * @param options Options to configure the source map upload.
    * @returns A promise that resolves when the upload has completed successfully.
    */
+  // TODO: Add `uploadSourceMaps()` to SourceMaps class as `.upload()` (keep it here too for backward compatibility)
```
szokeasaurusrex, line 123: @s1gr1d did you mean to address this TODO in the current changes?

s1gr1d: No, that would be another PR. It's an unrelated change

szokeasaurusrex: m: Got it, is there an issue for this already? If not, can you please make one? 🙏 

l: I would suggest we delete the `// TODO` comment as the issue should suffice.

```suggestion
```

PR #3089, opened 2026-01-15
meta: add legal boilerplate to PR template

dingsdax:
Added legal boilerplate to clarify contribution rights b/c CLI is FSL

### DIFF
--- a/.github/workflows/CLAUDE.md
+++ b/.github/workflows/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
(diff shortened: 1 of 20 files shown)


dingsdax:
I don't think this is necessary. Upon opening a PR an external contributor reads this language, there needs to be a  conscious action to remove it. I don't think we have such a check for our other FSL licensed repos either.

szokeasaurusrex:
it is possible that they miss the template if they are opening the PR from `gh` CLI, which may become a more frequent use case with agentic coding workflows

dingsdax:
still edge case territory for me. afaik GH CLI respect PR templates unless you override with "gh pr create --body", but that might be a nice addition to our Dangerfile … fine lets create a backlog issue

PR #3094, opened 2026-01-20
feat(build): Add --install-group parameter to build upload

noahsmartin:
Add support for install groups in build uploads. This parameter allows specifying one or more groups that control update visibility. Builds with at least one matching install group will be shown updates for each other.


src/commands/build/upload.rs
```diff
@@ -175,6 +185,10 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let build_configuration = matches.get_one("build_configuration").map(String::as_str);
     let release_notes = matches.get_one("release_notes").map(String::as_str);
+    let install_groups: Vec<String> = matches
+        .get_many::<String>("install_group")
```
szokeasaurusrex: l: this is redundant

```suggestion
        .get_many("install_group")
```

PR #3110, opened 2026-01-28
feat(preprod): Add snapshots subcommand

lcian:
Updated version of https://github.com/getsentry/sentry-cli/pull/3049 to discuss and iterate on things.

Notable changes:
- Removed `shard_index` parameter from the command. I'm not sure what the purpose of that was originally.
- This uses the new `many` (batch) API from `objectstore_client`. All uploads are executed as batch requests, reducing network overhead. Unfortunately, with they way things are implemented now, we will still have to buffer all files in memory before sending the request, as we need to hash their contents to determine the filename. If we could just use the filename as the key in objectstore, it would be much better because that way we could stream the files over.

Note that the objectstore endpoint is currently gated by a feature flag, and only enabled for internal orgs/teams.

Ref FS-233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Added experimental `sentry-cli build snapshots` command to upload build snapshots to a project ([#3110](https://github.com/getsentry/sentry-cli/pull/3110)).
+  - This command uploads files from a specified directory to Sentry's Objectstore, associating them with a snapshot identifier.
+  - The command is experimental and subject to breaking changes or removal in future releases.
+
 ## 3.2.2
 
 ### Fixes
@@ -17,9 +25,6 @@
 
 ### Features
 
-- Added experimental `sentry-cli build snapshots` command to upload build snapshots to a project ([#3110](https://github.com/getsentry/sentry-cli/pull/3110)). 
-  - This command uploads files from a specified directory to Sentry's Objectstore, associating them with a snapshot identifier.
-  - The command is experimental and subject to breaking changes or removal in future releases.
 - Add `sourceMaps.inject()` for injecting debug IDs ([#3088](https://github.com/getsentry/sentry-cli/pull/3088))
 - Add `--install-group` parameter to `sentry-cli build upload` for controlling update visibility between builds ([#3094](https://github.com/getsentry/sentry-cli/pull/3094))
 


src/api/mod.rs
```diff
@@ -1766,6 +1768,18 @@ pub struct Organization {
...
+#[derive(Deserialize, Debug)]
+pub struct OrganizationLinks {
+    #[serde(rename = "regionUrl")]
+    pub region_url: String,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct OrganizationDetails {
+    pub id: String,
+    pub links: OrganizationLinks,
```
lcian: This endpoint has a bunch more information but what we really care about is the `id` and `links`.
The `region_url` contains the regional API url, e.g. `us.sentry.io`. This way we can hit that endpoint directly instead of going through control silo (`sentry.io`).

If I understand correctly I think this information might already be somehow embedded in org tokens, but not in personal tokens. Or maybe I'm confusing this with something else.

src/utils/api/mod.rs
```diff
@@ -0,0 +1,41 @@
+//! Utilities that rely on the Sentry API.
+
+use crate::api::{Api, AuthenticatedApi};
+use anyhow::{Context, Result};
+
+/// Given an org and project slugs or IDs, returns the IDs of both.
+pub fn get_org_project_id(api: impl AsRef<Api>, org: &str, project: &str) -> Result<(u64, u64)> {
```
lcian: As far as I know, for all commands a user can provide `--project` and `--org` as slugs or IDs.
These utils are needed to get the corresponding IDs, so that objects from the same org/proj all have the same paths in objectstore regardless if the user passed in slugs or IDs.

lcian: It might be weird to take in `Api` and it might be possible that these functions should live somewhere else, IDK.
They certainly don't belong to the `Api` struct as its methods seem to all map 1to1 to Sentry API calls.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ java-properties = "2.0.0"
 lazy_static = "1.4.0"
 libc = "0.2.139"
 log = { version = "0.4.17", features = ["std"] }
+objectstore-client = { git =  "https://github.com/getsentry/objectstore.git", branch = "lcian/feat/rust-batch-client" }
```
lcian: This indirectly adds a dependency to `reqwest` and we need to double check what the implications of that are, especially in regard to `rustls` vs `native-tls` which could be problematic.

lcian: It also adds a bunch of deps but I think most of them are inevitable.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,210 @@
...
+    files: &[std::path::PathBuf],
+    org: &str,
+    project: &str,
+    snapshot_id: &str,
+) -> Result<()> {
+    let url = get_objectstore_url(Api::current(), org)?;
+    let client = Client::new(url)?;
+
+    let (org, project) = get_org_project_id(Api::current(), org, project)?;
+    let session = Usecase::new("preprod")
```
lcian: We should set an expiration policy here.
I think based on our discussions TTL should be ideal for this usecase?

lcian: As we discussed we want different TTL based on customer/plan. We could introduce a new endpoint in sentry that tells you this info and set the TTL here based on that.

szokeasaurusrex:
Hey @lcian, are you ready for me to review this, or are you still planning to check the feedback from Bugbot and/or iterate further?

src/commands/build/snapshots.rs
rbro112: Yep, for now I think we're good with nothing set just for testing, but I'll track this here for my team to follow up on in the near future: https://linear.app/getsentry/issue/EME-839/endpoint-for-determining-ttl-for-an-org

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,231 @@
...
+            let filename = file_path
+                .file_name()
+                .and_then(|name| name.to_str())
+                .unwrap_or("unknown.json");
+            format!("{org}/{snapshot_id}/{filename}")
+        } else {
+            // For other files, use {org}/{project}/{hash}
+            let hash = compute_sha256_hash(&contents);
+            format!("{org}/{project}/{hash}")
+        };
```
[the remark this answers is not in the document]
rbro112: I'll handle this and other business logic related to the emerge snapshot features

rbro112:
* **#3110** <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3110?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/utils/api/mod.rs
runningcode: Can the backend interpret the slugs in to ids instead of the frontend?

szokeasaurusrex: > Can the backend interpret the slugs in to ids instead of the frontend?

Indeed, I agree that this would be preferable. The current code adds two additional API calls that could be avoided if the backend resolves the slugs/IDs appropriately.

src/api/mod.rs
```diff
@@ -1766,6 +1788,18 @@ pub struct Organization {
     pub features: Vec<String>,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct OrganizationLinks {
+    #[serde(rename = "regionUrl")]
+    pub region_url: String,
+}
```
szokeasaurusrex: nit: you can use [rename_all](https://serde.rs/container-attrs.html#rename_all) to avoid duplicating the field name

```suggestion
#[derive(Deserialize, Debug)]
#[serde(rename_all = "camelCase")]
pub struct OrganizationLinks {
    pub region_url: String,
}
```

src/api/mod.rs
szokeasaurusrex: > If I understand correctly I think this information might already be somehow embedded in org tokens, but not in personal tokens

This is 100% correct

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+        )
+}
+
+#[derive(Deserialize)]
+struct CreateSnapshotResponse {
+    #[serde(rename = "artifactId")]
+    artifact_id: String,
+    #[serde(rename = "imageCount")]
+    image_count: u64,
+}
```
szokeasaurusrex: l:

```suggestion
#[derive(Deserialize)]
#[serde(rename_all = "camelCase")]
struct CreateSnapshotResponse {
    artifact_id: String,
    image_count: u64,
}
```

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+        style(">").dim(),
+        style(images.len()).yellow(),
+        if images.len() == 1 { "file" } else { "files" }
+    );
+
+    // Build manifest from discovered images
+    let mut manifest = json!({
+        "app_id": app_id,
+        "images": {},
+    });
```
szokeasaurusrex: m: Why is this being custom-constructed as a JSON object?

I think it might make more sense to make a dedicated `struct` type for this manifest. Is there any reason why this was not done?

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+    for image in &images {
+        images_obj.insert(
+            image.hash.clone(),
+            json!({
+                "file_name": image.relative_path,
+                "width": image.width,
+                "height": image.height,
+            }),
+        );
+    }
```
szokeasaurusrex: l: Following from my comment above, if we make a `struct Manifest`, this logic could be an implementation on the struct.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+            path: path.to_path_buf(),
+            relative_path: relative,
+            hash,
+            width,
+            height,
+        });
+    }
+
+    Ok(images)
+}
```
szokeasaurusrex: l: I think an iterator chain would be more readable than a long for loop with a mutable vector here. 

I vibe-refactored it like this, with a helper function:

```suggestion
fn collect_images(dir: &Path) -> Result<Vec<ImageInfo>> {
    WalkDir::new(dir)
        .follow_links(true)
        .into_iter()
        .filter_entry(|e| !is_hidden(e.path()))
        .filter_map(Result::ok)
        .filter(|entry| entry.file_type().is_file())
        .filter(|entry| is_image_file(entry.path()))
        .map(|entry| collect_image_info(dir, entry.path()))
        .filter_map(|result| result.transpose())
        .collect()
}

/// Builds [`ImageInfo`] for a discovered image path during snapshot collection.
///
/// Returns `Ok(Some(ImageInfo))` when the file is readable and its dimensions can
/// be parsed, `Ok(None)` when the file should be skipped (currently when image
/// dimensions cannot be determined), and `Err` for hard failures such as I/O
/// errors reading the file.
fn collect_image_info(dir: &Path, path: &Path) -> Result<Option<ImageInfo>> {
    let contents = fs::read(path).with_context(|| format!("Failed to read: {}", path.display()))?;

    let (width, height) = match read_image_dimensions(&contents) {
        Some(dims) => dims,
        None => {
            warn!("Could not read dimensions from: {}", path.display());
            return Ok(None);
        }
    };

    let relative = path
        .strip_prefix(dir)
        .unwrap_or(path)
        .to_string_lossy()
        .to_string();

    let hash = compute_sha256_hash(&contents);

    Ok(Some(ImageInfo {
        path: path.to_path_buf(),
        relative_path: relative,
        hash,
        width,
        height,
    }))
}
```

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+        if i + 1 < data.len() {
+            let length = u16::from_be_bytes([data[i], data[i + 1]]) as usize;
+            i += length;
+        } else {
+            break;
+        }
+    }
+
+    None
+}
```
szokeasaurusrex: l: I wonder if there's a crate that does this.

m: If we want to implement the logic ourselves, I would suggest we move this code to a utility file and also add some tests that check it against real PNG and JPEG files

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+
+fn upload_images(images: &[ImageInfo], org: &str, project: &str) -> Result<()> {
+    let config = Config::current();
+    let auth = config
+        .get_auth()
+        .ok_or_else(|| anyhow::anyhow!("Authentication required"))?;
+    let token = match auth {
+        Auth::Token(token) => token.raw().expose_secret(),
+    };
```
szokeasaurusrex: m/h: These lines are potentially risky.

The CLI stores tokens as a [`SecretString`](https://docs.rs/secret-string/latest/secret_string/struct.SecretString.html) to prevent tokens from accidentally being logged in debug logs (e.g. if a panic were to occur). As soon as we expose the secret, we risk leaking the token, so we should only do this as late as possible.

Ideally, the objectstore SDK would allow us to pass the token as a `SecretString`

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    let url = get_objectstore_url(Api::current(), org)?;
+    let client = ClientBuilder::new(url.clone())
+        .configure_reqwest(move |r| {
+            let mut headers = http::HeaderMap::new();
+            headers.insert(
+                AUTHORIZATION,
+                HeaderValue::from_str(&format!("Bearer {token}"))
+                    .expect("always a valid header value"),
```
szokeasaurusrex: m/h: I'm pretty sure this could fail; we should not `expect` here. Tokens are not hard-validated in the CLI, and so could contain invisible or non-ASCII characters, which are [invalid in headers](https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.from_str). This is potentially risky as the panic might log the token value (see my comment above).

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        let contents = fs::read(&image.path)
+            .with_context(|| format!("Failed to read image: {}", image.path.display()))?;
+
+        let obj_key = format!("{org_id}/{project_id}/{}", image.hash);
+
+        info!("Queueing {} as {obj_key}", image.path.display());
+
+        many_builder = many_builder.push(session.put(contents).key(&obj_key));
+    }
```
szokeasaurusrex, line 258: h: The `fs::read` call will read the images into memory, and my clanker thinks (seemingly correctly) that all the images will remain in memory until the upload occurs:

> - In `sentry-cli`, the loop reads each file into memory ([`fs::read(&image.path)`](https://github.com/getsentry/sentry-cli/blob/0d59c90b648c741d857248544e2fe48f77651644/src/commands/build/snapshots.rs#L306-L307)), pushes it into `many_builder` ([`session.put(contents)`](https://github.com/getsentry/sentry-cli/blob/0d59c90b648c741d857248544e2fe48f77651644/src/commands/build/snapshots.rs#L313)), and only then calls [`many_builder.send()`](https://github.com/getsentry/sentry-cli/blob/0d59c90b648c741d857248544e2fe48f77651644/src/commands/build/snapshots.rs#L316-L317) after the loop.
> - In `objectstore-client`, `Session::put` stores input as [`PutBody::Buffer(body.into())`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/put.rs#L53-L54), `PutBuilder` owns [`body: PutBody`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/put.rs#L74-L78), and `ManyBuilder` stores operations in [`Vec<BatchOperation>`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/many.rs#L33-L35) via [`push`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/many.rs#L369-L370); the `PutBuilder` body is moved into batch operations ([`body: value.body`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/many.rs#L76-L85)) and processed when [`send`](https://github.com/getsentry/objectstore/blob/ba4015698dbf3f84f4bc11e5d95b42e55cc1ee73/clients/rust/src/many.rs#L312-L320) drains that vector.

Does this look accurate to you?

If it does, and if the cumulative size of the snapshots could be large, we should probably find a way to avoid loading all the snapshots into memory. Otherwise, if the cumulative size should be small, we should perhaps limit this here and error if the size exceeds a threshold.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+fn collect_images(dir: &Path) -> Result<Vec<ImageInfo>> {
+    let mut images = Vec::new();
+
+    for entry in WalkDir::new(dir)
+        .follow_links(true)
+        .into_iter()
+        .filter_entry(|e| !is_hidden(e.path()))
+        .filter_map(Result::ok)
```
szokeasaurusrex: m: This will silently discard any errors as we iterate the directory (e.g. due to missing permissions). This could lead to a scenario where an upload succeeds, but only some of the files the user intended to upload are uploaded. 

We should probably log any errors as warning messages prior to filtering them out like this.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+    );
+
+    Ok(())
+}
+
+fn collect_images(dir: &Path) -> Result<Vec<ImageInfo>> {
+    let mut images = Vec::new();
+
+    for entry in WalkDir::new(dir)
+        .follow_links(true)
```
szokeasaurusrex, line 131: l: This will also follow symbolic links to locations outside the upload path, which may not be desirable. However, we use this option in other places where we iterate over directories, so I suppose it is ok to keep

src/utils/objectstore/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Utilities to work with the Objectstore service.
+
+use anyhow::Result;
+
+use crate::api::Api;
+
+pub fn get_objectstore_url(api: impl AsRef<Api>, org: &str) -> Result<String> {
+    let api = api.as_ref().authenticated()?;
+    let base = api.fetch_organization_details(org)?.links.region_url;
+    Ok(format!("{base}/api/0/organizations/{org}/objectstore"))
```
szokeasaurusrex: m: We should be using [`PathArg`](https://github.com/getsentry/sentry-cli/blob/0d59c90b648c741d857248544e2fe48f77651644/src/api/encoding.rs#L24) here to properly escape the org.

```suggestion
    Ok(format!("{base}/api/0/organizations/{}/objectstore", PathArg(org)))
```

src/utils/objectstore/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Utilities to work with the Objectstore service.
+
+use anyhow::Result;
+
+use crate::api::Api;
+
+pub fn get_objectstore_url(api: impl AsRef<Api>, org: &str) -> Result<String> {
+    let api = api.as_ref().authenticated()?;
```
szokeasaurusrex: l: The pattern elsewhere in the CLI codebase is typically for functions to take an `AuthenticatedApi` where it is required, allowing the calling code to handle the case where auth is missing.

```suggestion
pub fn get_objectstore_url(api: &AuthenticatedApi<'_>, org: &str) -> Result<String> {
```

Cargo.toml
szokeasaurusrex: Let's discuss the TLS stuff next week, because I am uncertain what problems you're concerned about.

szokeasaurusrex: h: Before we merge these changes, we should ensure we are depending on a version that's been published to crates.io, not a Git branch.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,349 @@
...
+    let config = Config::current();
+    let auth = config
+        .get_auth()
+        .ok_or_else(|| anyhow::anyhow!("Authentication required"))?;
+    let token = match auth {
+        Auth::Token(token) => token.raw().expose_secret(),
+    };
+
+    let url = get_objectstore_url(Api::current(), org)?;
+    let client = ClientBuilder::new(url.clone())
```
szokeasaurusrex: l: `clone()` appears to be unnecessary 

```suggestion
    let client = ClientBuilder::new(url)
```

src/utils/api/mod.rs
szokeasaurusrex, line 1: l: I would move this to `src/utils/api.rs` for now

src/utils/objectstore/mod.rs
szokeasaurusrex, line 1: l: Would move to `src/utils/objectstore.rs` for now.

src/utils/api/mod.rs
lcian: Not sure what backend and frontend here refers to.
How objectstore works is that the scope (org and project in this case) is an arbitrary sequence of key-value pairs. We recommend using org and project ID, or at least org ID, but that's not mandatory.
Therefore it's not a responsibility of objectstore or its endpoint in the monolith to normalize org and project, we simply take what the user provides. It's responsibility of the user to normalize.

src/api/mod.rs
lcian: Ok then I should see how to resolve the id from the token instead of using the API call when we have an org token.

szokeasaurusrex: To clarify, only the region URL is embedded. Org ID is not

szokeasaurusrex: Let's discuss about the need to resolve the org and project IDs in the first place. Almost all other endpoints are capable of accepting both slugs and IDs interchangeably in the backend (Sentry side). To me it seems like a pretty serious limitation if objectstore cannot accept either.

That said, what you mentioned about having arbitrary keys in your other comment makes sense, and I guess it makes it harder to accept both IDs and slugs. Let's see if we can come up with a solution here, though

src/commands/build/snapshots.rs
lcian: That's true. Objectstore and its clients are designed for streaming, so it would be great if we could iterate on the files one by one.
Unfortunately we need to hash each image (which obviously requires loading the contents of the file in memory) and that seems to be a core way of how snapshots diffing works in the backend.

What we could do is to call `many_builder.send()` periodically on every X requests instead of just once (possibly in its own task so that we can keep iterating on the files concurrently), that would increase the number of batch requests to objectstore, with the advantage of being able to free up memory as we send those requests out.
So it seems that this is a latency/memory utilization tradeoff after all.

src/api/mod.rs
lcian: Objectstore can accept either but the problem is that if a user runs
`sentry-cli ... snapshots upload --org lorenzo`
and then
`sentry-cli ... snapshots upload --org 1337`
where 1337 is the org id for `lorenzo`, then we need to normalize it somewhere. 
Otherwise, if we take the org verbatim, then the snapshots will be stored under different paths in objectstore, as if the two were different orgs.
And the responsibility of normalizing should not be in the objectstore endpoint in Sentry as that's supposed to just be an proxy that takes whatever request and forwards it with (ideally) no manipulation involved.
That's why I'm saying it should be responsibility of the user to normalize the IDs to write to the expected path if needed.
(Maybe this was already clear but wanted to be explicit in describing the situation)

Anyways let's discuss this next week.

szokeasaurusrex: I understand, and this is the point I'd like us to think about:

> That's why I'm saying it should be responsibility of the user to normalize the IDs to write to the expected path if needed.

Would be nice if the CLI did not need to perform this normalization. If objectstore doesn't do it, maybe we can introduce a layer in between on the server side that does. 

Admittedly I'm not too familiar with Objectstore's architecture, so maybe this is not a great idea at all, let's think about it though 😅

src/api/mod.rs
rbro112: @lcian will let you take this one as this is objectstore-related

src/commands/build/snapshots.rs
rbro112: No particular reason, will update!

src/commands/build/snapshots.rs
rbro112: @lcian will let you take this one as it's more objectstore related

src/commands/build/snapshots.rs
rbro112: Done!

src/commands/build/snapshots.rs
rbro112: I'd be happy to use a crate but wasn't sure of our general stance on adding dependencies so admittedly had claude spin this one up. If you're open to it, I think https://crates.io/crates/imagesize would be a perfect fit here and I can add it/replace these usages in its favor

src/commands/build/snapshots.rs
rbro112: Done!

src/commands/build/snapshots.rs
rbro112: @lcian will let you take this one as it's more objectstore related

src/commands/build/snapshots.rs
rbro112: Yeah I was following other implementations here as the source of truth, I don't think this edge case is really much of a concern. If a user has a symlink in place and uploading that image is unexpected behavior, I think that's something we can iterate on in the future to try to see what behavior they would expect.

src/commands/build/snapshots.rs
rbro112: Added some logging here!

src/utils/objectstore/mod.rs
rbro112: @lcian will let you take this one as it's more objectstore related

src/utils/objectstore/mod.rs
rbro112: @lcian will let you take this one as it's more objectstore related

Cargo.toml
rbro112: @lcian will let you take this one as it's more objectstore related

src/commands/build/snapshots.rs
rbro112: @lcian went ahead and did this as it was easy enough

src/utils/api/mod.rs
rbro112: @lcian will let you take this one as it's more objectstore related

src/utils/objectstore/mod.rs
rbro112: @lcian will let you take this one as it's more objectstore related

Cargo.toml
szokeasaurusrex: @lcian and I discussed offline today. We agreed we should use `native-tis` and `native-tls-vendored` (the latter possibly only on Linux), as these should pull in the same underlying dependencies as `curl`, thus limiting the increase in binary size.

src/api/mod.rs
runningcode: I'm happy to jump in on this discussion!

lcian: @runningcode We discussed this today and came up with a better idea: we could extend the `/preprod/retention` endpoint (https://github.com/getsentry/sentry-cli/pull/3110/changes#diff-4715d1ca31922b72b1d5c42d64759dfa3e247bf3fb4a9066d1c4711e6f8478f5R994)
to be something like `/preprod/upload-options` instead.
This could return the retention to use for a particular org, the Objectstore proxy URL where to perform the upload (with org and project normalized to IDs), as well as the token to use for Objectstore auth once we roll that out.

This would require just a single request pre-upload and has the additional benefit of leaving the door open if we ever decide to expose Objectstore to the internet at some point in the future: we would just need to change the URL we return to CLI.

What do you think @runningcode @rbro112?

lcian: PR to illustrate the idea: https://github.com/getsentry/sentry/pull/108312

rbro112: I like this idea quite a bit and have gone ahead and pre-stamped that sentry PR. I'm good if you want to land this backend API then update this command to use it @lcian !

lcian: Implemented this

lcian:
Converting to draft until CI passes and this is reviewable again

src/commands/build/snapshots.rs
rbro112: @szokeasaurusrex I pushed an update to use the imagesize crate, much nicer now!

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,378 @@
...
+            i += 1;
+            continue;
+        }
+
+        let marker = data[i + 1];
+        i += 2;
+
+        // SOF markers: C0-CF except C4 (DHT), C8 (JPG extension), and CC (DAC)
+        if (0xC0..=0xCF).contains(&marker) && marker != 0xC4 && marker != 0xC8 && marker != 0xCC {
+            if i + 7 < data.len() {
```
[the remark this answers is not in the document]
lcian: This should be outdated as we moved to the crate.

Cargo.toml
lcian: Addressed

src/utils/api/mod.rs
lcian: Addressed with a dedicated `snapshots/upload-options` endpoint.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,378 @@
...
+        .build()
+        .context("Failed to create tokio runtime")?;
+
+    let mut many_builder = session.many();
+
+    for image in images {
+        debug!("Processing image: {}", image.path.display());
+
+        let contents = fs::read(&image.path)
+            .with_context(|| format!("Failed to read image: {}", image.path.display()))?;
```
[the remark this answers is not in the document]
lcian: Fixed in https://github.com/getsentry/sentry-cli/pull/3110/changes/a5148fd3de72fea26392bb5619de950542182a12
using `imagesize::path` 
cc @rbro112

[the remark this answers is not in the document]
lcian: Fixed in https://github.com/getsentry/sentry-cli/pull/3110/changes/a5148fd3de72fea26392bb5619de950542182a12
using `imagesize::path` 
cc @rbro112

src/commands/build/snapshots.rs
lcian: According to my discussion with @rbro112 this is unavoidable as we need hashes for selective diffing in the backend.

lcian:
bugbot run

lcian:
Addressed almost all of the feedback. The only missing thing on my side is a release of the `objectstore-client` and subsequent change of the dependency here, which will come soon.
@szokeasaurusrex could we have a re-review?

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+            Err(err) => {
+                warn!("Failed to access file during directory walk: {err}");
+                None
+            }
+        })
+        .filter(|entry| entry.file_type().is_file())
+        .filter(|entry| is_image_file(entry.path()))
+        .filter_map(|entry| collect_image_info(dir, entry.path()))
+        .collect()
+}
```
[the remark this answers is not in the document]
lcian: Leaving this to @rbro112

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,307 @@
...
+            .into_owned();
+        manifest_entries.insert(
+            hash,
+            ImageMetadata {
+                image_file_name,
+                width: image.width,
+                height: image.height,
+            },
+        );
+    }
```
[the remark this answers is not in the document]
lcian: I believe this could've been an issue from the get go. Leaving it to @rbro112 to address this.

src/api/mod.rs
```diff
@@ -448,6 +448,14 @@ impl AuthenticatedApi<'_> {
         self.api.request(method, url, None)
     }
 
+    /// Returns the auth info for use in external service authorization.
+    pub fn get_auth(&self) -> &Auth {
```
szokeasaurusrex: l: This should be named `auth`; the Rust API Guidelines [generally recommend against using the `get_` prefix](https://rust-lang.github.io/api-guidelines/naming.html?utm_source=chatgpt.com#getter-names-follow-rust-convention-c-getter) in getters.

```suggestion
    pub fn auth(&self) -> &Auth {
```

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+use clap::{Arg, ArgMatches, Command};
+use console::style;
+use http::header::AUTHORIZATION;
+use log::{debug, info, warn};
+use objectstore_client::{ClientBuilder, ExpirationPolicy, Usecase};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest as _, Sha256};
+use walkdir::WalkDir;
+
+use secrecy::ExposeSecret as _;
```
szokeasaurusrex, line 11: l: `secrecy` [should be with](https://develop.sentry.dev/engineering-practices/rust/#import-order) the other crate imports

```suggestion
use objectstore_client::{ClientBuilder, ExpirationPolicy, Usecase};
use secrecy::ExposeSecret as _;
use serde::{Deserialize, Serialize};
use sha2::{Digest as _, Sha256};
use walkdir::WalkDir;
```

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+const IMAGE_EXTENSIONS: &[&str] = &["png", "jpg", "jpeg"];
+
+pub fn make_command(command: Command) -> Command {
+    command
+        .about("[EXPERIMENTAL] Upload build snapshots to a project.")
+        .long_about(format!(
+            "Upload build snapshots to a project.\n\n{EXPERIMENTAL_WARNING}"
+        ))
+        .org_arg()
+        .project_arg(false)
```
szokeasaurusrex, line 32: [Question] I am a bit confused that the help text says that this command uploads snapshots "to a project," but then, this command has `project_arg(false)`. 

Can you please explain this discrepancy? Perhaps we should consider updating help text wording.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+    images: HashMap<String, ImageMetadata>,
+}
+
+// Keep in sync with https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/snapshots/manifest.py
+#[derive(Serialize)]
+struct ImageMetadata {
+    image_file_name: String,
+    width: u32,
+    height: u32,
+}
```
szokeasaurusrex: l: I think these should be defined in the API module (or a submodule under API). That is where we typically define Serialization/Deserialization types.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+
+struct ImageInfo {
+    path: std::path::PathBuf,
+    relative_path: String,
```
szokeasaurusrex: l: Shouldn't we use `PathBuf` for both the `path` and the `relative_path`? It seems strange to use different types for each.

src/commands/build/snapshots.rs
szokeasaurusrex: m: I discussed offline with @lcian; we decided that we can avoid loading all files into memory, and instead load them only one at a time, by reading the files twice: once to compute the hash, and another time by `objectstore` as it uploads the files (via a stream).

This approach has the limitation that it assumes that the files remain unchanged between when the hashes are computed and when `objectstore` uploads them. This seems like a reasonable limitation, as the files are not expected to change. But, we should document this behavior in code comments in case it causes a bug in the future.

**Note**: It is OK with me if we merge this PR without this change, but it should be done in a follow-up soon afterwards

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+    let result = hasher.finalize();
+    format!("{result:x}")
+}
+
+fn is_hidden(path: &Path) -> bool {
+    path.file_name()
+        .and_then(|name| name.to_str())
+        .map(|name| name.starts_with('.'))
+        .unwrap_or(false)
+}
```
szokeasaurusrex: m: This implementation is brittle on Unix, as it will fail for any filenames that are not UTF-8, and it is incorrect for Windows, as Windows does not use `.` prefixes to indicate hidden files. See [my ChatGPT conversation](https://chatgpt.com/share/e/699ef48f-934c-800b-9851-aa1c5640dbc0) for more context.

The following is a more reliable platform-specific implementation.

```suggestion
/// Check if a file is hidden.
/// On Unix, a file is hidden if it starts with a `.`
/// On Windows, we check the file metadata's [`FILE_ATTRIBUTE_HIDDEN`] bit.
fn is_hidden(path: &Path) -> bool {
    #[cfg(unix)]
    let rv = path
        .file_name()
        .map(|name| name.as_bytes().first() == Some(&b'.'))
        .unwrap_or(false);

    #[cfg(windows)]
    let rv = path
        .metadata()
        .map(|meta| meta.file_attributes() & FILE_ATTRIBUTE_HIDDEN != 0)
        .unwrap_or(false);

    rv
}
```

The following imports need to be added, as well:

```rust
#[cfg(unix)]
use std::os::unix::ffi::OsStrExt as _;
#[cfg(windows)]
use std::os::windows::fs::MetadataExt as _;
#[cfg(windows)]
use windows_sys::Win32::Storage::FileSystem::FILE_ATTRIBUTE_HIDDEN;
```

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    let expiration = ExpirationPolicy::from_str(&options.objectstore.expiration_policy)
+        .context("Failed to parse expiration policy from upload options")?;
+
+    // TODO: replace with auth from `ObjectstoreUploadOptions` when available
+    let auth = match *authenticated_api.get_auth() {
+        Auth::Token(ref token) => {
+            format!("Bearer {}", token.raw().expose_secret())
+        }
+    };
```
szokeasaurusrex: m/h: [This comment](https://github.com/getsentry/sentry-cli/pull/3110#discussion_r2803741952) (copied below) still needs to be addressed.

> m/h: These lines are potentially risky.
> 
> The CLI stores tokens as a [`SecretString`](https://docs.rs/secret-string/latest/secret_string/struct.SecretString.html) to prevent tokens from accidentally being logged in debug logs (e.g. if a panic were to occur). As soon as we expose the secret, we risk leaking the token, so we should only do this as late as possible.
> 
> Ideally, the objectstore SDK would allow us to pass the token as a `SecretString`

szokeasaurusrex: Especially because, on the line below, we might print an error message that contains the token value

Cargo.toml
```diff
@@ -42,6 +42,7 @@ java-properties = "2.0.0"
 lazy_static = "1.4.0"
 libc = "0.2.139"
 log = { version = "0.4.17", features = ["std"] }
+objectstore-client = { git =  "https://github.com/getsentry/objectstore.git", branch = "lcian/feat/rust-batch-client", default-features = false, features = ["native-tls"] }
```
szokeasaurusrex: h: This should be replaced with a released version of `objectstore-client` before this PR gets merged

szokeasaurusrex (changes requested):
Looking mostly good! 

Most of the comments I left are optional (marked with "l"). The comment about optimizing memory usage (marked "m") can be implemented in a follow-up PR, as we agreed offline.

The two things blocking a merge, from my perspective, are that:
- We are depending on a branch of the `objectstore-client` repository, rather than a released version.
- I am a bit concerned about how we handle the auth token; it seems that we could easily end up logging the token in an error message

src/commands/build/snapshots.rs
lcian: `project_arg(false)` means it takes a single project, as opposed to multiple projects.

szokeasaurusrex: Huh, TIL 😂

src/commands/build/snapshots.rs
szokeasaurusrex: @rbro112 this one is probably for you

src/commands/build/snapshots.rs
lcian: Changed it to limit the scope of `auth` and use `objectstore_client::ClientBuilder::token`.
This internally stores the token in a struct with a custom debug implementation that redacts the token itself, see https://github.com/getsentry/objectstore/blob/3aa9fa58c8dcba9080437835b7bb635a5370c10c/clients/rust/src/auth.rs#L40.
That's equivalent to what `secrecy::SecretString` does, with the only difference being that `SecretString` also ensures that memory is zeroed on drop, which IMO we are not concerned with within the scope of CLI.

src/commands/build/snapshots.rs
lcian: changed

src/commands/build/snapshots.rs
lcian: Will change it in a follow-up, should be an easy change.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,281 @@
...
+    let api = Api::current();
+    let authenticated_api = api.authenticated()?;
+    let options = authenticated_api.fetch_snapshots_upload_options(org, project)?;
+
+    let expiration = ExpirationPolicy::from_str(&options.objectstore.expiration_policy)
+        .context("Failed to parse expiration policy from upload options")?;
+
+    let client = ClientBuilder::new(options.objectstore.url)
+        .token({
+            // TODO: replace with auth from `ObjectstoreUploadOptions` when appropriate
```
szokeasaurusrex, line 205:
```suggestion
```

l: I'd suggest we make an issue and no `TODO` comment

src/commands/build/snapshots.rs
szokeasaurusrex: l: consider extracting to separate utility

lcian: I've changed it to use `to_string_lossy`.
That works cross-platform and we're only interested in the first character anyways.
If anyone ever complains about this, we can add Windows-specific logic.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,281 @@
...
+            Err(err) => {
+                warn!("Failed to access file during directory walk: {err}");
+                None
+            }
+        })
+        .filter(|entry| entry.file_type().is_file())
+        .filter(|entry| is_image_file(entry.path()))
+        .filter_map(|entry| collect_image_info(dir, entry.path()))
+        .collect()
+}
```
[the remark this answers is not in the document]
lcian: fixed

src/commands/build/snapshots.rs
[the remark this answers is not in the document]
lcian: Fixed

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    let client = ClientBuilder::new(options.objectstore.url)
+        .token({
+            // TODO: replace with auth from `ObjectstoreUploadOptions` when appropriate
+            let auth = match authenticated_api.auth() {
+                Auth::Token(token) => token.raw().expose_secret().to_owned(),
+            };
+            auth
+        })
+        .build()?;
```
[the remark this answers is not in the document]
lcian: Intended, it will fail to compile if it ever changes.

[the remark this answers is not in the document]
lcian: This is passed to `reqwest::ClientBuider::bearer_auth` which internally creates `http::HeaderValue` with the sensitive flag enabled. Its debug impl looks at the flag and redacts the value if sensitive.

src/commands/build/snapshots.rs
lcian: Created a followup issue.

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,282 @@
...
+                .long("app-id")
+                .value_name("APP_ID")
+                .help("The application identifier.")
+                .required(true),
+        )
+}
+
+struct ImageInfo {
+    path: std::path::PathBuf,
+    relative_path: std::path::PathBuf,
```
szokeasaurusrex: l/nit: `use std::path::PathBuf` so we don't need to have the full name in the code

```suggestion
    path: PathBuf,
    relative_path: PathBuf,
```

src/commands/build/snapshots.rs
szokeasaurusrex: 🤷 technically it will allocate a string if conversion fails, but I guess this is OK

src/commands/build/snapshots.rs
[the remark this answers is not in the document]
szokeasaurusrex: @lcian Please consider making an issue so this comment doesn't get lost after we merge the PR

src/commands/build/snapshots.rs
lcian: https://github.com/getsentry/sentry-cli/issues/3159

src/commands/build/snapshots.rs
```diff
@@ -0,0 +1,282 @@
...
+                .key(&hash)
+                .expiration_policy(expiration),
+        );
+
+        let image_file_name = image
+            .relative_path
+            .file_name()
+            .unwrap_or_default()
+            .to_string_lossy()
+            .into_owned();
```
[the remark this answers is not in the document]
lcian: It was like this from the get go, I assume this was taken into account.

Issue #3063, opened 2026-01-07
`dart-symbol-map upload` does not resolve org from token


szokeasaurusrex:
All other commands typically resolve the org from an org auth token, but `dart-symbol-map upload` does not. We should align this behavior with other commands.

szokeasaurusrex:
@cursoragent attempt a fix. Make sure to add a test for the correct behavior. Observe how other file-uploading commands parse the org Auth Token

PR #3113, opened 2026-01-28
fix(dart-symbol-map): Resolve org from token via config

szokeasaurusrex:
### Description
- Switch `dart-symbol-map` to the clap builder pattern to align with most commands.
- Resolve org/project via existing `Config` helpers so org auth tokens take precedence.
- Add integration coverage for org-from-token behavior.

### Issues
* resolves: #3063
* resolves: [CLI-260](https://linear.app/getsentry/issue/CLI-260/dart-symbol-map-upload-does-not-resolve-org-from-token)

### Legal Boilerplate

### DIFF
--- a/tests/integration/upload_dart_symbol_map.rs
+++ b/tests/integration/upload_dart_symbol_map.rs
@@ -1,12 +1,10 @@
 use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::LazyLock;
+
+use serde_json::Value;
 
 use crate::integration::test_utils::AssertCommand;
 use crate::integration::{MockEndpointBuilder, TestManager};
-
-/// A test org auth token with org="wat-org" and empty URL.
-/// Format: sntrys_{base64_payload}_{base64_secret}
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,12 +1,3 @@
-# Adding new commands
-For new commands, it is recommended to use clap's [Derive API](https://docs.rs/clap/latest/clap/_derive/index.html).
-In contrast to the [Builder API](https://docs.rs/clap/latest/clap/_tutorial/index.html), the Derive API makes it:
-- Easier to read, write, and modify commands and arguments.
-- Easier to keep argument declaration and reading in sync.
-- Easier to reuse shared arguments.
-
-An existing example of how to use the Derive API is the `send-metric` command.
-
```
szokeasaurusrex, line 9: I still would like to eventually migrate to the derive API, but mixing the derive API with the builder API seems to be problematic; it is also the root cause of this bug.

So, if we migrate we should do it all at once

PR #3120, opened 2026-02-02
ref: Test PR

65a28e4:
This PR is built on #3119 and is meant to test https://github.com/getsentry/github-workflows/pull/145. If the workflow is working correctly, this PR would be rejected, as it is missing the legal notice.


empty
```diff
@@ -0,0 +1 @@
+empty
```
szokeasaurusrex, line 1: This is not empty 😭

PR #3127, opened 2026-02-03
fix(preprod): Understand ghe.com as github_enterprise

chromy:
To determine vcs_provider we use part of the origin of the URL:

https://github.com/foo/bar -> github
https://gitlab.com/abcdef -> gitlab

This adds special handling for github enterprize remotes. Now instead
of:

https://foo.ghe.com/foo/bar -> ghe

we end up with:

https://foo.ghe.com/foo/bar -> github_enterprise

as the backend expects.

Resolves EME-821

### DIFF
--- a/src/utils/vcs.rs
+++ b/src/utils/vcs.rs
@@ -220,6 +220,12 @@ impl VcsUrl {
 
 fn extract_provider_name(host: &str) -> &str {
     let trimmed = host.trim_end_matches('.');
+
+    // GitHub Enterprise Server uses *.ghe.com domains
+    if trimmed.ends_with(".ghe.com") {
+        return "github_enterprise";
+    }
+
     trimmed.rsplit('.').nth(1).unwrap_or(trimmed)
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/vcs.rs
```diff
@@ -220,6 +220,12 @@ impl VcsUrl {
 
 fn extract_provider_name(host: &str) -> &str {
     let trimmed = host.trim_end_matches('.');
+
+    // GitHub Enterprise Server uses *.ghe.com domains
+    if trimmed.ends_with(".ghe.com") {
+        return "github_enterprise";
```
szokeasaurusrex, line 226: Some optional improvements we could do here (both can wait for a separate PR):

1. We could extract `"github_enterprise"` to a constant.
2. We could define an enum for the supported providers, and include a fallback for unknown providers, like so:

    ```rust
    enum VcsProvider {
        Github,
        GithubEnterprise
        Gitlab,
        Other(String), // or &str
    }
    ```

   Then, we would return the enum from this method. We would probably need to add some serialization logic on the enum, as well as a method to convert an arbitrary string to the enum (probably a `From<&str>` impl works), so I guess there's some added complexity, but it would be clear what providers we support natively.

szokeasaurusrex (approved):
lgtm!

src/utils/vcs.rs
chromy: sounds good! will try to send a follow up shortly.

szokeasaurusrex: Cool! And as I said, I think these are just optional improvements, so no rush. It is also fine if we don't do it.

PR #3152, opened 2026-02-23
chore: Update minimatch to a non-vulnerable version

JPeer264:
Basically the same as this: https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/885/changes/7cf8cd004f1f95d11555df198e39032405ce857d

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3643,9 +3643,9 @@
       "dev": true
     },
     "minimatch": {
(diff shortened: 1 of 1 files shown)


szokeasaurusrex:
@JPeer264 do we also need to do this on `master`?

JPeer264:
Yes, but this one should go in first to unblock the others

PR #3154, opened 2026-02-23
ci(release): Default merge target to v2

szokeasaurusrex:
Ensure release runs from the v2 line merge release branches back into v2 instead of falling back to the repository default branch.

This avoids post-publish merge conflicts when cutting previous-major releases (e.g. 2.x) while master is on v3.

The workflow now sets a v2 default in dispatch inputs and also applies a runtime fallback when invoking action-prepare-release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 2.58.5
 
 ### Fixes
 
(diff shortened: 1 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -33,4 +34,4 @@ jobs:
         with:
           version: ${{ github.event.inputs.version }}
           force: ${{ github.event.inputs.force }}
-          merge_target: ${{ github.event.inputs.merge_target }}
+          merge_target: ${{ github.event.inputs.merge_target || 'v2' }}
```
szokeasaurusrex:
```suggestion
          merge_target: ${{ github.event.inputs.merge_target }}
```

PR #3163, opened 2026-02-27
feat(snapshots): Upload image metadata json data as part of snapshots job

rbro112:
### Description
First-class supported platforms (web library, iOS/Android in future) will be outputting a "sidecar" json file with each image containing custom metadata set by the framework and provided by the user. This will contain things like the "display name" among other items.

This updates our snapshots command to process these sidecar files if present and upload them as part of the image manifest data in the POST body.

### DIFF
--- a/src/api/data_types/snapshots.rs
+++ b/src/api/data_types/snapshots.rs
@@ -29,8 +29,6 @@ pub struct SnapshotsManifest {
 // Keep in sync with https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/snapshots/manifest.py
 /// Metadata for a single image in a snapshot manifest.
 ///
-/// Serializes as a flat JSON object.
-///
 /// CLI-managed fields (`image_file_name`, `width`, `height`) override any
 /// identically named fields provided by user sidecar metadata.
 #[derive(Debug, Serialize)]
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -351,3 +351,28 @@ fn upload_images(
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn make_image(width: u32, height: u32) -> ImageInfo {
+        ImageInfo {
+            path: PathBuf::from("img.png"),
+            relative_path: PathBuf::from("img.png"),
+            width,
+            height,
+        }
+    }
+
+    #[test]
+    fn test_validate_image_sizes_at_limit_passes() {
+        assert!(validate_image_sizes(&[make_image(8000, 5000)]).is_ok());
+    }
+
+    #[test]
+    fn test_validate_image_sizes_over_limit_fails() {
+        let err = validate_image_sizes(&[make_image(8001, 5000)]).unwrap_err();
+        assert!(err.to_string().contains("exceed the maximum pixel limit"));
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Experimental Feature 🧑‍🔬 (internal-only)
+
+- Pipe snapshot sidecar metadata into upload as part of `sentry-cli build snapshots` command ([#3163](https://github.com/getsentry/sentry-cli/pull/3163)).
+
 ## 3.3.0
 
 ### New Features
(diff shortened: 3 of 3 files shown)


rbro112:
* **#3163** <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3163?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3163?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

szokeasaurusrex:
Hey @rbro112, can you please resolve or mark as irrelevant all of the Seer code review comments before I review in more depth?

rbro112:
@szokeasaurusrex this is ready for review. I'm ignoring the Danger comment as the last item is `Added experimental sentry-cli build snapshots command`, so there's no need to explicitly mention this change assuming it makes it in before the next release

src/api/data_types/snapshots.rs
```diff
@@ -22,9 +23,41 @@ pub struct SnapshotsManifest {
...
     pub image_file_name: String,
     pub width: u32,
     pub height: u32,
+    pub extra: HashMap<String, serde_json::Value>,
+}
+
+const RESERVED_KEYS: &[&str] = &["image_file_name", "width", "height"];
+
+impl Serialize for ImageMetadata {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
```
rbro112: My rust knowledge isn't robust enough, but basically this is what I believe should be equivalent of the typescript "splat". Essentially what we want to do here is always override `"image_file_name", "width", "height"` if the user has provided those values. These are first-class fields we will set (will be documented) and users cannot be allowed to override those.

In typescript, this would be how I'd override:
```
const mergedMetadata = {
   ...extras, // User's fields
   width: ourWidth,
   height: ourHeight
   image_file_name: ourImageFileName,
}
```

It seems rust doesn't have an "override"/"splat" functionality like this, so the serializer/reserved keys is our workaround. Open to more elegant solutions here nonetheless.

szokeasaurusrex: m: @rbro112 I made a suggestion in https://github.com/getsentry/sentry-cli/pull/3178; basically, I would just have the struct only contain the `HashMap` and control field precedence by implementing a custom constructor.

src/commands/build/snapshots.rs
```diff
@@ -188,6 +189,40 @@ fn is_image_file(path: &Path) -> bool {
...
+        Ok(map) => map,
+        Err(err) => {
+            warn!(
+                "Failed to parse sidecar file {}: {err}",
+                sidecar_path.display()
+            );
+            HashMap::new()
+        }
+    }
+}
```
szokeasaurusrex: l: A few minor suggestions in one.

```suggestion
fn read_sidecar_metadata(image_path: &Path) -> Result<HashMap<String, Value>> {
    let sidecar_path = image_path.with_extension("json");
    if !sidecar_path.is_file() {
        return Ok(HashMap::new());
    }

    debug!("Reading sidecar metadata: {}", sidecar_path.display());

    let sidecar_file = File::open(&sidecar_path)
        .with_context(|| format!("Failed to open sidecar file {}", sidecar_path.display()))?;

    serde_json::from_reader(BufReader::new(sidecar_file)).with_context(|| {
        format!(
            "Failed to read sidecar file {} as JSON",
            sidecar_path.display()
        )
    })
}
```

The notable changes in order of importance:

1. Use [`serde_json::from_reader`](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html) to avoid reading the entire sidecar file to a string (should reduce memory usage for large sidecar files).
2. I'd return the error and handle it in the calling code; this allows you to use the `?` operator in the function, and factor out common error handling code (see associated comment at the call site)
3. Style change: import `serde_json::Value` so you can reference it as `Value` directly, this is the style recommended for types in [Rust's style guide](https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.).

You need to add the following imports at the beginning of the file, as well.

```rust
// std imports
use std::fs::{self, File};
use std::io::BufReader;

// crate imports
use serde_json::Value;
```

src/commands/build/snapshots.rs
```diff
@@ -247,13 +282,12 @@ fn upload_images(
             .unwrap_or_default()
             .to_string_lossy()
             .into_owned();
+
+        let extra = read_sidecar_metadata(&image.path);
```
szokeasaurusrex: This comment is associated with item (2) in my [previous comment](https://github.com/getsentry/sentry-cli/pull/3163/changes#r2882953883)

```suggestion
        let extra = read_sidecar_metadata(&image.path).unwrap_or_else(|err| {
            warn!("Error reading sidecar metadata, ignoring it instead: {err:#}");
            HashMap::new()
        });```

PR #3167, opened 2026-03-02
feat(snapshots): Print snapshot URL after upload

runningcode:
## Summary
- Adds `snapshot_url` (optional) to `CreateSnapshotResponse`
- Prints the snapshot URL after a successful upload so users can navigate directly to it

Depends on companion sentry PR: https://github.com/getsentry/sentry/pull/109662

Fixes EME-898

### DIFF
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -125,7 +125,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     if let Some(url) = &response.snapshot_url {
         println!(
-            "{} View snapshot at {}",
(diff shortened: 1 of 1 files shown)


src/commands/build/snapshots.rs
```diff
@@ -123,6 +123,14 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         }
     );
 
+    if let Some(url) = &response.snapshot_url {
+        println!(
+            "{} View snapshot at {}",
```
rbro112: "View snapshots"

szokeasaurusrex (approved):

Issue #3173, opened 2026-03-03
Expose command to compute UUID for ProGuard mapping


szokeasaurusrex:
Create a command, which can take a ProGuard mapping file, and log to `stdout` the UUID that we would assign it when uploading the file (assuming no `--uuid` manual override)

PR #3174, opened 2026-03-03
feat(proguard): `upload-proguard` → `proguard upload`

szokeasaurusrex:
Move the `sentry-cli upload-proguard` command to `sentry-cli proguard upload`.

This new API matches the API for the `debug-files upload` and `sourcemap upload` commands. This API will also enable us to add new subcommands, which will be useful for #3173.

`sentry-cli upload-proguard` is retained as an alias to `sentry-cli proguard upload`, and we intend to keep it around for the foreseeable future, so no one needs to migrate.


src/commands/upload_proguard/mod.rs
szokeasaurusrex, line 1: I put this in `upload_proguard/mod.rs` so that the `upload_proguard.rs` → `proguard/upload.rs` change shows up as a simple rename in this PR; that way, the Git history also is preserved.

I will open a follow-up PR on top to move this file from `upload_proguard/mod.rs` to `upload_proguard.rs` (see #3175)

tests/integration/proguard/upload.rs
szokeasaurusrex, line 1: These tests are basically copied from [`tests/integration/upload_proguard.rs`](https://github.com/getsentry/sentry-cli/blob/124418a39b58a45b0363489051991946b96dcdb6/tests/integration/upload_proguard.rs), but we use `json!` instead of manually constructing JSON strings.

src/commands/upload_proguard/mod.rs
lcian: Nice!

PR #3179, opened 2026-03-03
feat(snapshots): Add 40M pixel limit validation for snapshot images

NicoHinderling:
The backend enforces a 40M pixel limit per image during snapshot comparison
(`MAX_DIFF_PIXELS`), but oversized images were only rejected at comparison
time, wasting upload bandwidth. The CLI now checks `width * height` against
the 40,000,000 pixel threshold before uploading and reports all violations
with their dimensions.

Refs EME-885

### DIFF
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -6,6 +6,7 @@ use std::str::FromStr as _;
 use anyhow::{Context as _, Result};
 use clap::{Arg, ArgMatches, Command};
 use console::style;
+use itertools::Itertools as _;
 use log::{debug, info, warn};
 use objectstore_client::{ClientBuilder, ExpirationPolicy, Usecase};
 use secrecy::ExposeSecret as _;
@@ -51,6 +52,12 @@ struct ImageInfo {
     relative_path: PathBuf,
     width: u32,
     height: u32,
+}
+
+impl ImageInfo {
+    fn pixels(&self) -> u64 {
+        u64::from(self.width) * u64::from(self.height)
+    }
 }
 
 pub fn execute(matches: &ArgMatches) -> Result<()> {
@@ -330,6 +337,6 @@ mod tests {
     #[test]
     fn test_validate_image_sizes_over_limit_fails() {
         let err = validate_image_sizes(&[make_image(8001, 5000)]).unwrap_err();
-        assert!(err.to_string().contains("exceeded the maximum pixel limit"));
+        assert!(err.to_string().contains("exceed the maximum pixel limit"));
     }
 }


NicoHinderling:
actually not going to bother w a changelog since we're actively developing a feature that isn't available yet, so sharing about feature changes for it doesn't really make sense

src/commands/build/snapshots.rs
```diff
@@ -166,6 +169,37 @@ fn collect_image_info(dir: &Path, path: &Path) -> Option<ImageInfo> {
     })
 }
 
+fn validate_image_sizes(images: &[ImageInfo]) -> Result<()> {
```
rbro112, line 179: Why not just do this inline on line 161 and return None/error log if the size is larger than 40M?

NicoHinderling: The separate validate_image_sizes function is intentional. It collects and reports ALL oversized images in a single error message, so the user sees every violation at once.

Inlining with return None would silently skip oversized images and produce an incomplete snapshot without a hard failure. 

We want to fail loudly here rather than let users unknowingly upload partial snapshots.

src/commands/build/snapshots.rs
```diff
@@ -166,6 +169,37 @@ fn collect_image_info(dir: &Path, path: &Path) -> Option<ImageInfo> {
...
+        }
+        anyhow::bail!(
+            "{} image{} exceeded the maximum pixel limit of 40,000,000",
+            violations.len(),
+            if violations.len() == 1 { "" } else { "s" }
+        );
+    }
+
+    Ok(())
+}
```
szokeasaurusrex, line 202: l: I think the following may be a bit more readable:

```suggestion
fn validate_image_sizes(images: &[ImageInfo]) -> Result<()> {
    let mut violations = images
        .iter()
        .filter(|img| img.pixels() > MAX_PIXELS_PER_IMAGE)
        .map(|img| {
            let path = img.relative_path.display();
            let width = img.width;
            let height = img.height;
            let pixels = img.pixels();

            format!("  {path} ({width}x{height} = {pixels} pixels)")
        })
        .peekable();

    if violations.peek().is_some() {
        let violation_messages = violations.join("\n");

        anyhow::bail!(
            "The following images exceed the maximum pixel limit of {MAX_PIXELS_PER_IMAGE}:\n{violation_messages}",
        );
    }

    Ok(())
}
```

You'll also need to add the following:

```rust
// New import near top of file (adds .join method for iterators)
use itertools::Itertools as _;

// New .pixels() convenience method later in file
impl ImageInfo {
    /// Computes the total number of pixels in the image.
    fn pixels(&self) -> u64 {
        u64::from(self.width) * u64::from(self.height)
    }
}
```

szokeasaurusrex: In any case, I would use the constant `MAX_PIXELS_PER_IMAGE` in the error message (as I do in the suggestion above) to prevent the constant's value from diverging from the error message in the future.

src/commands/build/snapshots.rs
NicoHinderling: Keeping the separate validation function intentionally — it collects ALL violations and reports them in a single error message, so the user sees every oversized image at once rather than discovering them one-by-one. The reviewer's refactoring commit (4ec1b228) already refined this function with the cleaner peekable iterator approach.

- Claude Code

PR #3180, opened 2026-03-03
feat(snapshots): Use retention endpoint for objectstore expiration

NicoHinderling:
Related PR: https://github.com/getsentry/sentry/pull/109842

Fetch the snapshot retention value (in days) from
`GET /organizations/{org}/preprod/retention/` and use it to construct a
`TimeToLive` expiration policy for objectstore uploads, instead of using the
`expiration_policy` string from the upload-options response.

This allows snapshot retention to be configured per-organization via the
backend rather than being hardcoded in the objectstore upload options.

Refs EME-885

### DIFF
--- a/src/api/data_types/snapshots.rs
+++ b/src/api/data_types/snapshots.rs
@@ -12,6 +12,13 @@ pub struct CreateSnapshotResponse {
     pub image_count: u64,
 }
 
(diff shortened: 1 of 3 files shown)


NicoHinderling:
synced offline, forgot we have an endpoint dedicated for this case specifically for CLI already

Issue #3160, opened 2026-02-26
Improve Objectstore batching efficiency


lcian:
### Description

In the snapshots subcommand, we should avoid reading the file for the hash and buffering it.
We should instead compute the hash, and then stream it to Objectstore.

PR #3185, opened 2026-03-04
ref(snapshots): Use file-based upload for objectstore puts

lcian:
Use `put_file` instead of `put` when uploading images to objectstore,
streaming file contents from disk rather than loading entire files into
memory. This leverages the `File` variant added in objectstore-client
0.1.2.

Hash computation also now uses buffered reads instead of loading the
entire file into memory at once.

Close https://github.com/getsentry/sentry-cli/issues/3160
Close CLI-293

### DIFF
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::fs::{self, File};
+use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::str::FromStr as _;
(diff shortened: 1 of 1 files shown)


src/commands/build/snapshots.rs
```diff
@@ -271,15 +284,16 @@ fn upload_images(
     for image in images {
         debug!("Processing image: {}", image.path.display());
 
-        let contents = fs::read(&image.path)
-            .with_context(|| format!("Failed to read image: {}", image.path.display()))?;
-        let hash = compute_sha256_hash(&contents);
+        let hash = compute_sha256_hash(&image.path)?;
+        let file = runtime.block_on(File::open(&image.path)).with_context(|| {
```
lcian: This will be improved in https://github.com/getsentry/sentry-cli/pull/3187 by making `compute_sha256_hash` async too and entering the runtime earlier so we can `await` async things instead of using `block_on`.

lcian:
Merging now as it solves a problem for @NicoHinderling where images larger than a certain threshold would fail to upload due to an object size limit we have in the batch endpoint.
The newly introduced `put_file` API is size-aware and dispatches to the single object endpoint as needed, to ensure this failure mode is avoided.
We can still discuss this tomorrow if you wish @szokeasaurusrex.

PR #3192, opened 2026-03-09
fix(proguard): Drop redundant line info validation for mapping files

romtsn:
## Summary
- Remove the `has_line_info()` check that rejected proguard mapping files without line information — our proguard mapper can handle these files fine
- Remove `ProguardMappingError` enum and convert `TryFrom<ByteView>` to `From<ByteView>`
- Update `DifFile::is_usable()` to always return `true` for proguard mappings, consistent with the change

## Test plan
- [x] All 170 integration tests pass
- [x] Clippy clean
- [x] Removed obsolete `proguard-uuid-invalid-mapping.trycmd` test
- [x] Updated `proguard-upload-no-upload` and `upload_proguard-no-upload` test snapshots

### DIFF
--- a/src/utils/dif.rs
+++ b/src/utils/dif.rs
@@ -385,13 +385,17 @@ impl<'a> DifFile<'a> {
     }
 
     pub fn get_problem(&self) -> Option<&'static str> {
+        if self.is_usable() {
+            return None;
+        }
+
         match self {
-            DifFile::Archive(..) if !self.is_usable() => Some(if !self.has_ids() {
+            DifFile::Archive(..) => Some(if !self.has_ids() {
                 "missing debug identifier, likely stripped"
             } else {
                 "missing debug or unwind information"
             }),
-            DifFile::Archive(..) | DifFile::Proguard(..) => None,
+            DifFile::Proguard(..) => None,
         }
     }
 


src/utils/dif.rs
```diff
@@ -380,7 +380,7 @@ impl<'a> DifFile<'a> {
     pub fn is_usable(&self) -> bool {
         match self {
             DifFile::Archive(_) => self.has_ids() && self.features().has_some(),
-            DifFile::Proguard(pg) => pg.get().has_line_info(),
+            DifFile::Proguard(..) => true,
```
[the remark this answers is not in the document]
szokeasaurusrex: Seems accurate @romtsn https://github.com/getsentry/sentry-cli/blob/7a418025491f7d5af6558cc6d193b3575fdf0904/src/utils/dif.rs#L399

src/utils/dif.rs
```diff
@@ -380,24 +380,18 @@ impl<'a> DifFile<'a> {
     pub fn is_usable(&self) -> bool {
         match self {
             DifFile::Archive(_) => self.has_ids() && self.features().has_some(),
-            DifFile::Proguard(pg) => pg.get().has_line_info(),
+            _ => true,
```
szokeasaurusrex: l: Let's exhaustively match here; that ensures that if we add a new `DifFile` type in the future, we remember to update these lines.

```suggestion
            DifFile::Proguard(_) => true,
```

src/utils/dif.rs
```diff
@@ -380,24 +380,18 @@ impl<'a> DifFile<'a> {
...
-                }
-                DifFile::Proguard(..) => "missing line information",
-            })
+        match self {
+            DifFile::Archive(..) if !self.is_usable() => Some(if !self.has_ids() {
+                "missing debug identifier, likely stripped"
+            } else {
+                "missing debug or unwind information"
+            }),
+            _ => None,
```
szokeasaurusrex: l: Same comment as above

```suggestion
            DifFile::ProGuard(_) => None,
```

tests/integration/_cases/proguard/proguard-uuid-invalid-mapping.trycmd
szokeasaurusrex, line 1: m: Shouldn't we change this test (possibly with a file name rename) to ensure that this mapping can get uploaded successfully?

szokeasaurusrex (commented):
Left some comments; most important thing is about the tests

src/utils/dif.rs
```diff
@@ -380,25 +380,25 @@ impl<'a> DifFile<'a> {
...
+        Some(match self {
+            DifFile::Archive(..) => {
+                if !self.has_ids() {
+                    "missing debug identifier, likely stripped"
+                } else {
+                    "missing debug or unwind information"
+                }
+            }
+            DifFile::Proguard(..) => return None,
+        })
```
szokeasaurusrex: [nit] this is written kinda awkwardly; I think this version is a bit easier to understand. But up to you

```suggestion
        match self {
            DifFile::Archive(..) => Some(if !self.has_ids() {
                "missing debug identifier, likely stripped"
            } else {
                "missing debug or unwind information"
            }),
            DifFile::Proguard(..) => None,
        }
```

romtsn: so many ways to write it 😅

PR #3200, opened 2026-03-10
feat(preprod): Add VCS parameters to snapshots upload command

rbro112:
### Description
Adds vcs parameters to preprod snapshots upload command. Shares utilities with existing build upload command.

### Legal Boilerplate

### DIFF
--- a/tests/integration/_cases/build/build-upload-help-not-macos.trycmd
+++ b/tests/integration/_cases/build/build-upload-help-not-macos.trycmd
@@ -39,7 +39,7 @@ Options:
       --quiet
           Do not print any output while preserving correct exit code. This flag is currently
           implemented only for selected subcommands.
-
+          
           [aliases: --silent]
 
       --vcs-provider <vcs_provider>


rbro112:
* **#3200** <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/getsentry/sentry-cli/3200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/utils/build_vcs.rs
```diff
@@ -0,0 +1,205 @@
...
+    self, get_github_base_ref, get_github_head_ref, get_github_pr_number, get_provider_from_remote,
+    get_repo_from_remote_preserve_case, git_repo_base_ref, git_repo_base_repo_name_preserve_case,
+    git_repo_head_ref, git_repo_remote_url,
+};
+
+/// Collects git metadata from arguments and VCS introspection.
+///
+/// When `auto_collect` is false, only explicitly provided values are collected;
+/// automatic inference from git repository and CI environment is skipped.
+pub fn collect_git_metadata(
```
rbro112, line 20: No new logic in here, just extracted to be usable from both `upload.rs` and `snapshots.rs`

src/api/data_types/snapshots.rs
```diff
@@ -1,9 +1,10 @@
 //! Data types for the snapshots API.
 
-use std::collections::HashMap;
+use std::{borrow::Cow, collections::HashMap};
```
szokeasaurusrex: l/nit: Usually I prefer to only combine imports on the same line when the last line is in common, e.g. I would write `use a::b::{c,d}`, but would split `use a::{b::c, d::e}` instead of writing it in a single line. This prevents very long import statements.

```suggestion
use std::borrow::Cow
use std::collections::HashMap;
```

CHANGELOG.md
```diff
@@ -2,13 +2,10 @@
...
 ## Unreleased
 
-### Fixes
-
-- Accept ProGuard mapping files without line information instead of rejecting them ([#3192](https://github.com/getsentry/sentry-cli/pull/3192)).
-
 ### Experimental Feature 🧑‍🔬 (internal-only)
 
 - Pipe snapshot sidecar metadata into upload as part of `sentry-cli build snapshots` command ([#3163](https://github.com/getsentry/sentry-cli/pull/3163)).
+- Add VCS metadata parameters to `sentry-cli build snapshots` command ([#3200](https://github.com/getsentry/sentry-cli/pull/3200)).
```
szokeasaurusrex: l: Tbh, might make more sense to skip the changelog entry here, given this is internal-only for now.

src/api/data_types/snapshots.rs
```diff
@@ -21,9 +22,26 @@ pub struct CreateSnapshotResponse {
...
+    #[serde(skip_serializing_if = "str::is_empty")]
+    pub head_repo_name: Cow<'a, str>,
+    #[serde(skip_serializing_if = "str::is_empty")]
+    pub base_repo_name: Cow<'a, str>,
+    #[serde(skip_serializing_if = "str::is_empty")]
+    pub head_ref: Cow<'a, str>,
+    #[serde(skip_serializing_if = "str::is_empty")]
+    pub base_ref: Cow<'a, str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pr_number: Option<u32>,
```
szokeasaurusrex: m/h: why not reuse the existing [`VcsInfo`](https://github.com/getsentry/sentry-cli/blob/be4044c283f31b26f3554688bbbf88465f52e450/src/api/data_types/chunking/build.rs#L31-L50) type here, like we already do in [`ChunkedBuildRequest`](https://github.com/getsentry/sentry-cli/blob/be4044c283f31b26f3554688bbbf88465f52e450/src/api/data_types/chunking/build.rs#L8-L20)?

These seem to be copy-pasted from that existing type; I struggle to see why we would not want to reuse `VcsInfo` here.


```suggestion
    #[serde(flatten)]
    pub vcs_info: VcsInfo<'a>,
```

src/commands/build/snapshots.rs
```diff
@@ -47,6 +49,68 @@ pub fn make_command(command: Command) -> Command {
...
+                    If neither this nor --no-git-metadata is specified, git metadata is \
+                    automatically collected when running in most CI environments.")
+        )
+        .arg(
+            Arg::new("no_git_metadata")
+                .long("no-git-metadata")
+                .action(ArgAction::SetTrue)
+                .conflicts_with("force_git_metadata")
+                .help("Disable collection and sending of git metadata.")
+        )
```
szokeasaurusrex: l: Since these parameters also exist on [`upload.rs`](https://github.com/getsentry/sentry-cli/blob/be4044c283f31b26f3554688bbbf88465f52e450/src/commands/build/upload.rs), we may want to factor out this code to an extension trait, as we already do [for some commonly used arguments](https://github.com/getsentry/sentry-cli/blob/be4044c283f31b26f3554688bbbf88465f52e450/src/utils/args.rs#L84-L145). 

We could call the method `git_metadata_args()` or something like that.

src/utils/build_vcs.rs
szokeasaurusrex, line 1: h: This should go under `src/utils`, as it breaks the pattern of every file under `src/commands` corresponding to a single subcommand of Sentry CLI.

PR #3202, opened 2026-03-10
Revert Jest upgrade

JPeer264:
revert #3201 and #3191 because releasing won't work with this update.


JPeer264:
closed in favor in favor of #3194

PR #3207, opened 2026-03-12
feat(code-mappings): Add `code-mappings upload` command scaffold

romtsn:
_#skip-changelog_

Add the `sentry-cli code-mappings upload` subcommand group and the `upload`
subcommand with file parsing and validation.

This is the first in a stack of 4 PRs to support bulk uploading code mappings
from a JSON file — useful for Java/Android multi-module projects that need
dozens of mappings.

This PR adds:
- Command scaffold following the `repos`/`deploys` pattern
- JSON file reading and validation (empty arrays, empty stackRoot/sourceRoot)
- CLI args: positional `PATH`, `--repo`, `--default-branch`
- Help and no-subcommand trycmd integration tests

Stack: **#3207** → #3208 → #3209 → #3210

Backend PRs: getsentry/sentry#109783, getsentry/sentry#109785, getsentry/sentry#109786

Closes getsentry/sentry-android-gradle-plugin#1076
Closes getsentry/sentry-android-gradle-plugin#1077

PR #3208, opened 2026-03-12
feat(code-mappings): Add git inference for repo and default branch

romtsn:
*#skip-changelog*

When `--repo` or `--default-branch` are not provided, infer them from the<br>local git repository. Uses the configured VCS remote (SENTRY_VCS_REMOTE / ini)<br>first, then falls back to best-effort remote detection (upstream > origin > first).

Also extracts `find_best_remote()` as a shared utility in `src/utils/vcs.rs`,<br>replacing the inline logic that was duplicated in `git_repo_base_repo_name_preserve_case`.

Stack: getsentry/sentry-cli#3207 → [#3208](<https://github.com/getsentry/sentry-cli/issues/3208>) → [#3209](<https://github.com/getsentry/sentry-cli/issues/3209>) → [#3210](<https://github.com/getsentry/sentry-cli/issues/3210>)

Backend PRs: [getsentry/sentry#109783](<https://github.com/getsentry/sentry/issues/109783>), [getsentry/sentry#109785](<https://github.com/getsentry/sentry/issues/109785>), [getsentry/sentry#109786](<https://github.com/getsentry/sentry/issues/109786>)

Closes [GRADLE-79](https://linear.app/getsentry/issue/GRADLE-79/add-git-inference-for-repo-name-and-default-branch)

### DIFF
--- a/src/commands/code_mappings/upload.rs
+++ b/src/commands/code_mappings/upload.rs
@@ -87,6 +87,8 @@ fn resolve_repo_and_branch(
     git_repo: Option<&git2::Repository>,
 ) -> Result<(String, String)> {
     let (repo_name, remote_name) = if let Some(r) = explicit_repo {
+        // Try to find a local remote whose URL matches the explicit repo name,
+        // so we can infer the default branch from it. Falls back to None (-> "main").
         let remote = git_repo.and_then(|repo| find_remote_for_repo(repo, r));
         (r.to_owned(), remote)
     } else {


szokeasaurusrex:
Please check and resolve the open comments from the AI reviewers, before I do a full review. If any comments are inaccurate or non-actionable, please indicate that with a `👎` and mark the comment as resolved. Thanks! 🙏

src/commands/code_mappings/upload.rs
```diff
@@ -56,7 +59,70 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                configured_remote
+            } else if let Some(best) = vcs::find_best_remote(&git_repo)? {
+                debug!("Configured remote '{configured_remote}' not found, using: {best}");
+                best
+            } else {
+                bail!(
+                    "No remotes found in the git repository. \
+                         Use --repo and --default-branch to specify manually."
+                );
+            };
```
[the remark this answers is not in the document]
romtsn: Fixed. The git repo and remote resolution is now lazy — when `--repo` is provided, branch inference gracefully falls back to `main` if no git repo or remotes are available, instead of failing.

*— Claude Code*

src/commands/code_mappings/upload.rs
```diff
@@ -56,7 +59,70 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+            // Prefer explicit config (SENTRY_VCS_REMOTE / ini), then inspect
+            // the repo for the best remote (upstream > origin > first).
+            let config = Config::current();
+            let configured_remote = config.get_cached_vcs_remote();
+            let remote_name = if vcs::git_repo_remote_url(&git_repo, &configured_remote).is_ok() {
+                debug!("Using configured VCS remote: {configured_remote}");
+                configured_remote
+            } else if let Some(best) = vcs::find_best_remote(&git_repo)? {
+                debug!("Configured remote '{configured_remote}' not found, using: {best}");
+                best
```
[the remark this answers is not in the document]
romtsn: This is a false positive. The `get_cached_vcs_remote()` default of `"origin"` is the standard config-driven behavior used throughout the codebase (e.g. `set_commits.rs`, `build_vcs.rs`). Users can override it via `SENTRY_VCS_REMOTE` env var or ini config. `find_best_remote` is intentionally a fallback for when the configured remote doesn't exist, not the primary selection mechanism.

*— Claude Code*

src/commands/code_mappings/upload.rs
```diff
@@ -56,7 +59,95 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                    })?;
+                    let remote_name = remote_name.as_deref().ok_or_else(|| {
+                        anyhow::anyhow!(
+                            "No remotes found in the git repository. \
+                             Use --repo to specify manually."
+                        )
+                    })?;
+                    let remote_url = vcs::git_repo_remote_url(git_repo, remote_name)?;
+                    debug!("Found remote '{remote_name}': {remote_url}");
+                    let inferred = vcs::get_repo_from_remote(&remote_url);
```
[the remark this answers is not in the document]
romtsn: Good catch. Switched to `get_repo_from_remote_preserve_case` to avoid lowercasing the repo name.

*— Claude Code*

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,95 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+                            "main".to_owned()
+                        });
+                    println!("Inferred default branch: {inferred}");
+                    inferred
+                }
+            };
+
+            (repo_name, default_branch)
+        }
+    };
```
szokeasaurusrex: m: This logic is quite difficult for me to follow due to the deep nesting, and I suspect we can greatly simplify it.

As written, the match statement is coupling the values of `explicit_repo` and `explicit_branch`, behaving differently when they are both defined versus when only one of them is defined. I guess this is done to avoid needing to open the repo with `git2::Repository::open_from_env()` more than once, and to also avoid opening it altogether when we don't need to open it, but I think the complexity of avoiding this is not worth it.

Conceptually, in Rust-pseudocode, I think this is best written as:

```rust
let repo_name = matches.get_one::<String>("repo").unwrap_or_else(|| todo!("compute fallback"));
let default_branch = matches.get_one::<String>("default_branch").unwrap_or_else(|| todo!("compute fallback"));
```

l: Additionally, we might be able to extract some of this logic to utility functions.

szokeasaurusrex: Happy to help here if needed 🙏

romtsn: Refactored as suggested — flattened into two independent lookups with `resolve_git_remote`, `infer_repo_name`, and `infer_default_branch` extracted as helper functions.

*— Claude Code*

romtsn: clankers doing their job 😅

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,89 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        r.to_owned()
+    } else {
+        infer_repo_name(git_repo.as_ref(), remote_name.as_deref())?
+    };
+
+    let default_branch = if let Some(b) = explicit_branch {
+        b.to_owned()
+    } else {
+        infer_default_branch(git_repo.as_ref(), remote_name.as_deref())
+    };
```
loewenheim: This logic strikes me as a bit weird. If e.g. `explicit_repo` is `Some("repo_name")` but `explicit_branch` is `None`, on line 81 we would call `infer_default_branch` with `git_repo` (the repo inferred by `git2`), despite the fact that an explicit repo was given. Is that intentional?

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,89 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    let remote_name = remote_name.ok_or_else(|| {
+        anyhow::anyhow!("No remotes found in the git repository. Use --repo to specify manually.")
+    })?;
+    let remote_url = vcs::git_repo_remote_url(git_repo, remote_name)?;
+    debug!("Found remote '{remote_name}': {remote_url}");
+    let inferred = vcs::get_repo_from_remote_preserve_case(&remote_url);
+    if inferred.is_empty() {
+        bail!("Could not parse repository name from remote URL: {remote_url}");
+    }
+    println!("Inferred repository: {inferred}");
```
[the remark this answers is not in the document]
loewenheim: I believe this is correct—the `println`s in the `infer` functions should be replaced by debug log lines or removed entirely.

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,299 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        explicit_branch.map(|s| s.as_str()),
+        git_repo.as_ref(),
+    )?;
+
     println!("Found {} code mapping(s) in {path}", mappings.len());
+    println!("Repository: {repo_name}");
+    println!("Default branch: {default_branch}");
 
     Ok(())
 }
```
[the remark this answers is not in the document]
romtsn: this will be done in the next PR as mentioned in the description

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,298 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+
+    let (repo_name, default_branch) = resolve_repo_and_branch(
+        explicit_repo.map(|s| s.as_str()),
+        explicit_branch.map(|s| s.as_str()),
+        git_repo.as_ref(),
+    )?;
+
     println!("Found {} code mapping(s) in {path}", mappings.len());
+    println!("Repository: {repo_name}");
+    println!("Default branch: {default_branch}");
```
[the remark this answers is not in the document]
romtsn: this will be gone in the follow up PRs

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,298 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
 }
+
+/// Resolves the repository name and default branch from explicit args and git inference.
+fn resolve_repo_and_branch(
+    explicit_repo: Option<&str>,
+    explicit_branch: Option<&str>,
+    git_repo: Option<&git2::Repository>,
+) -> Result<(String, String)> {
+    let (repo_name, remote_name) = if let Some(r) = explicit_repo {
+        let remote = git_repo.and_then(|repo| find_remote_for_repo(repo, r));
```
loewenheim, line 92: Here you're calling the `find_remote_for_repo` function with both `git_repo` and `explicit_repo` (if they are both `Some`). What is the intended meaning of this? The `remote` only seems to be used in case `explicit_branch` is `None`. So the relevant situation here is that we have an `explicit_repo` but no `explicit_branch`. In that case, would it make sense to just default the branch to `main`?

romtsn: The idea is that if the user has e.g. an `upstream` remote pointing to `MyOrg/MyRepo` with default branch `develop`, and passes `--repo MyOrg/MyRepo`, we can give a more accurate default than `main`, so kind of best-effort and then still fall back to `main`. Does that make sense?

loewenheim: But the upstream is determined from the `git_repo`, which is obtained from the `env`. I guess my question is why we do this even if a repo name is passed manually as a parameter. Are they intended to be the same repo?

romtsn: as discussed internally, this is an additional validation layer to check that the remote URL indeed matches the passed `--repo` name.

src/commands/code_mappings/upload.rs
```diff
@@ -57,7 +60,293 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    }
+}
+
+/// Finds the remote whose URL matches the given repository name (e.g. "owner/repo").
+fn find_remote_for_repo(repo: &git2::Repository, repo_name: &str) -> Option<String> {
+    let remotes = repo.remotes().ok()?;
+    let found = remotes.iter().flatten().find(|name| {
+        vcs::git_repo_remote_url(repo, name)
+            .map(|url| vcs::get_repo_from_remote_preserve_case(&url) == repo_name)
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
romtsn: we don't need it here, the API will validate and reject if it doesn't match a configured repo on the Sentry side

PR #3209, opened 2026-03-12
feat(code-mappings): Wire up API integration for code-mappings upload

romtsn:
_#skip-changelog_

Connect the `code-mappings upload` command to the bulk code mappings API
endpoint (`POST /api/0/organizations/{org}/code-mappings/bulk/`).

Adds:
- `bulk_upload_code_mappings()` method on `AuthenticatedApi`
- Request/response data types in `src/api/data_types/code_mappings.rs`
- Summary table and error reporting in the command output
- Happy-path integration test with mock endpoint

Stack: #3207 → #3208 → **#3209** → #3210

Backend PRs: getsentry/sentry#109783, getsentry/sentry#109785, getsentry/sentry#109786

Closes getsentry/sentry-android-gradle-plugin#1079

### DIFF
--- a/src/api/data_types/code_mappings.rs
+++ b/src/api/data_types/code_mappings.rs
@@ -4,11 +4,11 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct BulkCodeMappingsRequest {
-    pub project: String,
-    pub repository: String,
-    pub default_branch: String,
-    pub mappings: Vec<BulkCodeMapping>,
+pub struct BulkCodeMappingsRequest<'a> {
+    pub project: &'a str,
+    pub repository: &'a str,
+    pub default_branch: &'a str,
+    pub mappings: &'a [BulkCodeMapping],
 }
 
 #[derive(Debug, Deserialize, Serialize)]
--- a/src/commands/code_mappings/upload.rs
+++ b/src/commands/code_mappings/upload.rs
@@ -73,10 +73,10 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
 
     let mapping_count = mappings.len();
     let request = BulkCodeMappingsRequest {
-        project,
-        repository: repo_name,
-        default_branch,
-        mappings,
+        project: &project,
+        repository: &repo_name,
+        default_branch: &default_branch,
+        mappings: &mappings,
     };
 
     println!("Uploading {mapping_count} code mapping(s)...");
@@ -194,6 +194,30 @@ fn infer_default_branch(git_repo: Option<&git2::Repository>, remote_name: Option
             debug!("No git repo or remote available, falling back to 'main'");
             "main".to_owned()
         })
+}
+
+fn print_results_table(mappings: Vec<BulkCodeMappingResult>) {
+    let mut table = Table::new();
+    table
+        .title_row()
+        .add("Stack Root")
+        .add("Source Root")
+        .add("Status");
+
+    for result in mappings {
+        let status = match result.detail {
+            Some(detail) if result.status == "error" => format!("error: {detail}"),
+            _ => result.status,
+        };
+        table
+            .add_row()
+            .add(&result.stack_root)
+            .add(&result.source_root)
+            .add(&status);
+    }
+
+    table.print();
+    println!();
 }
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


src/commands/code_mappings/upload.rs
```diff
@@ -120,9 +117,57 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         }
     };
 
-    println!("Found {} code mapping(s) in {path}", mappings.len());
-    println!("Repository: {repo_name}");
-    println!("Default branch: {default_branch}");
+    let mapping_count = mappings.len();
```
szokeasaurusrex, line 74: l: I guess we are not validating that this is under 300, because the follow-up PR handles bulk uploads

src/commands/code_mappings/upload.rs
```diff
@@ -120,9 +117,57 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        .add("Status");
+
+    for result in &response.mappings {
+        let status = match result.status.as_str() {
+            "error" => match &result.detail {
+                Some(detail) => format!("error: {detail}"),
+                None => "error".to_owned(),
+            },
+            s => s.to_owned(),
+        };
```
szokeasaurusrex: l: Two minor simplifications:

1. You can iterate directly over the mappings, to avoid needing to deal with `&str` references and owned `Strings` in the match.
2. You can also "invert" the match so that there are only two cases instead of three.

```suggestion
    for result in response.mappings {
        let status = match result.detail {
            Some(detail) if result.status == "error" => format!("error {detail}"),
            _ => result.status,
        };
```

src/commands/code_mappings/upload.rs
```diff
@@ -120,9 +117,57 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        table
+            .add_row()
+            .add(&result.stack_root)
+            .add(&result.source_root)
+            .add(&status);
+    }
```
szokeasaurusrex: l: You could consider extracting the logic to build the table to its own function.

To avoid ownership problems if you also apply my previous suggestion, make sure to only pass `response.mappings`, and not the entire `response` struct, to the function.

src/api/mod.rs
```diff
@@ -978,6 +978,17 @@ impl AuthenticatedApi<'_> {
...
 
+    /// Bulk uploads code mappings for an organization.
+    pub fn bulk_upload_code_mappings(
+        &self,
+        org: &str,
+        body: &BulkCodeMappingsRequest,
+    ) -> ApiResult<BulkCodeMappingsResponse> {
+        let path = format!("/organizations/{}/code-mappings/bulk/", PathArg(org));
+        self.post(&path, body)?
+            .convert_rnf(ApiErrorKind::ResourceNotFound)
```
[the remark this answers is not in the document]
romtsn: Fixed — switched to `ApiErrorKind::OrganizationNotFound` to match other org-scoped endpoints.

*— Claude Code*

src/commands/code_mappings/upload.rs
romtsn: Skipping the limit validation here since the follow-up PR (#3210) handles batching into 300-item chunks automatically.

*— Claude Code*

src/commands/code_mappings/upload.rs
romtsn: Applied both simplifications — iterating directly over `mappings` and inverting the match to two cases.

*— Claude Code*

src/commands/code_mappings/upload.rs
romtsn: Extracted the table building into a `print_results_table` function, taking `Vec<BulkCodeMappingResult>` to avoid ownership issues.

*— Claude Code*

src/api/data_types/code_mappings.rs
```diff
@@ -0,0 +1,37 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BulkCodeMappingsRequest {
+    pub project: String,
+    pub repository: String,
+    pub default_branch: String,
+    pub mappings: Vec<BulkCodeMapping>,
+}
```
szokeasaurusrex, line 12: m: Because of how we use this in #3210, I think this struct should probably use borrowed, instead of owned, types. This would avoid cloning in #3210 

```rust
pub struct BulkCodeMappingsRequest<'a> {
    pub project: &'a str,
    pub repository: &'a str,
    pub default_branch: &'a str,
    pub mappings: &'a [BulkCodeMapping],
}
```

romtsn: done in https://github.com/getsentry/sentry-cli/pull/3209/commits/bbd9516ae12d3df7dd60825c9a5d7e8818257b03

PR #3210, opened 2026-03-12
feat(code-mappings): Add batch splitting for large uploads

romtsn:
_#skip-changelog_

Split large mapping files into batches of 300 (the backend limit) per request.
Each batch is sent sequentially with progress reporting, and results are merged
into a single summary.

Also changes the output table to only show error rows — for large uploads
(hundreds of mappings), printing every row would flood the terminal. Successful
mappings are reflected in the summary counts instead.

Stack: #3207 → #3208 → #3209 → **#3210**

Backend PRs: getsentry/sentry#109783, getsentry/sentry#109785, getsentry/sentry#109786


szokeasaurusrex:
Please check and resolve the open comments from the AI reviewers, before I do a full review. If any comments are inaccurate or non-actionable, please indicate that with a 👎 and mark the comment as resolved. Thanks! 🙏

src/commands/code_mappings/upload.rs
```diff
@@ -118,56 +123,95 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
     );
 
-    if response.errors > 0 {
-        bail!(
-            "{} mapping(s) failed to upload. See errors above.",
-            response.errors
-        );
+    if merged.errors > 0 || !merged.batch_errors.is_empty() {
+        let total_errors = merged.errors + merged.batch_errors.len() as u64;
+        bail!("{total_errors} error(s) during upload. See details above.");
```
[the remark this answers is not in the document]
romtsn: Good catch. The summary `Errors:` count now includes both per-mapping errors and batch-level failures, consistent with the bail message.

*— Claude Code*

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
 
     let mapping_count = mappings.len();
-    let request = BulkCodeMappingsRequest {
-        project,
-        repository: repo_name,
-        default_branch,
-        mappings,
-    };
+    let batches: Vec<&[BulkCodeMapping]> = mappings.chunks(BATCH_SIZE).collect();
+    let total_batches = batches.len();
```
szokeasaurusrex: l: You can avoid allocating a `Vec` for the chunks pretty easily. Here, you compute the total batches using [`div_ceil`](https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil), and in the loop, you iterate the batches directly, computing them lazily on each iteration.


```suggestion
    let total_batches = mapping_count.div_ceil(BATCH_SIZE);
```

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
 
     let api = Api::current();
-    let response = api
-        .authenticated()?
-        .bulk_upload_code_mappings(&org, &request)?;
+    let authenticated = api.authenticated()?;
+
+    let mut merged = MergedResponse::default();
+
+    for (i, batch) in batches.iter().enumerate() {
```
szokeasaurusrex: l: [companion comment](https://github.com/getsentry/sentry-cli/pull/3210/changes#r2959390676)

```suggestion
    for (i, batch) in mappings.chunks(BATCH_SIZE).enumerate() {
```

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+        };
+        match authenticated.bulk_upload_code_mappings(&org, &request) {
+            Ok(response) => merged.add(response),
+            Err(err) => {
+                merged
+                    .batch_errors
+                    .push(format!("Batch {}/{total_batches} failed: {err}", i + 1));
+            }
+        }
+    }
```
szokeasaurusrex: l: I think it is a bit cleaner if you separate the merging the response from actually making the bulk requests. You can do this by implementing `FromIterator<Result<BulkCodeMappingsResponse, String>> for MergedResponse` (see [companion comment](https://github.com/getsentry/sentry-cli/pull/3210/changes#r2959476108)) and then just `collect`-ing into the `MergedResponse`

```suggestion
    let merged: MergedResponse = mappings
        .chunks(BATCH_SIZE)
        .enumerate()
        .map(|(i, batch)| {
            if total_batches > 1 {
                println!("Sending batch {}/{total_batches}...", i + 1);
            }
            let request = BulkCodeMappingsRequest {
                project: project.clone(),
                repository: repo_name.clone(),
                default_branch: default_branch.clone(),
                mappings: batch.to_vec(),
            };
            authenticated
                .bulk_upload_code_mappings(&org, &request)
                .map_err(|err| format!("Batch {}/{total_batches} failed: {err}", i + 1))
        })
        .collect();
```

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+}
+
+impl MergedResponse {
+    fn add(&mut self, response: BulkCodeMappingsResponse) {
+        self.created += response.created;
+        self.updated += response.updated;
+        self.errors += response.errors;
+        self.mappings.extend(response.mappings);
+    }
+}
```
szokeasaurusrex: l: By implementing `FromIterator`, you can just `.collect` directly to the `MergedResponse`. [See here](https://github.com/getsentry/sentry-cli/pull/3210/changes#r2959466365).

```suggestion
impl FromIterator<Result<BulkCodeMappingsResponse, String>> for MergedResponse {
    fn from_iter<I>(iter: I) -> Self
    where
        I: IntoIterator<Item = Result<BulkCodeMappingsResponse, String>>,
    {
        let mut merged = Self::default();
        for result in iter {
            match result {
                Ok(response) => {
                    merged.created += response.created;
                    merged.updated += response.updated;
                    merged.errors += response.errors;
                    merged.mappings.extend(response.mappings);
                }
                Err(err) => merged.batch_errors.push(err),
            }
        }
        merged
    }
}
```

tests/integration/code_mappings/upload.rs
```diff
@@ -10,3 +12,84 @@ fn command_code_mappings_upload() {
...
+    for i in 0..301 {
+        mappings.push(serde_json::json!({
+            "stackRoot": format!("com/example/m{i}"),
+            "sourceRoot": format!("modules/m{i}/src/main/java/com/example/m{i}"),
+        }));
+    }
+    let fixture = tempfile::NamedTempFile::new().expect("failed to create temp file");
+    serde_json::to_writer(&fixture, &mappings).expect("failed to write fixture");
+
+    let call_count = AtomicU16::new(0);
```
szokeasaurusrex: l: 8 bits is enough

```suggestion
    let call_count = AtomicU8::new(0);
```

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
         .add("Source Root")
-        .add("Status");
+        .add("Detail");
 
-    for result in mappings {
-        let status = match result.detail {
-            Some(detail) if result.status == "error" => format!("error: {detail}"),
-            _ => result.status,
-        };
+    for result in &error_mappings {
```
szokeasaurusrex: l: We can avoid allocating a vector for the `error_mappings` by directly iterating over the filter

```suggestion
    if !mappings.iter().any(|r| r.status == "error") {
        return;
    }

    let mut table = Table::new();
    table
        .title_row()
        .add("Stack Root")
        .add("Source Root")
        .add("Detail");

    for result in mappings.iter().filter(|r| r.status == "error") {
```

src/commands/code_mappings/upload.rs
```diff
@@ -144,56 +149,97 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    for (i, batch) in batches.iter().enumerate() {
+        if total_batches > 1 {
+            println!("Sending batch {}/{total_batches}...", i + 1);
+        }
+        let request = BulkCodeMappingsRequest {
+            project: project.clone(),
+            repository: repo_name.clone(),
+            default_branch: default_branch.clone(),
+            mappings: batch.to_vec(),
+        };
```
szokeasaurusrex: l: Please see https://github.com/getsentry/sentry-cli/pull/3209/changes#r2959847064; we can and probably should avoid cloning these strings and the vector.

szokeasaurusrex:
Ah also @romtsn, please add a changelog entry, either in this PR or a follow-up 🙏

romtsn:
addressed PR review in https://github.com/getsentry/sentry-cli/pull/3210/commits/38e43fa35298e0b6e23d5d863538fa27d06f048a + added a changelog entry

Issue #2919, opened 2025-11-07
build upload fails on XCode Cloud with .xarchive path


chdsbd:
### CLI Version

2.58.0

### Operating System and Architecture

- [ ] macOS (arm64)
- [x] macOS (x86_64)
- [ ] Linux (i686)
- [ ] Linux (x86_64)
- [ ] Linux (armv7)
- [ ] Linux (aarch64)
- [ ] Windows (i686)
- [ ] Windows (x86_64)

### Operating System Version

macOS Tahoe 26.0.1 (25A362)

### CLI Command

sentry-cli build upload $CI_ARCHIVE_PATH --build-configuration Release

### Exact Reproduction Steps

In my Xcode Cloud ci_scripts/ci_post_xcodebuild.sh file, I try to upload my app like the following:

```
sentry-cli build upload $CI_ARCHIVE_PATH --build-configuration Release
```

The `CI_ARCHIVE_PATH` variable is provided by Xcode Cloud and evaluates to `Volumes/workspace/build.xcarchive`

https://docs.sentry.io/platforms/apple/guides/ios/size-analysis/

### Expected Results

I expect the `.xarchive` to upload

### Actual Results

The command errors:

```
error: File is not a recognized supported build format (APK, or AAB): /Volumes/workspace/build.xcarchive
```

### Logs

```
  DEBUG   2025-11-06 19:36:42.474910 -08:00 Starting upload for 1 paths
  DEBUG   2025-11-06 19:36:42.474931 -08:00 Processing artifact at path: /Volumes/workspace/build.xcarchive
  DEBUG   2025-11-06 19:36:42.474994 -08:00 Loaded file with 0 bytes
  DEBUG   2025-11-06 19:36:42.475006 -08:00 Validating build format for: /Volumes/workspace/build.xcarchive
  DEBUG   2025-11-06 19:36:42.475016 -08:00 File format validation failed
error: File is not a recognized supported build format (APK, or AAB): /Volumes/workspace/build.xcarchive
```

runningcode:
Thanks for reporting this.
Two questions to help us debug:

How did you install `sentry-cli`?
Do you know if you installed the `x86-64` or the `arm64` version?

chdsbd:
@runningcode I used this script to install sentry-cli

```
curl -sL https://sentry.io/get-cli/ | bash
```

It looks like the x86 version was installed:

```
[...]

PR #3211, opened 2026-03-12
fix(build): Improve error for xcarchive/IPA uploads on non-ARM64

runningcode:
## Summary
- When users upload `.xcarchive` or `.ipa` files from a non-Apple Silicon Mac (e.g. Intel-based Xcode Cloud), the error now clearly states the Apple Silicon requirement instead of showing a confusing message that only lists Android formats (APK, AAB).

Fixes EME-951

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -209,6 +209,20 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
             return Err(anyhow!("Path does not exist: {}", path.display()));
         }
 
+        // On non-Apple Silicon, reject xcarchive/IPA early before trying to
+        // open the path as a file (xcarchive is a directory, so ByteView::open
+        // would fail with a confusing I/O error).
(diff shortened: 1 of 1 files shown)


src/commands/build/upload.rs
```diff
@@ -301,6 +542,18 @@ fn validate_is_supported_build(path: &Path, bytes: &[u8]) -> Result<()> {
     }
 
     debug!("File format validation failed");
+
+    #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
```
szokeasaurusrex: l: I was thinking it might make more sense to have this error only on `x86_64` macOS builds, given that the users in #2919 were only confused about the lack of support on Intel Macs, and Xcode Cloud specifically.

runningcode: Yes, that's a good point but this error would also occur if somehow a user tried to upload an xcarchive on a linux or other machine.

PR #3212, opened 2026-03-12
ci(build): Add snapshot release workflow

runningcode:
## Summary

Adds a `workflow_dispatch`-triggered snapshot release pipeline that reuses the existing build workflow, enabling on-demand pre-release builds from any branch. For now we are only publishing to npm under the snapshot tag.

## Background

We want to allow easy snapshot releases to ease internal testing as well as customer testing. 

### What it does

- **New `snapshot.yml` workflow**: Computes a semver-compliant snapshot version (e.g. `3.4.0-snapshot.20260312.abc1234`) by bumping the minor version from `Cargo.toml`, patches all version files via `scripts/bump-version.sh`, pushes a temporary snapshot branch, then triggers the build
- **Makes `build.yml` callable**: Adds `workflow_call` inputs (`skip-signing`, `is-snapshot`, `checkout-ref`) so the snapshot workflow can reuse the full build matrix
- **Skips unnecessary steps for snapshots**: macOS code signing/notarization, Python package builds, Docker image builds, and the release merge job are all conditionally skipped
- **Publishes to npm under `snapshot` tag**: Platform-specific binary packages and the main `@sentry/cli` package are published with `--tag snapshot` so they don't affect the `latest` tag
- **Cleans up after itself**: The temporary snapshot branch is deleted after publishing (or on failure)

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -1,7 +1,8 @@
 name: Snapshot Release
 
 on:
-  workflow_dispatch:
+  push:
+    branches: [master]
 
 permissions:
   contents: write


.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          npm publish node-package/*.tgz --tag snapshot
+
+  cleanup:
+    name: Cleanup
+    needs: [prepare, publish-npm]
+    if: always()
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Delete snapshot branch
```
runningcode, line 112: optionally, we can also keep these snapshot branches, but i don't think they will be useful.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,114 @@
+name: Snapshot Release
+
+on:
```
runningcode, line 3: manual trigger to test. we can change this to build on every push to master once we know it is working.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,114 @@
...
+          VERSION="${MAJOR}.${NEXT_MINOR}.0-snapshot.${DATE}.${SHORT_SHA}"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "current=$CURRENT" >> "$GITHUB_OUTPUT"
+          echo "Snapshot version: $VERSION"
+
+      - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # 6.2.0
+        with:
+          node-version: '20.10.0'
+
+      - name: Bump versions
```
runningcode, line 45: so another way to do this would be to modify the versions in place in the action/workflow using sed and regexes.

this would remove the need to commit and create the snapshot branch somewhere but then we have two different mechanisms to patch the versions which could lead to drift. without modifying in place, there's no way to share the patched repo between workflows.

.github/workflows/build.yml
```diff
@@ -366,6 +401,7 @@ jobs:
           if-no-files-found: 'error'
 
   platform-specific-docker:
+    if: ${{ !inputs.is-snapshot }}
```
runningcode, line 395: skip docker publishing

.github/workflows/build.yml
```diff
@@ -297,11 +327,14 @@ jobs:
           if-no-files-found: 'error'
 
   python:
```
runningcode, line 320: not publishing to pip

.github/workflows/build.yml
```diff
@@ -179,6 +199,7 @@ jobs:
           name: unsigned-bin-macos-${{ matrix.arch }}
 
       - name: Sign binary
```
runningcode, line 195: no signing for snapshots

.github/workflows/build.yml
```diff
@@ -4,6 +4,17 @@ on:
...
       - release/**
+  workflow_call:
+    inputs:
+      skip-signing:
+        type: boolean
+        default: false
+      is-snapshot:
+        type: boolean
+        default: false
+      checkout-ref:
```
runningcode, line 12: we need this checkout-ref to share the version patched repo between workflows. see comment here for more details: https://github.com/getsentry/sentry-cli/pull/3212/changes#r2924915514

.github/workflows/build.yml
```diff
@@ -366,6 +401,7 @@ jobs:
           if-no-files-found: 'error'
 
   platform-specific-docker:
+    if: ${{ !inputs.is-snapshot }}
     name: Build Docker Image (${{ matrix.platform }})
     strategy:
       matrix:
```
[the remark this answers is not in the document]
runningcode: This is real but this was already an issue on master.

szokeasaurusrex:
Hey @runningcode, please check the AI review comments and resolve them, or mark as inapplicable, then I'll do a full review 🙏

.github/workflows/build.yml
```diff
@@ -188,17 +209,19 @@ jobs:
...
           rcodesign notary-submit \
             --api-key-file ${{ env.APPLE_API_KEY_PATH }} \
             --wait \
             sentry-cli-Darwin-${{ matrix.arch }}.zip
 
-      - name: Upload signed binary
+      - name: Upload binary
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
         with:
           name: artifact-bin-macos-${{ matrix.arch }}
```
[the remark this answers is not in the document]
runningcode: There is no `if: ${{ !inputs.skip-signing }}` for the upload step of the `sign-macos-binaries` job. This is false.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,114 @@
...
+    name: Cleanup
+    needs: [prepare, publish-npm]
+    if: always()
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Delete snapshot branch
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh api -X DELETE "repos/${{ github.repository }}/git/refs/heads/${{ needs.prepare.outputs.ref }}" || true
```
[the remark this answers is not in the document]
runningcode: I'm open to other thoughts but I think we don't want the job to fail if we fail to delete the temporary branch here.

runningcode:
Done. One is false, the other is a design choice. I commented on both!

.github/workflows/build.yml
```diff
@@ -68,6 +81,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ inputs.checkout-ref }}
```
szokeasaurusrex, line 82: m: What happens when `inputs.checkout-ref` is `''` (the default value) here? Is it equivalent to omitting the `with: ref:`?

h: In any case, please push this branch to `release/test` (or any other branch starting with `release/`) to trigger the "Release Build" workflow. I would like to make sure that these changes don't break the workflow for normal releases.

.github/workflows/build.yml
szokeasaurusrex: m: Why not sign the snapshot builds? It doesn't take that long to sign the binaries (I think about a minute or so typically), so I see little downside to it. 

If we want external users to be able to test these snapshot builds, it could be helpful to have the binaries be signed, as their systems may be configured to only allow running signed executables.

.github/workflows/build.yml
```diff
@@ -247,10 +272,12 @@ jobs:
   node:
     name: NPM Package
     runs-on: ubuntu-24.04
-    needs: [linux, macos, macos_universal, windows]
+    needs: [linux, sign-macos-binaries, windows]
```
szokeasaurusrex, line 266: m: This seems like an unrelated, but valid change. I think we should split to a separate PR (I'll even prepare it, I am surprised this has worked so far)

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      # Computes a semver-compliant snapshot version based on the current
+      # version in Cargo.toml. The minor version is bumped so that the
+      # snapshot sorts higher than the current release but lower than the
+      # next real release. For example, if Cargo.toml has 3.3.1, the
+      # snapshot version will be 3.4.0-snapshot.20260312.abc1234.
+      - name: Compute snapshot version
+        id: version
+        run: |
+          CURRENT=$(grep '^version = ' Cargo.toml | head -1 | sed 's/version = "\(.*\)"/\1/')
```
szokeasaurusrex: m: The current approach is heuristic-based; a more reliable approach would be to parse the `cargo metadata` like so:

```suggestion
          CURRENT=$(cargo metadata --no-deps --format-version 1 \
            | jq -er '(.workspace_default_members[0]) as $id | .packages[] | select(.id == $id) | .version')```

File: .github/workflows/snapshot.yml [hunk 14]
```diff
@@ -0,0 +1,114 @@
...
+      version: ${{ steps.version.outputs.version }}
+      ref: ${{ steps.push.outputs.ref }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+
+      # Computes a semver-compliant snapshot version based on the current
+      # version in Cargo.toml. The minor version is bumped so that the
+      # snapshot sorts higher than the current release but lower than the
+      # next real release. For example, if Cargo.toml has 3.3.1, the
+      # snapshot version will be 3.4.0-snapshot.20260312.abc1234.
```
szokeasaurusrex: m: Why bump the minor? I think we should bump the patch, otherwise, if the next real release is a patch, then the older snapshots will still sort above the patch.

Unless there is a good reason to bump the minor, I would rather always bump the patch for the snapshots, as it is the smallest possible increase in a new release.

.github/workflows/snapshot.yml
szokeasaurusrex: No, they should get deleted I believe

.github/workflows/snapshot.yml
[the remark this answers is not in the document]
szokeasaurusrex: l: I think checking `needs.prepare.outputs.ref != ''` makes sense. 

Also, I disagree about not wanting the job to fail if we fail to delete the temporary branch. I think it would be good to know if we are failing to delete these branches so we can investigate, and the failure I don't think should block anything. But, I don't feel super strongly about this.

.github/workflows/build.yml
runningcode: Yeah, it was something that seer or cursor noticed and was legit so I fixed it here.

szokeasaurusrex: (totally forgot to make that split PR, doing it now!)

szokeasaurusrex: Split to #3222

.github/workflows/snapshot.yml
runningcode: will do!

.github/workflows/build.yml
runningcode: Good point, there's really no reason not to sign them.

.github/workflows/build.yml
runningcode: I pushed it to `release/snapshot-test` and workflow succeeded here: https://github.com/getsentry/sentry-cli/actions/runs/23295966917

.github/workflows/build.yml
```diff
@@ -198,7 +212,7 @@ jobs:
             --wait \
             sentry-cli-Darwin-${{ matrix.arch }}.zip
 
-      - name: Upload signed binary
+      - name: Upload binary
```
szokeasaurusrex: l: This change seems out of scope

```suggestion
      - name: Upload signed binary
```

.github/workflows/snapshot.yml
szokeasaurusrex: l: I would be fine with just having it set to every push to master with this PR. We can revert if needed

PR #3238, opened 2026-03-24
fix: gate goblin logs before formatting

supervacuus:
### Description

As evaluated in the context of https://github.com/getsentry/sentry-cli/pull/3237, in some configurations, just formatting the `goblin` logs can be costly enough to become a performance issue.

This PR moves the logging gate for `goblin` into `enabled()` so that we discard its logs before we format them.

### Issues

Raised as part of the fixes for https://github.com/getsentry/sentry/issues/104738

### Legal Boilerplate

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,6 +13,7 @@
 
 # Files co-owned by Emerge Tools team
 /apple-catalog-parsing @getsentry/emerge-tools @getsentry/owners-sentry-cli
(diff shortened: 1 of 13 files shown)


supervacuus:
@loewenheim, thx! Since Danger complains, does this justify an entry into the changelog?

loewenheim:
> @loewenheim, thx! Since Danger complains, does this justify an entry into the changelog?

No, I don't think so. I applied the label.

PR #3239, opened 2026-03-24
test: make tests independent of zip encoding

supervacuus:
### Description

Extracted from https://github.com/getsentry/sentry-cli/pull/3237

This change was triggered by bumping `symbolic` to `12.17.3` to include https://github.com/getsentry/symbolic/pull/960.

`symbolic` updated its `zip` dependency (`2.4.2` to `7.2.0`) since the last bump. This changed encoding internals and invalidated a bunch of test assertions.

The PR addresses the issue in the following way:

I rewrote the affected tests so that they decode the incoming chunks and compare their contents against the fixtures, rather than snapshotting, which could break with every change to the encoder. I also simplified the usage of `split_chunk_body()` a bit: now it returns a `Vec`, and callers decide how they want to view the data: the simple tests collect into a `HashSet` for set comparison, while the small-chunk tests use a `SHA1` digest multimap to verify exact chunk identity and multiplicity.

If you actually wanted an early warning whether the encoding was stable, then we can drop this PR. 

Specifically, for `build_deterministic()` in `src/utils/source_bundle.rs`, it was not entirely clear whether you want two runs of the same version to have predictably the same output (which I changed it to) or whether it should actually be stable across future versions (which a `zip` crate bump upstream would break).

### Issues

Raised as part of the fix proposals for https://github.com/getsentry/sentry/issues/104738

### Legal Boilerplate

### DIFF
--- a/tests/integration/test_utils/chunk_upload.rs
+++ b/tests/integration/test_utils/chunk_upload.rs
@@ -155,6 +155,9 @@ where
     counts
 }
 
+// These helpers and tests exercise the multipart parsing utilities in this
+// module itself. They are tests for test code, not integration tests for
+// sentry-cli behavior.
 fn gzip(data: &[u8]) -> Vec<u8> {
     let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
     encoder


szokeasaurusrex:
I'm a bit more hesitant with this change. I think it is good if we know when the encoding changes, as it is important that we use a zip encoding which the Sentry backend (including older self-hosted versions) properly support. If the encoding ever changes, we should validate the server still recognizes the files properly, then update the snapshots.

What do you all think @supervacuus and @loewenheim?

loewenheim:
> I'm a bit more hesitant with this change. I think it is good if we know when the encoding changes, as it is important that we use a zip encoding which the Sentry backend (including older self-hosted versions) properly support. If the encoding ever changes, we should validate the server still recognizes the files properly, then update the snapshots.
> 
> What do you all think @supervacuus and @loewenheim?

I'm not sure it's possible for the zip encoding to become genuinely incompatible. As far as I understand the format is fixed. But that doesn't mean that a change in the encoder can't lead to a file being compressed in a different (but equally valid) way.

szokeasaurusrex:
As far as I am aware, the ZIP format can use other compression algorithms (e.g. ZSTD to name one in particular) which are not universally supported. I'd be concerned about missing a change in the compression algorithm used, which could break compatibility with Sentry SaaS or self-hosted

supervacuus:
> I'm not sure it's possible for the zip encoding to become genuinely incompatible. As far as I understand the format is fixed. But that doesn't mean that a change in the encoder can't lead to a file being compressed in a different (but equally valid) way.

That is exactly the premise of my proposal. The test suite on `master` currently tests bit-equality _not_ compatibility. The encoder can be changed to produce a different binary output for a particular byte sequence, yet the output remains decodable by any decoder in the wild.

So, while it is true that a break here would signal a change in output, it does not indicate that any version of the backend would be negatively affected (because it does not tell us what changed). It is most likely a false positive. It would also have to be raised back upstream, which means early warnings would have to exist there, too. Even if you use the bit-level break as a signal for potential breakage, you'd still have to check the range of backend versions for compatibility (without blocking a zip dependency bump indefinitely).

I think a compatibility check that goes beyond bit-equality can be achieved either by introducing minimal end-to-end tests vs a particular set of pinned backend versions, or, likely more practical, by assuming a set of compression parameters inside the ZIP container that are part of the contract across backend versions. Something like

```rust
let mut archive =
    zip::ZipArchive::new(std::fs::File::open(bundle.path()).unwrap()).unwrap();
assert!(
    !archive.is_empty(),
    "bundle should contain at least one entry"
);
for i in 0..archive.len() {
    let entry = archive.by_index(i).unwrap();
    let method = entry.compression();
    assert!(
        matches!(
            method,
            zip::CompressionMethod::Stored | zip::CompressionMethod::Deflated
        ),
        "only assume store-as-is and deflate as compression methods",
    );
}
```
This allows you to create very tight, concrete bounds on the bundle ZIP container that you can tune (there are additional metadata items that you bound via tests, like "version made by" and [extra fields](https://libzip.org/specifications/extrafld.txt), which can be central to compatibility, but I think just a tight allowlist is a good start), which should also be easier to sync with backends, while keeping you isolated from trivial decoder output changes.

szokeasaurusrex:
Fair enough. Given that, I am okay with this approach generally 👍 <br><br>Will plan to take a look at this in more depth tomorrow

tests/integration/debug_files/upload.rs
```diff
@@ -407,20 +358,23 @@ fn chunk_upload_multiple_files_only_some() {
...
+
+                    // Only 2 of 3 files need uploading (one is already on the server).
+                    assert_eq!(decompressed.len(), 2, "expected exactly two chunks");
+                    let decompressed: std::collections::HashSet<_> =
+                        decompressed.into_iter().collect();
                     assert!(
-                        chunks == expected_chunks,
-                        "Uploaded chunks differ from the expected chunks"
+                        decompressed.is_subset(&all_fixtures),
+                        "uploaded chunks should be a subset of the fixture files"
```
szokeasaurusrex: m: If possible, we should check that the two chunks that get uploaded are the two we expect; otherwise, it is possible that we reuploaded the chunk that was already on the server

tests/integration/debug_files/upload.rs
```diff
@@ -407,20 +358,30 @@ fn chunk_upload_multiple_files_only_some() {
...
 
-                    // Using assert! because in case of failure, the output with assert_eq!
-                    // is too long to be useful.
-                    assert!(
-                        chunks == expected_chunks,
-                        "Uploaded chunks differ from the expected chunks"
+                    assert_eq!(
+                        HashSet::from_iter(decompressed),
+                        expected_uploads,
+                        "uploaded chunks should be exactly the two fixtures not already on the server",
```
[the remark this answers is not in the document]
supervacuus: I actually had that assert in there, but I dropped it because it was superfluous (in the sense that we shouldn't assert count and assert the equality of the data). However, I didn't change the comparison below and kept the collapsing `HashSet::from_iter()`.

So the right fix isn't an additional length assertion, but using a data structure that preserves multiplicity (a digest->count multiset).

tests/integration/debug_files/upload.rs
```diff
@@ -477,6 +438,59 @@ fn chunk_upload_multiple_files_only_some() {
...
+    let mut chunk_digests = BTreeMap::new();
+
+    for fixture_name in ["fibonacci", "fibonacci-fast", "main"] {
+        let fixture_data = fs::read(format!("{fixture_dir}/{fixture_name}"))
+            .expect("fixture file should be readable");
+        let file_checksum = sha1_hex(&fixture_data);
+        let file_response = assemble_response
+            .get(&file_checksum)
+            .expect("assemble response should contain an entry for every fixture file");
+
```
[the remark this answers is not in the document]
supervacuus: fair enough, fixed in https://github.com/getsentry/sentry-cli/pull/3239/commits/f1b2deb49ee284947d4fc903e9f60707f7d3cde3

tests/integration/debug_files/upload.rs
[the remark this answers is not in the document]
supervacuus: fixed in https://github.com/getsentry/sentry-cli/pull/3239/commits/7c2d9f09c8942eec050882b4c9607e86870ba64d for `chunk_upload_multiple_files()` and `chunk_upload_multiple_files_only_some()`

supervacuus:
> m: If possible, we should check that the two chunks that get uploaded are the two we expect; otherwise, it is possible that we reuploaded the chunk that was already on the server

Great catch!

`chunk_upload_multiple_files_only_some()` now asserts the exact expected chunk set. It derives the expectation by hashing the three fixtures, filtering out the one whose SHA1 matches the "already on the server" chunk from the `dif/assemble` mock, and then compares it against the decompressed request body using `chunk_upload::chunk_digest_counts()` (a `BTreeMap<String, usize>`).

Re-uploading the server-resident chunk (or duplicating either missing one) now fails the assertion. I applied the same multiset pattern to the sibling `chunk_upload_multiple_files()` test for consistency; all four `chunk_upload_multiple_files*` variants now share the same shape.

On the broader compatibility concern from the earlier thread: added a `build_compression_method_canary` in `src/utils/source_bundle.rs` that opens a freshly built bundle and asserts that each zip entry's compression method is in an allowlist of `{Stored, Deflated}`. It's a starting bound, not a verified server contract, of course. But it will fail loudly on a silent upstream switch to Zstd/Bzip2/etc, so we can consciously re-validate before accepting the bump.

szokeasaurusrex:
Excellent, thanks for all the work!

src/utils/source_bundle.rs
```diff
@@ -229,37 +257,35 @@ mod tests {
...
+            assert!(
+                matches!(
+                    method,
+                    zip::CompressionMethod::Stored | zip::CompressionMethod::Deflated
+                ),
+                "entry {:?} uses {method:?}, outside the current allowlist: \
+                 verify backend compatibility before widening",
+                entry.name(),
+            );
+        }
```
szokeasaurusrex, line 289: Nice 🚀

tests/integration/test_utils/chunk_upload.rs
```diff
@@ -105,3 +102,126 @@ fn entire_body_regex(regex_escaped_boundary: &str) -> Regex {
...
+        let mut sha = Sha1::new();
+        sha.update(chunk.as_ref());
+        *counts.entry(sha.digest().to_string()).or_default() += 1;
+    }
+
+    counts
+}
+
+#[cfg(test)]
+mod tests {
```
szokeasaurusrex: m: Why do we have a `mod tests` block here? This file is already inside an integration test, so the `#[cfg(test)]` on the module should have no effect AFAIK.

supervacuus: No particular reason. I am not a Rust native, and, as such, I assumed this was not a test module but a test utility, so I created a test sub-module for the tests 😅.

Flattened here: https://github.com/getsentry/sentry-cli/pull/3239/commits/9614ef2f4bc9b2b0177174b03fc9110e291b3077

szokeasaurusrex: l: Got it, so these are essentially tests for the test code. I would consider adding a code comment indicating that they are tests for test code on the respective test-of-test functions 😄

supervacuus: fair, added here: https://github.com/getsentry/sentry-cli/pull/3239/commits/f65f6afebbb2a553a3752019908590276c19453c

supervacuus:
Thanks for the approval. I cannot merge this, though, since the "Warden" check fails for me (it fails for me everywhere). So, feel free to merge whenever you have time.

szokeasaurusrex:
Will do! I also saw getsentry/sentry-cli#3240, can you please rebase that PR on top of this one, and ensure the tests pass? Once tests are passing on that PR, I'll be happy to review it as well

PR #3243, opened 2026-03-25
fix(inject): Skip non-base64 embedded sourcemaps

nyan-left:
### Description

When you bundle tools like terser or babel into a single JS file (common in browser-based IDEs or bundler workers), their minified code contains template literals that build `//# sourceMappingURL=data:application/json;base64,...` strings as output. After minification, these can land at the start of a line, and `discover_sourcemaps_location` treats them as real sourcemap directives.

The problem is the "base64" payload is actually just more JS code, so decoding fails and `sourcemaps inject` bails with `Invalid embedded sourcemap`, stopping the whole injection.

This changes the `bail!` to a `warn!` + `continue`, which matches how the injection flow already handles other non-fatal cases like missing sourcemaps, ambiguous paths, and unfound files. The file gets skipped and everything else processes normally.

Includes a test with a minimal fixture that reproduces the issue.

### Issues

https://github.com/getsentry/sentry-cli/issues/3244

### Legal Boilerplate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- (sourcemaps) Skip non-base64 embedded sourcemaps during injection ([#3243](https://github.com/getsentry/sentry-cli/pull/3243))
+
 ### New Features ✨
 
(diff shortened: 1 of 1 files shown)


loewenheim:
I took the liberty of adding a changelog entry.

nyan-left:
Thanks @loewenheim !  :)

PR #3258, opened 2026-04-09
feat(snapshots): Authenticate with Objectstore

### LINKED ISSUE #3159: Use Objectstore auth token instead of Sentry auth token
### Description

Change the implementation for the snapshots subcommand to use the Objectstore auth token instead of the Sentry auth token.
Prerequisites:
1. Remove Django auth from Objectstore endpoint
2. Mount secrets, access them from Sentry to mint JWTs, modify snapshots upload-options endpoint to provide JWT to sentry-cli
3. Change sentry-cli to receive the token and use it for auth, as opposed to the Sentry auth token

lcian:
Bump `objectstore-client` to 0.1.6 and use the dedicated auth token now returned by `ObjectstoreUploadOptions`.
`.token` in the newest version of Objectstore client passes it in the `X-Os-Auth` header.
The Sentry auth token is still passed through the standard `Authorization` header for auth with Django.

#skip-changelog

Close https://github.com/getsentry/sentry-cli/issues/3159
Close CLI-292

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,7 +79,7 @@ zip = "2.4.2"
 data-encoding = "2.3.3"
 magic_string = "0.3.4"
 chrono-tz = "0.8.4"
-secrecy = "0.8.0"
+secrecy = { version = "0.8.0", features = ["serde"] }
 lru = "0.16.3"
 backon = { version = "1.5.2", features = ["std", "std-blocking-sleep"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ java-properties = "2.0.0"
 lazy_static = "1.4.0"
 libc = "0.2.139"
 log = { version = "0.4.17", features = ["std"] }
-objectstore-client = { version = "0.1.4" , default-features = false, features = ["native-tls"] }
+objectstore-client = { version = "0.1.6" , default-features = false, features = ["native-tls"] }
 open = "3.2.0"
 parking_lot = "0.12.1"
 percent-encoding = "2.2.0"
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -336,6 +336,7 @@ fn upload_images(
     if let Some(token) = options.objectstore.auth_token {
         builder = builder.token(token.expose_secret().to_owned());
     }
+    let builder = builder;
 
     let sentry_token = match authenticated_api.auth() {
         Auth::Token(token) => token.raw().expose_secret().to_owned(),
--- a/src/commands/build/snapshots.rs
+++ b/src/commands/build/snapshots.rs
@@ -332,14 +332,25 @@ fn upload_images(
     let expiration = ExpirationPolicy::from_str(&options.objectstore.expiration_policy)
         .context("Failed to parse expiration policy from upload options")?;
 
-    let client = ClientBuilder::new(options.objectstore.url)
-        .token({
-            // TODO: replace with auth from `ObjectstoreUploadOptions` when appropriate
-            match authenticated_api.auth() {
-                Auth::Token(token) => token.raw().expose_secret().to_owned(),
-            }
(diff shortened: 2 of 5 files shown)


src/commands/build/snapshots.rs
```diff
@@ -295,15 +295,25 @@ fn upload_images(
...
-    let client = ClientBuilder::new(options.objectstore.url)
-        .token({
-            // TODO: replace with auth from `ObjectstoreUploadOptions` when appropriate
-            let auth = match authenticated_api.auth() {
-                Auth::Token(token) => token.raw().expose_secret().to_owned(),
-            };
-            auth
+    let mut builder = ClientBuilder::new(options.objectstore.url);
+    if let Some(token) = options.objectstore.auth_token {
+        builder = builder.token(token.expose_secret().to_owned());
```
szokeasaurusrex, line 337: l: Why not implement `From<Secret<String>>` for `TokenProvider`?

I think that probably reduces the scope where the secret string could accidentally get leaked in a panic.

src/commands/build/snapshots.rs
```diff
@@ -295,15 +295,25 @@ fn upload_images(
...
+    }
```
szokeasaurusrex, line 338: l: We can redeclare `builder` as immutable here (since we do not mutate any further)

```suggestion
    }
    let builder = builder;
```

src/commands/build/snapshots.rs
```diff
@@ -400,7 +410,7 @@ fn upload_images(
         warn!("Some images share identical file names. Only the first occurrence of each is included:{details}");
     }
 
-    let result = runtime.block_on(async { many_builder.send().error_for_failures().await });
+    let result = runtime.block_on(async { many_builder.send().await.error_for_failures().await });
```
szokeasaurusrex: l: How did this work before? 😂

Cargo.toml
```diff
@@ -79,7 +79,7 @@ zip = "2.4.2"
 data-encoding = "2.3.3"
 magic_string = "0.3.4"
 chrono-tz = "0.8.4"
-secrecy = "0.8.0"
+secrecy = { version = "0.8.0", features = ["serde"] }
```
szokeasaurusrex, line 82: m: Why do we now need the `serde` feature? Where are we using it?

szokeasaurusrex (commented):
Mostly looks good, have some questions and suggestions though

Cargo.toml
lcian: To define `pub auth_token: Option<SecretString>` in `ObjectstoreUploadOptions`.
Without `serde`, `SecretString` wouldn't be `Deserialize`, so we couldn't use it there.

src/commands/build/snapshots.rs
lcian: The new version of `objectstore-client` makes `ManyBuilder::send` async :)
That method now needs to be async for some internal stuff.
In practice, the requests are still lazily made only the moment when `error_for_failures.await` is called.

szokeasaurusrex: cool got it, makes sense

szokeasaurusrex (approved):

src/commands/build/snapshots.rs
lcian: As we discussed previously, internally the `TokenProvider` stores a newtype with a `Debug` implementation which redacts the key, so practically the effect is the same. 
I'm not opposed to implementing `From<Secret<String>>` for `TokenProvider` though, it just adds a single dep more to `objectstore-client`, I'll do it sometime soon.

szokeasaurusrex: You could place the `From` implementation behind a feature flag, that would make the `secrecy` crate an optional dependency

szokeasaurusrex: but this comment is nonblocking in any case

PR #3260, opened 2026-04-10
feat(bundle-jvm): Automatically collect JVM source files with respecting excludes and gitignore 

romtsn:
## Summary
- Filter collected sources by JVM file extensions (`java`, `kt`, `scala`, `groovy`) instead of collecting all files
- Exclude common build output directories by default (`build`, `.gradle`, `.cxx`, `node_modules`)
- Respect `.gitignore` / `.ignore` rules as an additional layer of defense
- Add `--exclude` CLI arg for user-provided glob patterns on top of defaults

This enables invoking `bundle-jvm` directly on a project root without needing to pre-copy sources into a separate directory (e.g. from the Android Gradle plugin's `CollectSourcesTask`).

This also does not break the existing plugins' workflows, only in subtle cases where a non-JVM file has been included into the `src/main/java` source set, but this is likely not supported on the product side either.

## Test plan
- [ ] Run `bundle-jvm` on an Android project root and verify only `.java`/`.kt` files are collected
- [ ] Verify `build/` directory contents are excluded
- [ ] Verify `.gitignore`d paths are respected
- [ ] Test `--exclude` with custom glob patterns
- [ ] Verify existing callers of `ReleaseFileSearch` are unaffected (default `respect_ignores: false`)

### DIFF
--- a/src/commands/debug_files/bundle_jvm.rs
+++ b/src/commands/debug_files/bundle_jvm.rs
@@ -8,11 +8,13 @@ use crate::utils::fs::path_as_url;
 use crate::utils::source_bundle::{self, BundleContext};
 use anyhow::{bail, Context as _, Result};
 use clap::{Arg, ArgAction, ArgMatches, Command};
+use log::debug;
 use sentry::types::DebugId;
 use std::borrow::Cow;
 use std::collections::BTreeMap;
+use std::ffi::OsStr;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr as _;
 use std::sync::Arc;
 use symbolic::debuginfo::sourcebundle::SourceFileType;
@@ -22,8 +24,9 @@ const JVM_EXTENSIONS: &[&str] = &[
     "java", "kt", "scala", "sc", "groovy", "gvy", "gy", "gsh", "clj", "cljc",
 ];
 
-/// Default directory patterns to exclude from source collection.
-const DEFAULT_EXCLUDES: &[&str] = &[
+/// Directory patterns that are always safe to exclude globally (can never be
+/// valid JVM package names due to leading dots or conventions).
+const SAFE_EXCLUDES: &[&str] = &[
     "!.cxx",
     "!.eclipse",
     "!.fleet",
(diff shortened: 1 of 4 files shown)


src/commands/debug_files/bundle_jvm.rs
```diff
@@ -16,6 +16,29 @@ use std::str::FromStr as _;
...
+    "!target",
+    "!.mvn",
+    "!.idea",
+    "!.vscode",
+    "!.eclipse",
+    "!.settings",
+    "!bin",
+    "!out",
+    "!.kotlin",
+    "!.fleet",
```
szokeasaurusrex: l: Let's sort this list alphabetically

src/commands/debug_files/bundle_jvm.rs
```diff
@@ -75,7 +109,22 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    let user_excludes: Vec<String> = matches
+        .get_many::<String>("exclude")
+        .map(|vals| vals.map(|v| format!("!{v}")).collect())
+        .unwrap_or_default();
+
+    let all_excludes: Vec<&str> = DEFAULT_EXCLUDES
+        .iter()
+        .copied()
+        .chain(user_excludes.iter().map(|s| s.as_str()))
+        .collect();
```
szokeasaurusrex: l: small simplification to avoid allocating two vectors (also, for the first let, you avoid the `.map(.map)` by turning the option into an iterator and flattening it)

```suggestion
    let user_excludes = matches
        .get_many::<String>("exclude")
        .into_iter()
        .flatten()
        .map(|v| format!("!{v}"));

    let all_excludes = DEFAULT_EXCLUDES
        .iter()
        .copied()
        .map(Cow::Borrowed)
        .chain(user_excludes.map(Cow::Owned));


```

szokeasaurusrex (approved):
Some small suggestions

src/commands/debug_files/bundle_jvm.rs
```diff
@@ -7,15 +7,70 @@ use crate::utils::file_upload::SourceFile;
...
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr as _;
 use std::sync::Arc;
 use symbolic::debuginfo::sourcebundle::SourceFileType;
 
+/// File extensions for JVM-based languages.
+const JVM_EXTENSIONS: &[&str] = &[
+    "java", "kt", "scala", "sc", "groovy", "gvy", "gy", "gsh", "clj", "cljc",
+];
```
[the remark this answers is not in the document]
romtsn: I'd keep sorting by order of importance here

szokeasaurusrex (approved):

PR #3268, opened 2026-04-14
feat(build): Add --selective flag to snapshots command

NicoHinderling:
Add `--selective` flag to `sentry-cli build snapshots` to indicate the upload
contains only a subset of images (e.g. due to sharding or selective test
execution).

The command validates that every uploaded image is listed in the provided
file, failing early if an unknown image is found.

### DIFF
--- a/tests/integration/_cases/build/build-snapshots-help.trycmd
+++ b/tests/integration/_cases/build/build-snapshots-help.trycmd
@@ -36,16 +36,21 @@ Options:
       --log-level <LOG_LEVEL>
           Set the log output verbosity. [possible values: trace, debug, info, warn, error]
 
-      --all-image-names <PATH>
-          Path to a file containing the full list of preview names (one per line, comma-separated,
-          or both). When provided, only images in the upload directory are compared; missing images
-          listed here are reported as 'skipped' rather than 'removed'.
-
       --quiet
           Do not print any output while preserving correct exit code. This flag is currently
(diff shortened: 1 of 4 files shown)


szokeasaurusrex (approved):

tests/integration/_cases/build/build-snapshots-help.trycmd
```diff
@@ -36,16 +36,21 @@ Options:
       --log-level <LOG_LEVEL>
           Set the log output verbosity. [possible values: trace, debug, info, warn, error]
 
-      --head-sha <head_sha>
-          The VCS commit sha to use for the upload. If not provided, the current commit sha will be
-          used.
+      --all-image-names <PATH>
```
rbro112: I'd elaborate on what a "missing" image is and what this list does to clarify this more. Maybe "Path to a file containing a list of preview names (...). Sentry will use this list to determine what images to diff. Any image in this list will be diffed against the base. Images not uploaded, but present in this list will be marked as 'skipped' rather than 'removed'."

rbro112: Also name suggestions: `--all-snapshot-file-names` or `--all-image-file-names`, or `--diffable-image-file-names`

NicoHinderling: switched to `all-image-file-names` (in addition to adding the new "selective" flag)

PR #3275, opened 2026-04-21
fix(bundle-jvm): Strip source-set prefix from bundle URLs

romtsn:
Strip the `[<module>/]src/<sourceset>/<lang>/` prefix from bundle URLs so Symbolicator can resolve source context from JVM stack traces.

JVM stack traces reference classes by fully qualified name (e.g. `io.sentry.android.core.ANRWatchDog`) with no knowledge of which Gradle module or source-set directory on disk produced them. The previous URL format included the full on-disk prefix (e.g. `~/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.jvm`), so Symbolicator's URL-based lookup against the bundle manifest could never match a stack frame. Stripping the prefix produces URLs that match the package layout Symbolicator expects (`~/io/sentry/android/core/ANRWatchDog.jvm`).

Because stripping collapses different source sets into the same URL (e.g. `src/main/` and `src/debug/` both contributing `com.example.Foo` in an Android build variant), the command now tracks seen URLs and emits a `warn!` on collision, skipping all but the first-seen file. The warning points users at `--exclude` so they can scope the bundle to a single variant rather than bundling sources that would otherwise silently overwrite each other in the zip.

Follow-up to #3260, which enabled running `bundle-jvm` directly on a multi-module project root.

### DIFF
--- a/src/commands/debug_files/bundle_jvm.rs
+++ b/src/commands/debug_files/bundle_jvm.rs
@@ -244,6 +244,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         .extensions(JVM_EXTENSIONS.iter().copied())
         .ignores(all_excludes)
         .respect_ignores(true)
+        .sort_entries(true)
         .collect_files()?;
 
     let (files, collisions) = build_source_files(sources);
@@ -452,9 +453,11 @@ mod tests {
 
     #[test]
     fn test_build_source_files_records_collision_for_android_variants() {
+        // Sources arrive pre-sorted from `ReleaseFileSearch` (which configures
+        // `WalkBuilder::sort_by_file_name`); the first-seen wins in the dedup.
         let sources = vec![
-            fake_source("/app", "src/main/java/com/example/Config.java"),
             fake_source("/app", "src/debug/java/com/example/Config.java"),
+            fake_source("/app", "src/main/java/com/example/Config.java"),
         ];
         let (files, collisions) = build_source_files(sources);
 
@@ -462,18 +465,18 @@ mod tests {
         assert_eq!(files[0].url, "~/com/example/Config.jvm");
         assert_eq!(
             files[0].path,
-            Path::new("/app/src/main/java/com/example/Config.java")
+            Path::new("/app/src/debug/java/com/example/Config.java")
         );
 
         assert_eq!(collisions.len(), 1);
         assert_eq!(collisions[0].url, "~/com/example/Config.jvm");
         assert_eq!(
             collisions[0].skipped_path,
-            Path::new("/app/src/debug/java/com/example/Config.java")
+            Path::new("/app/src/main/java/com/example/Config.java")
         );
         assert_eq!(
             collisions[0].kept_path,
-            Path::new("/app/src/main/java/com/example/Config.java")
+            Path::new("/app/src/debug/java/com/example/Config.java")
         );
     }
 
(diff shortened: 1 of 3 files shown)


src/commands/debug_files/bundle_jvm.rs
```diff
@@ -22,6 +23,100 @@ const JVM_EXTENSIONS: &[&str] = &[
...
+            .to_str()
+            .is_some_and(|s| SOURCE_SET_LANGS.contains(&s));
+        if src_two_back && curr_is_lang {
+            return iter.collect();
+        }
+        src_two_back = src_one_back;
+        src_one_back = curr.as_os_str() == "src";
+    }
+    relative_path.to_path_buf()
+}
```
szokeasaurusrex, line 53: m: I'm finding it a bit difficult to parse how this code works with all the mutable variables, the while loop, etc.

Based on the documentation comment, it seems we might be able to use a regex to achieve the desired outcome, instead? If a regex is not possible or desirable, it might be a good idea to refactor this as an iterator chain.

src/commands/debug_files/bundle_jvm.rs
```diff
@@ -22,6 +23,100 @@ const JVM_EXTENSIONS: &[&str] = &[
...
+                        url,
+                        path: source.path,
+                        contents: Arc::new(source.contents),
+                        ty: SourceFileType::Source,
+                        headers: BTreeMap::new(),
+                        messages: vec![],
+                        already_uploaded: false,
+                    })
+                }
+            }
```
szokeasaurusrex: l: this `filter_map` is doing quite a lot of things all at once; I think it would make sense to separate it out into several different steps. Also, seems like we have a lot of clones; I think we may be able to avoid some of these by using reference types in the `SourceFile` and/or the `UrlCollision` structs

src/commands/debug_files/bundle_jvm.rs
```diff
@@ -149,27 +244,20 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
...
+    let (files, collisions) = build_source_files(sources);
+    for c in &collisions {
+        warn!(
+            "URL collision on {}: skipping '{}' (already bundled from '{}'). \
+             Use --exclude to drop the unwanted source set \
+             (e.g. --exclude='**src/debug/**').",
+            c.url,
+            c.skipped_path.display(),
+            c.kept_path.display(),
+        );
```
szokeasaurusrex: l: If we emit these warnings within `build_source_files`, I think we could avoid building the collisions list entirely. I am also fine with how you have it now, it has the benefit of splitting the concerns of building the source files and emitting warnings.

romtsn: yeah I also separated it out to make it testable, so I guess i'd keep it as-is 👍

src/commands/debug_files/bundle_jvm.rs
romtsn: converted this to a regex

src/commands/debug_files/bundle_jvm.rs
romtsn: split into two-step process - first collect candidate paths and then iterate and detect collisions.

src/commands/debug_files/bundle_jvm.rs
romtsn: ok, scratch that -- i got rid of the collisions struct to address your comment above (about many clones) and started emitting warnings inline + changed the test to capture log output and assert against it

szokeasaurusrex (approved):
Lgtm! I am assuming you've already tested this manually; if not, please do so before merging

romtsn:
> Lgtm! I am assuming you've already tested this manually; if not, please do so before merging

thanks! yeah I've tested it on our sentry-java monorepo, this is a sample event if you're interested (has source context applied to frames from different modules): https://sentry-sdks.sentry.io/issues/7235688870/events/72e9dde6502c454386a1315805753810/

PR #3277, opened 2026-04-22
docs(snapshots): Correct ImageMetadata CLI-managed fields list

runningcode:
## Summary
- The doc comment on `ImageMetadata` listed `image_file_name` as a CLI-managed override, but no such field is written by the CLI.
- `ImageMetadata::new` only sets `width` and `height`; the Sentry backend derives `image_file_name` from the manifest map key, not from a metadata field.
- Updates the comment to match reality.

### DIFF
--- a/src/api/data_types/snapshots.rs
+++ b/src/api/data_types/snapshots.rs
@@ -40,8 +40,8 @@ pub struct SnapshotsManifest<'a> {
 // Keep in sync with https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/snapshots/manifest.py
 /// Metadata for a single image in a snapshot manifest.
 ///
-/// CLI-managed fields (`image_file_name`, `width`, `height`) override any
-/// identically named fields provided by user sidecar metadata.
+/// CLI-managed fields (`width`, `height`) override any identically named
+/// fields provided by user sidecar metadata.
 #[derive(Debug, Serialize)]
 pub struct ImageMetadata {
     #[serde(flatten)]


cameroncooke (approved):
Yep I've confirmed image_file_name is not used. LGTM!

cameroncooke:
Claude backs this up:

> The sentry-cli pairs a sidecar JSON with an image purely by filename-stem match in the same directory — `button.png` loads `button.json` via `image_path.with_extension("json")`, so `image_file_name` is not needed (and can't be used) to match JSON to images ([sentry-cli/src/commands/build/snapshots.rs](https://github.com/getsentry/sentry-cli/blob/master/src/commands/build/snapshots.rs)). The server-side typed sidecar shape is `ImageMetadata` with `display_name`, `group`, `content_hash`, `width`, `height` plus `extra = "allow"` for arbitrary passthrough keys ([manifest.py](https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/snapshots/manifest.py)), but the CLI overwrites `content_hash`, `width`, and `height` before upload, which leaves only `display_name` and `group` as genuinely user-settable sidecar fields (confirmed by the JS fixture [snapshot-image-metadata.ts](https://github.com/getsentry/sentry/blob/master/tests/js/sentry-test/snapshots/snapshot-image-metadata.ts)); `image_file_name` is actually the key of the manifest `images` dict (the image's relative path), and the GET response explicitly rebuilds it from that key while stripping any `image_file_name` you smuggled through the sidecar's extras ([preprod_artifact_snapshot.py](https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/api/endpoints/preprod_artifact_snapshot.py), [comparison_categorizer.py](https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/snapshots/comparison_categorizer.py), and see the CLI's own note in [snapshots.rs data types](https://github.com/getsentry/sentry-cli/blob/master/src/api/data_types/snapshots.rs)).

NicoHinderling (approved):

rbro112 (approved):

PR #3280, opened 2026-04-24
ref(upload): Remove redundant function parameter

loewenheim:
### Description
The `options` parameter on `upload_files_chunked` and `poll_assemble` is already contained in the `context` parameter for both functions.


szokeasaurusrex (approved):
cool

PR #3287, opened 2026-04-30
fix(js): Prevent argument injection via type coercion in serializeOptions

JPeer264:
### Description

Add runtime type validation for string/number options to match existing validation for array and boolean types. Wrap arguments in array literal to prevent concat() from flattening array inputs.

(there is no ticket for it)

### Legal Boilerplate


szokeasaurusrex:
I'll let someone from the JS SDK team review this

andreiborza (approved):

PR #3305, opened 2026-05-19
perf(snapshots): Skip uploading images that already exist in objectstore

NicoHinderling:
Uses the new batch HEAD operation added in [objectstore#478](https://github.com/getsentry/objectstore/pull/478) to check which images already exist before uploading. Since image keys are content-addressed (SHA-256 hash), unchanged images between runs produce the same key and don't need re-uploading.

**How it works:**
- After hashing all images, batch HEAD all keys via `session.many()` to check existence
- Only batch PUT images where HEAD returned 404 (not found)
- HEAD errors are treated as "doesn't exist" — safe fallback that just re-uploads

**Performance with 40k images:**
- First upload (all new): ~33 minutes
- Second upload (all cached): ~4.5 minutes (~7.5x faster)

Bumps `objectstore-client` from 0.1.6 to 0.1.9 and adds `futures-util` for `StreamExt::next()` on the HEAD results stream.

Note: the objectstore expiration policy is still TTL. Once it's changed to TTI, the HEAD requests will also bump expiration timers, preventing frequently-used images from expiring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Performance
+
+- (snapshots) Skip uploading images that already exist in objectstore by batch-checking with HEAD requests first ([#3305](https://github.com/getsentry/sentry-cli/pull/3305))
+
 ### Fixes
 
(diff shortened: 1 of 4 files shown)


matt-codecov (approved):
lgtm from objectstore PoV

szokeasaurusrex (approved):

PR #3311, opened 2026-05-26
fix(config): Improve error message for missing organization

NicoHinderling:
When no organization is provided, the error message previously only
mentioned the `--org` flag. This can be confusing for users who may
not know about other ways to configure the organization.

The updated error message now lists all three options:
- `--org` flag
- `SENTRY_ORG` environment variable
- Org-scoped auth token

This helps users resolve the issue faster, especially in CI
environments where env vars or scoped tokens are the preferred
approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


mtopo27 (approved):

szokeasaurusrex (approved):
lgtm 🚀

PR #3336, opened 2026-06-16
feat(snapshots): Compress preprod snapshot manifest with zstd

NicoHinderling:
## Summary

The preprod snapshot manifest sent by `create_preprod_snapshot` can be large. This change compresses the manifest JSON body with zstd before uploading, reducing the number of bytes transferred over the wire.

A new `with_zstd_json_body` request builder serializes the payload to JSON, compresses it with `zstd::encode_all`, and sets `Content-Type: application/json` alongside `Content-Encoding: zstd` (the content type describes the decoded payload; the encoding describes the wire transform).

## Changes

- Add `zstd = "0.13.3"` dependency.
- Add `ApiRequest::with_zstd_json_body` helper.
- Route `create_preprod_snapshot` through the zstd-compressed body path.

## Notes

- Requires the receiving endpoint to decode `Content-Encoding: zstd` on request bodies — backend support should be confirmed deployed before this ships.


rbro112 (approved):

src/api/mod.rs
```diff
@@ -1360,6 +1362,23 @@ impl ApiRequest {
...
+        debug!(
+            "zstd json body: {} bytes compressed to {} bytes",
+            body_bytes.len(),
+            compressed.len()
+        );
+        self.body = Some(compressed);
+        self.headers.append("Content-Type: application/json")?;
+        self.headers.append("Content-Encoding: zstd")?;
+        Ok(self)
+    }
```
szokeasaurusrex, line 1388: m: The current code unnecessarily serializes the entire body into memory uncompressed prior to compressing it. 

With the following patch, we avoid the extra allocation by writing directly into a `zstd` encoder.

```patch
diff --git a/src/api/mod.rs b/src/api/mod.rs
index e7ca1c00..24a37418 100644
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -42,6 +42,7 @@ use symbolic::common::DebugId;
 use symbolic::debuginfo::ObjectKind;
 use url::Url;
 use uuid::Uuid;
+use zstd::Encoder as ZstdEncoder;
 
 use crate::api::errors::{ProjectRenamedError, RetryError};
 use crate::config::{Auth, Config};
@@ -1363,16 +1364,23 @@ impl ApiRequest {
     }
 
     pub fn with_zstd_json_body<S: Serialize>(mut self, body: &S) -> ApiResult<Self> {
-        let mut body_bytes: Vec<u8> = vec![];
-        serde_json::to_writer(&mut body_bytes, &body)
-            .map_err(|err| ApiError::with_source(ApiErrorKind::CannotSerializeAsJson, err))?;
-        let compressed = zstd::encode_all(body_bytes.as_slice(), 0)
+        let mut encoder = ZstdEncoder::new(Vec::new(), 0)
             .map_err(|err| ApiError::with_source(ApiErrorKind::CompressionFailed, err))?;
-        debug!(
-            "zstd json body: {} bytes compressed to {} bytes",
-            body_bytes.len(),
-            compressed.len()
-        );
+
+        serde_json::to_writer(&mut encoder, &body).map_err(|err| {
+            let kind = if err.is_io() {
+                ApiErrorKind::CompressionFailed
+            } else {
+                ApiErrorKind::CannotSerializeAsJson
+            };
+            ApiError::with_source(kind, err)
+        })?;
+
+        let compressed = encoder
+            .finish()
+            .map_err(|err| ApiError::with_source(ApiErrorKind::CompressionFailed, err))?;
+
+        debug!("zstd json body: {} bytes compressed", compressed.len());
         self.body = Some(compressed);
         self.headers.append("Content-Type: application/json")?;
         self.headers.append("Content-Encoding: zstd")?;

```

lcian (approved):
Stamping to unblock (I disabled auto-merge), please address or dismiss Daniel's comment.

szokeasaurusrex (approved):
also stamping, the rest should be fine once my previous comment is addressed

src/api/mod.rs
NicoHinderling: thanks for flagging this 🙏

PR #3342, opened 2026-06-23
fix(auth): Route org token requests to region_url

NicoHinderling:
## Summary

Org auth tokens (`sntrys_`) for multi-region SaaS orgs carry **two** hosts in their payload:

- `url` — the control silo (`https://sentry.io`)
- `region_url` — where the org's API requests are actually served (`https://<region>.sentry.io`)

The CLI only parsed `url` and **discarded `region_url` entirely**. For region-scoped org tokens (e.g. `us2`, `s4s2`), that meant region-siloed requests — like `snapshots upload` — were sent to the control silo and rejected with `401 Invalid org token`. Decoded examples (both real region tokens):

```json
{ "url": "https://sentry.io", "region_url": "https://us2.sentry.io", "org": "us2-test-org-sc" }
{ "url": "https://sentry.io", "region_url": "https://s4s2.sentry.io", "org": "sentry-s4s2" }
```

Passing `--url` did not help: when a token embeds a URL, the CLI overrides `--url` with the token's URL — so requests still went to `sentry.io`.

## Fix

- Parse `region_url` from the token payload (`AuthTokenPayload`).
- Add a `base_url()` helper that prefers `region_url`, falling back to `url` when absent.
- Use `base_url()` everywhere `config.rs` derives the API host (`from_file`, `set_auth`, `set_base_url`).

Result: every command routes to the correct region automatically, with **no `--url` flag needed**. Verified end-to-end — a `snapshots upload` to `sentry-s4s2` now succeeds (previously `401`).

## Compatibility / behavior notes

- Tokens where `url == region_url` (self-hosted, classic SaaS) and legacy tokens with no `region_url` are **unaffected** — `base_url()` returns the same value as before.
- **Behavior change:** a token with an empty `url` but a present `region_url` now routes to `region_url` and overrides `--url`, whereas it previously let `--url` take precedence. This is the correct behavior for region-scoped tokens and keeps the no-flag default routing to the right region.
- This does **not** affect personal/user tokens, which carry no region info; those still rely on default routing or an explicit `--url`.

## Tests

- Unit tests for `base_url()` precedence: differing `url`/`region_url`, null `url` with present `region_url`, and a legacy token with `region_url` absent.
- Integration (`trycmd`): new `region-url-routing` case asserting `region_url` is preferred over both the token's `url` and `--url`; updated `url-mismatch-empty-token` to reflect the new behavior.
- Full unit suite (186) + `org_tokens` integration cases pass; `clippy` clean; `cargo fmt` applied.

### DIFF
--- a/src/utils/auth_token/org_auth_token.rs
+++ b/src/utils/auth_token/org_auth_token.rs
@@ -19,6 +19,23 @@ pub struct AuthTokenPayload {
     // URL may be missing from some old auth tokens, see getsentry/sentry#57123
     #[serde(deserialize_with = "url_deserializer")]
     pub url: String,
+
+    // Region URL is the host that actually serves this org's API requests. It is
+    // absent from older tokens, in which case requests fall back to `url`.
+    #[serde(default, deserialize_with = "url_deserializer")]
+    pub region_url: String,
+}
+
+impl AuthTokenPayload {
+    /// Returns the base URL that API requests for this org should target,
+    /// preferring the region URL and falling back to `url` when it is absent.
+    pub fn base_url(&self) -> &str {
+        if self.region_url.is_empty() {
+            &self.url
+        } else {
+            &self.region_url
+        }
+    }
 }
 
 /// Deserializes a URL from a string, returning an empty string if the URL is missing or null.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - (snapshots) Compress preprod snapshot manifest with zstd ([#3336](https://github.com/getsentry/sentry-cli/pull/3336))
 
+### Fixes
+
+- (auth) Route org token requests to region_url ([#3342](https://github.com/getsentry/sentry-cli/pull/3342))
+
 ## 3.5.1
 
 ### Internal Changes 🔧
--- a/src/config.rs
+++ b/src/config.rs
@@ -75,7 +75,7 @@ impl Config {
         let manually_configured_url = configured_url(&ini);
         let token_url = token_embedded_data
             .as_ref()
-            .map(|td| td.url.as_str())
+            .map(|td| td.base_url())
             .unwrap_or_default();
 
         let url = if token_url.is_empty() {
@@ -201,7 +201,7 @@ impl Config {
             Some(Auth::Token(ref val)) => {
                 self.cached_token_data = val.payload().cloned();
 
-                if let Some(token_url) = self.cached_token_data.as_ref().map(|td| td.url.as_str()) {
+                if let Some(token_url) = self.cached_token_data.as_ref().map(|td| td.base_url()) {
                     self.cached_base_url = token_url.to_owned();
                 }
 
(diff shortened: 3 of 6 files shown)


src/utils/auth_token/org_auth_token.rs
```diff
@@ -19,6 +19,23 @@ pub struct AuthTokenPayload {
     // URL may be missing from some old auth tokens, see getsentry/sentry#57123
     #[serde(deserialize_with = "url_deserializer")]
     pub url: String,
+
+    // Region URL is the host that actually serves this org's API requests. It is
+    // absent from older tokens, in which case requests fall back to `url`.
+    #[serde(default, deserialize_with = "url_deserializer")]
+    pub region_url: String,
```
lcian, line 26:
```suggestion
    pub region_url: Option<String>,
```
This would be the proper type, instead of using an empty string as a sentinel value

NicoHinderling:
@szokeasaurusrex this change enabled me to locally upload snapshots to us2 but im not 100% confident as i'm not working on the us2 project directly so ive shared this PR with that team and won't merge until they signoff too

src/utils/auth_token/org_auth_token.rs
lcian: And I guess the reason the clanker used String is that we're gonna need a new `opt_url_deserializer` to do it that way

src/utils/auth_token/org_auth_token.rs
```diff
@@ -19,6 +19,23 @@ pub struct AuthTokenPayload {
...
+
+impl AuthTokenPayload {
+    /// Returns the base URL that API requests for this org should target,
+    /// preferring the region URL and falling back to `url` when it is absent.
+    pub fn base_url(&self) -> &str {
+        if self.region_url.is_empty() {
+            &self.url
+        } else {
+            &self.region_url
+        }
```
markstory, line 37: This part makes sense to me as most CLI operations (and org auth tokens in general) work on regional resources. This would make it so that org-auth-tokens can't be used on anything that involves control though.

szokeasaurusrex: +1 on this: Broadly switching all operations to prefer the region URL could be risky; we should validate that all endpoints supported by org auth tokens also are available on the region.

Less risky could be to expose a way to get the region URL while still maintaining a way to get the base URL, although that is likely a bit complex. That would allow us to use the region URL where needed and for commands where we have validated that it works, while continuing to use the base URL elsewhere.

lcian: I noticed that an old version of CLI had this utility https://github.com/getsentry/sentry-cli/blob/de5cda234ac06bae346e6f3c7991e4f8fa36e228/src/api/mod.rs#L1191
Maybe relevant for the discussion here?

szokeasaurusrex: Yeah, we had used this for the pre-chunked-uploading uploads that were removed in Sentry CLI 3.x. We no longer needed it for chunked uploads, as the chunked upload endpoint already returns the region-specific URL as the upload URL, which is why this was removed.

If we go with a fix in the CLI, though, we might want something like this utility so that we only migrate the API calls that need to go to the region-specific endpoints over to the region specific URLs; that lowers the risk of this change by keeping most things on control silo.

NicoHinderling:
We can avoid doing this by changing the nginx routing logic fortunately. i didn't fully deploy the changes yesterday and something broke

PR #3344, opened 2026-06-24
feat(snapshots): Download via archive endpoint with build-and-poll

NicoHinderling:
The server-side snapshot `download/` endpoint was removed and replaced by an `archive/` endpoint that separates archive creation from download. A `GET .../archive/` probes readiness (`{"ready": bool}`), a `POST .../archive/` triggers an async build (`202`), and `GET .../archive/?download` streams the zip once it exists.

`snapshots download` now matches that flow: it probes readiness, triggers a build when the archive isn't ready, polls until ready (2s interval, 300s timeout), then downloads and extracts as before. This applies to both `--snapshot-id` and `--app-id` resolution, since the latter just resolves to a snapshot id first.

The build-and-poll loop is the new behavior worth reviewing — `wait_for_archive` in `src/commands/snapshots/download.rs`. Integration tests cover the two distinct paths: archive already ready (no build triggered) and not-ready (build + poll + download).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- (snapshots) Update `snapshots download` to use the new archive endpoint, triggering the archive build and polling until it is ready before downloading ([#3344](https://github.com/getsentry/sentry-cli/pull/3344))
 - (snapshots) Show a clear "project was renamed" error instead of a cryptic JSON parse failure when uploading to a renamed project slug ([#3341](https://github.com/getsentry/sentry-cli/pull/3341))
 
 ## 3.5.1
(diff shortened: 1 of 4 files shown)


szokeasaurusrex (approved):
Lgtm, but I would consider adding parameters to override the timeout and possibly also the polling interval

NicoHinderling:
> Lgtm, but I would consider adding parameters to override the timeout and possibly also the polling interval

Most of the time this polling should only happen once or twice before succeeding so it should be really fast

Also i dont think this feature will get a ton of usage so im inclined to not worry too much about such customization..

szokeasaurusrex:
@NicoHinderling fair enough, but if we expect it to complete quickly, we might want to lower the timeout to just 10-20 seconds.

But not a huge deal if it is not expected to be used widely

PR #3347, opened 2026-06-26
fix(update): correct inverted version comparison in update nagger

### LINKED ISSUE #3290: Fix inverted update nagger version comparison
`LastUpdateCheck::is_outdated()` appears to compare the fetched release version and current CLI version in the wrong direction. In [`src/utils/update.rs`](https://github.com/getsentry/sentry-cli/blob/b383d388a3167937ac78a7f294214838e45a649b/src/utils/update.rs#L105-L116), the current condition checks whether the fetched release version is older than the running CLI version. For a normal outdated install, such as current version `3.3.0` with fetched latest release `3.4.1`, the condition evaluates to `3.4.1 < 3.3.0`, so the update-available message is not shown.

The comparison should check whether the running CLI version is older than the fetched release version, for example `Version::parse(VERSION).unwrap() < Version::parse(release_v.as_str()).unwrap()`. With that direction, the update nagger will show `sentry-cli update to <version> is available!` when an update is available.

syf2211:
## Summary

Fix the inverted semver comparison in `LastUpdateCheck::is_outdated()` so the update nagger correctly prompts when a newer release is available.

## Motivation

Fixes #3290

When the running CLI version is older than the latest fetched release (e.g. running `3.3.0` with fetched `3.4.1`), the update nagger never showed because the comparison checked `release < running` instead of `running < release`.

## Changes

- Flip the version comparison in `is_outdated()` to `VERSION < release_v`
- Add four unit tests covering outdated, up-to-date, dev-ahead, and stale-cache scenarios
- Add CHANGELOG entry under Unreleased → Fixes

## Tests

```bash
cargo test utils::update::tests
```

Result: 4 passed, 0 failed

## Notes

- The `check_v == VERSION` guard is unchanged — it prevents nagging from stale cache after the binary was upgraded since the last check.
- `Version::parse(...).unwrap()` panic behavior is unchanged legacy code (file already has `clippy::unwrap_used` expect).


src/utils/update.rs
```diff
@@ -324,3 +324,40 @@ pub fn run_sentrycli_update_nagger() {
...
+        let check = check_with_versions("0.0.1", VERSION);
+        assert!(!check.is_outdated());
+    }
+
+    #[test]
+    fn is_not_outdated_when_last_check_version_differs_from_running_version() {
+        let check = check_with_versions("99.0.0", "0.0.1");
+        assert!(!check.is_outdated());
+    }
+}
```
szokeasaurusrex: I don't think we need to have these tests

```suggestion
```

szokeasaurusrex (commented):
lgtm in general, thanks for the contribution @syf2211

szokeasaurusrex (approved):

PR #3348, opened 2026-06-26
fix(releases): resolve remote name to URL in set-commits --local

### LINKED ISSUE #3254: set-commits --local sends remote name instead of parsed repo name as repository
## Environment

- sentry-cli version: 3.3.5 (latest)
- Platform: Linux (GitHub Actions, `actions/checkout`)
- Git remote URL: `https://github.com/our-org/our-repo` (set by `actions/checkout`)

## Description

When using `sentry-cli releases set-commits --local`, the `repository` field in the commits payload is set to the git remote **name** (e.g. `"origin"`) instead of the parsed repository identifier (e.g. `"our-org/our-repo"`).

This causes the commits to be associated with a stale `"origin"` repository entry in Sentry (Unknown Provider) instead of the correct GitHub-integrated repository.

## Root Cause

In `src/commands/releases/set_commits.rs` (lines 203-204), the `--local` flow does:

```rust
let remote = config.get_cached_vcs_remote();       // returns "origin" (remote name)
let parsed = get_repo_from_remote(&remote);         // passes "origin" to URL parser
```

`get_cached_vcs_remote()` returns the remote **name** (default `"origin"` from `get_default_vcs_remote()` in `src/config.rs`).

`get_repo_from_remote()` calls `VcsUrl::parse()` which expects a **URL**. Since `"origin"` matches neither `GIT_URL_RE` nor `GIT_SSH_RE`, it falls through to the default case:

```rust
VcsUrl {
    provider: "".into(),
    id: url.into(),  // "origin" is returned as-is
}
```

## Comparison with `--auto` flow

The `--auto` flow correctly handles this. In `find_matching_rev()` (vcs.rs line 474):

```rust
if let Ok(remote) = repo.find_remote(&remote_name.unwrap_or_else(|| "origin".to_owned()));
if let Some(url) = remote.url();
```

Here, `"origin"` is used as a remote **name** to look up the actual URL via `git2::Repository::find_remote()`, and the URL is then used for matching.

The `--local` flow skips this URL resolution step.

## Expected Fix

In the `--local` flow, resolve the remote name to its URL before passing to `get_repo_from_remote()`:

```rust
let remote_name = config.get_cached_vcs_remote();
let repo = git2::Repository::open_from_env()?;
let remote_url = git_repo_remote_url(&repo, &remote_name)?;  // already exists in vcs.rs
let parsed = get_repo_from_remote(&remote_url);
```

`git_repo_remote_url()` already exists in `src/utils/vcs.rs` (line 253) and does exactly this lookup.

## Current Workaround

Setting `SENTRY_VCS_REMOTE` to the full URL instead of relying on the default:

```bash
export SENTRY_VCS_REMOTE="$(git remote get-url origin)"
sentry-cli releases set-commits --local "$VERSION" --ignore-missing --initial-depth 20
```

This works because `get_default_vcs_remote()` prioritizes the environment variable, and `VcsUrl::parse()` can correctly parse the full URL.

## Impact

- `--local` mode sends `"repository": "origin"` in the commits payload
- patchSet data is computed correctly but associated with the wrong repository
- Suspect Commits feature does not work because commits are not linked to the GitHub-integrated repository

syf2211:
## Summary

Fix `releases set-commits --local` sending the git remote name (e.g. `origin`) as the repository identifier instead of the parsed `org/repo` slug.

## Motivation

`get_cached_vcs_remote()` returns a remote **name** by default (`origin`), but the `--local` flow passed that value directly to `get_repo_from_remote()`, which expects a URL. When the value is not a URL, `VcsUrl::parse` falls through and uses the string as-is, so commits were associated with a stale `"origin"` repository entry in Sentry.

The `--auto` flow already resolves the remote name to its URL via `git2::Repository::find_remote()` before parsing. This change brings `--local` in line with that behavior.

Fixes #3254

## Changes

- In `set_commits.rs`, resolve the configured VCS remote name to its URL via `git_repo_remote_url()` before calling `get_repo_from_remote()`.
- Fall back to the configured value when it is already a URL (e.g. `SENTRY_VCS_REMOTE=https://github.com/org/repo`).
- Add a regression test in `vcs.rs` demonstrating correct URL resolution vs the previous failure mode.
- Add CHANGELOG entry.

## Tests

```bash
cargo test test_get_repo_from_remote_resolves_git_remote_name_to_url -- --nocapture
```

Result: **1 passed**

## Notes

- Uses the same resolve-then-parse pattern as `code_mappings/upload.rs`.
- When a remote exists but has no URL, behavior degrades to the pre-existing fallback (rare misconfiguration).


szokeasaurusrex:
Hi @syf2211, it looks like some unintentional changes have also been introduced with this PR, which makes it impossible to review.

Please feel free to open a new PR with only the intended changes.

PR #3352, opened 2026-07-02
ref(organizations): Use control silo endpoint for org listing

ldelvoye:
### Description
The organization listing endpoint (GET /organizations/) is now served from the control silo and returns every organization the authenticated user belongs to across all regions in a single paginated response. This replaces the previous per-region fan-out (which called /users/me/regions/ and then queried each region) with one call.

Also aligns the Organization model with the current API response: drops requireEmailVerification (removed in getsentry/sentry#115003) and makes features optional (removed from the listing endpoint in #115007), so no longer fails to deserialize the live response.

### Issues
* resolves: INFRENG-199

### Legal Boilerplate

### DIFF
--- a/tests/integration/organizations.rs
+++ b/tests/integration/organizations.rs
@@ -3,9 +3,6 @@ use crate::integration::{MockEndpointBuilder, TestManager};
 #[test]
 fn command_organizations() {
     TestManager::new()
-        // Mock is for the organizations list command. The listing is served
-        // from a single (control silo) endpoint, so there is no per-region
-        // fan-out and no `/users/me/regions/` call to mock.
         .mock_endpoint(
             MockEndpointBuilder::new("GET", "/api/0/organizations/?cursor=")
                 .with_response_file("organizations/get-organizations.json"),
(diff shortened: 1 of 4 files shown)


tests/integration/organizations.rs
```diff
@@ -2,27 +2,14 @@ use crate::integration::{MockEndpointBuilder, TestManager};
...
-        }}"#,
-        manager.server_url(),
-    );
-
-    manager
-        // Mocks are for the organizations list command.
+    TestManager::new()
+        // Mock is for the organizations list command. The listing is served
+        // from a single (control silo) endpoint, so there is no per-region
+        // fan-out and no `/users/me/regions/` call to mock.
```
szokeasaurusrex: [l:](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale) imo we can remove this comment because it only makes sense in the context of the code's previous state. I believe this was the only endpoint which ever had a fan-out, so leaving the comment here risks confusing future maintainers.

```suggestion
```

szokeasaurusrex (approved):
lgtm! Feel free to merge after addressing the inline comment I left below.

One note I found while testing: the command does not work with organization auth tokens, only user tokens. But, that was already the case before this change, as well, and it might be an intentional limitation of the endpoint. So, I would say this is all good 👍

PR #3362, opened 2026-07-11
fix(sourcemaps): Guess appended .map before replaced extension

### LINKED ISSUE #2802: Reorder `guess_sourcemap_reference` guesses
Per the [source map format specification](https://tc39.es/ecma426/#sec-source-map-naming), we should reorder the guesses we make in [`guess_sourcemap_reference`](https://github.com/getsentry/sentry-cli/blob/5217470c0244b5dc44495f1ddd8c7f4171118a58/src/utils/sourcemaps.rs#L109) to first guess `foo.min.js -> foo.min.js.map`, as this is the default naming convention for sourcemaps.

vismaytiwari:
### Description

`guess_sourcemap_reference` guessed `foo.min.js` → `foo.map` (replacing the extension) *before* `foo.min.js` → `foo.min.js.map` (appending `.map`). Per the source map naming convention, appending `.map` to the file name is the default reference form, so when both a `<name>.map` and a `<name>.min.js.map` are present the appended one should win.

This reorders the guesses so the appended reference is tried first, and adds a unit test covering the ordering and each individual form.

### Issues

* resolves: getsentry/sentry-cli#2802

### Legal Boilerplate

### DIFF
--- a/src/utils/sourcemaps.rs
+++ b/src/utils/sourcemaps.rs
@@ -109,6 +109,20 @@ fn guess_sourcemap_reference(
     let map_ext = "map";
     let (path, basename, ext) = split_url(min_url);
 
+    if let Some(ext) = ext.as_ref() {
+        // foo.min.js -> foo.min.js.map
+        // Per the source map spec, a reference is formed by appending `.map` to
+        // the file name, so prefer this over replacing the extension (gh-2802).
+        let new_ext = format!("{ext}.{map_ext}");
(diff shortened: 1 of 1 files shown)


szokeasaurusrex:
Thanks for the contribution. This issue was reported internally and is currently low priority for us. We are not planning to invest maintainer time in reviewing or accepting external changes for it at this time, so I’m going to close this PR.

PR #3377, opened 2026-07-21
fix(auth): use org token region_url as API base URL

shayna-ch:
Org auth tokens embed both `url` (always the primary US URL) and `region_url`, but the CLI only read `url`. For orgs with non-default data residency (e.g. EU), uploads routed to the wrong region and failed with `Invalid org token`, ignoring `SENTRY_URL`/`--url`.

Now prefers `region_url` as the API base URL when present, falling back to `url` for older tokens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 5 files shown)


szokeasaurusrex (changes requested):
Not all endpoints support region URLs as some of the endpoints are on the control silo (sentry.io), but all of the endpoints we call directly are reachable through sentry.io. I am not deeply familiar with the exact mechanisms for this, but essentially there is a reverse proxy from the control silo to the region silos. 

This change should not be necessary at all. I'd appreciate more context about the problem so we can identify the root cause and find an acceptable solution

PR #3378, opened 2026-07-22
fix: Ensure URL and Auth Token come from same source

szokeasaurusrex:
Fixes [SDK-1233](https://linear.app/getsentry/issue/SDK-1233/)


src/config.rs
```diff
@@ -791,6 +911,104 @@ mod tests {
...
+            Some("https://global.invalid")
+        );
+        assert_eq!(base.get_from(Some("auth"), "token"), Some(USER_TOKEN));
+        assert_eq!(base.get_from(Some("http"), "verify_ssl"), Some("false"));
+    }
+
+    #[test]
+    fn cli_and_environment_form_one_runtime_source() {
+        let cli_token = AuthToken::from("cli-token");
+        let mut auth_and_url = AuthAndUrl {
```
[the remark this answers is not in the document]
szokeasaurusrex: That's on purpose

src/config.rs
```diff
@@ -58,44 +59,48 @@ pub struct Config {
 }
 
 impl Config {
-    /// Loads the CLI config from the default location and returns it.
-    pub fn from_cli_config() -> Result<Config> {
-        let (filename, ini) = load_cli_config()?;
-        Ok(Config::from_file(filename, ini))
+    /// Loads config files and applies URL and token values from the environment and CLI.
+    pub fn from_cli_config(cli_url: Option<&str>, cli_token: Option<&AuthToken>) -> Result<Config> {
+        load_cli_config(cli_url, cli_token)
```
loewenheim: Is there a reason this needs to be a separate function?

src/config.rs
```diff
@@ -572,12 +572,140 @@ fn max_retries_from_ini(ini: &Ini) -> Result<Option<u32>, ParseIntError> {
...
+        environment_url: Option<String>,
+        environment_token: Option<AuthToken>,
+        cli_url: Option<&str>,
+        cli_token: Option<&AuthToken>,
+    ) -> Option<DiscardedAuthUrlValue> {
+        self.merge(AuthAndUrl {
+            url: cli_url.map(str::to_owned).or(environment_url),
+            token: cli_token.cloned().or(environment_token),
+        })
+    }
```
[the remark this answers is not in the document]
szokeasaurusrex: This is intentional

src/commands/login.rs
```diff
@@ -126,7 +128,7 @@ pub fn execute(matches: &ArgMatches) -> Result<()> {
         }
     }
 
-    update_config(&config_to_update, token)?;
+    update_config(&config_to_update, token, &validated_url)?;
```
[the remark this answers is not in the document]
szokeasaurusrex: This is an unavoidable side effect of these changes. URLs must be set together with auth tokens now

src/config.rs
```diff
@@ -659,11 +790,7 @@ fn load_cli_config() -> Result<(PathBuf, Ini)> {
         let mut f = fs::File::open(&project_config_path)
             .context(failed_local_config_load_message(&file_desc))?;
         let ini = Ini::read_from(&mut f).context(format!("Failed to parse {file_desc}"))?;
-        for (section, props) in ini.iter() {
-            for (key, value) in props.iter() {
-                rv.set_to(section, key.to_owned(), value.to_owned());
-            }
-        }
+        warning = merge_config_source(&mut rv, &ini).or(warning);
```
loewenheim: This `or` and the one below are unnecessary, `warning` is definitely `None` prior to this point.

src/config.rs
szokeasaurusrex: I guess not anymore, as the function was only called here. I guess it is still here because it made sense before.

src/config.rs
szokeasaurusrex: This `or` does seem unnecessary, but I don't think that the one below is.

loewenheim: I don't see how `warning` can be `Some` before line 822; the only two assignments to it are in two different branches of an `if`, right?

szokeasaurusrex: No, I think the way the GitHub PR diff view collapses the diff makes it look that way, but if you expand the lines, you'll see that the `else` block is pretty short, and line 822 comes after it. 

So this line can set `warning` to `Some`, then line 822 can encounter this value.

loewenheim: Oh, you're right, I did misread that. So the second `or` is necessary.

loewenheim (approved):

sysmat:
- this is breaking and you bump patch version
- most modern software support configuration file and env variables, this is very bad design

szokeasaurusrex:
> - this is breaking and you bump patch version
> - most modern software support configuration file and env variables, this is very bad design

The changes are intentional. Please see https://github.com/getsentry/sentry-cli/issues/3380#issuecomment-5059013026.

PR #3387, opened 2026-07-31
fix(logs): Correct severity query example

jamieQ:
## Summary

- replace the misleading `level:error` Logs query example with `severity:error`
- update the CLI help snapshot

## Why

The Logs API uses `severity` for the canonical log severity. `level` is treated as an independent custom attribute, not an alias, so the existing example commonly returns no matches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


szokeasaurusrex (approved):
Nice find, thank you for fixing!

PR #3393, opened 2026-08-13
feat(build): Support dSYMs with IPA uploads

jamieQ:
Adds support for a new `--dsym` option when uploading IPA files through the `build` command. This allows clients to associate app debug symbols with an IPA upload (which typically does not include them). Internally the uploader follows the existing logic to construct a synthetic XCArchive, but will copy the specified dSYM bundles it finds into that archive before uploading. This will give clients a better experience when they upload builds that have undergone "app thinning" since they will not have to perform the synthetic archive construction themselves.

Refs EME-1285

### DIFF
--- a/src/commands/build/upload.rs
+++ b/src/commands/build/upload.rs
@@ -35,7 +35,7 @@ pub fn make_command(command: Command) -> Command {
     #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
     const HELP_TEXT: &str =
         "The path to the build to upload. Supported files include Apk, and Aab.";
-    command
+    let command = command
         .about("Upload builds to a project.")
         .long_about("Upload builds to a project.\n\nThis feature only works with Sentry SaaS.")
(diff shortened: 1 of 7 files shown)


NicoHinderling (approved):

src/commands/build/upload.rs
```diff
@@ -47,7 +47,8 @@ pub fn make_command(command: Command) -> Command {
                 .num_args(1..)
                 .action(ArgAction::Append)
                 .required(true),
-        )
+        );
+    let command = command
```
szokeasaurusrex: l: unsure why this change is necessary


```suggestion
        )
```

szokeasaurusrex (approved):
I only skimmed this, relying on @NicoHinderling's approval here

src/commands/build/upload.rs
jamieQ: Thanks for pointing that out! Fixed.

PR #3396, opened 2026-08-19
feat: Add Linux riscv64 support

Xeonacid:
### Description

Build and distribute sentry-cli for Linux riscv64 across GitHub releases, npm, PyPI, and Homebrew. Leave package-lock.json unchanged because @sentry/cli-linux-riscv64 is not published yet.

### Legal Boilerplate

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -27,6 +27,9 @@ targets:
     id: '@sentry/cli-linux-i686'
     includeNames: /^sentry-cli-linux-i686-\d.*\.tgz$/
   - name: npm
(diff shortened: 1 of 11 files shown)


JPeer264:
@szokeasaurusrex whats your take on this given that we move soon to the new cli which has no binaries at all

szokeasaurusrex:
@Xeonacid for the reason @JPeer264 mentioned, and also because this would be an additional maintenance burden for us, I don't think we will add any additional officially built binaries to the Sentry CLI beyond what we already have, at least not for now

PR #3403, opened 2026-09-09
feat(wasm-debug-files): Add prepare command for WASM debug setup

d2anamaria:
### Problem

WebAssembly needs manual work before Sentry can symbolicate crashes. Today users run `wasm-split` to inject a `build_id` and split DWARF into a companion file, then run `sentry-cli debug-files upload --type wasm` on that companion. Two tools, two steps, easy to get wrong. Nothing warns you when a module has no usable debug info, so silent failures show up later as unsymbolicated stack traces.

### Solution

`sentry-cli debug-files prepare <PATH>` does the whole pipeline in one command.

```bash
sentry-cli debug-files prepare ./dist
```

It scans for `.wasm` modules, classifies each one, splits the ones that have DWARF, and uploads the companions.

### Behavior

Every module the command inspects gets a `build_id` if it has none, whether or not it can be split. The `build_id` is how Sentry ties a module in a stack trace to its debug file, so a module without one can never be symbolicated — not even by a debug file uploaded later. Stamping up front keeps that door open.

For each `.wasm` module with embedded DWARF:

1. Inject a `build_id` if the module has none.
2. Write a `*.debug.wasm` companion that keeps the Code section and DWARF.
3. Strip `.debug_*` sections from the deployable `.wasm`, in place.
4. Point the stripped module at its companion via `external_debug_info`.
5. Upload the companion to Sentry.

The stripped module stays at the original path, so your deploy artifact does not move. Both files carry the same `build_id`, which is how Sentry matches a crash to its debug file.

### Outcomes

| Outcome | When |
| --- | --- |
| Split | Module had DWARF; companion written |
| Already prepared | Companion exists with matching `build_id` |
| Would split | `--dry-run` preview |
| Skipped | Module is not splittable; stamped, warning printed |

### Warnings

A module without usable debug info does not stop the run. It is stamped and left otherwise untouched, and the command explains why it was not split:

- Name/symtab only: `no line-level symbolication (name/symtab only)`
- No debug sections: `no debug information; rebuild with DWARF (Emscripten -g, wasm-pack dwarf-debug-info)`
- Already stripped: `already stripped (build_id present, no debug sections); splitting would produce a useless companion`
- Dangling reference: `has external_debug_info but no local companion with matching build_id`

Running the command twice is safe. It detects prepared pairs and skips them rather than overwriting a good companion with an empty one, and a module stamped on the first run keeps the same `build_id` on later runs.

### Options

| Flag | Effect |
| --- | --- |
| `--dry-run` | Classify and preview; write nothing, upload nothing |
| `--no-upload` | Split only |
| `--require-dwarf` | Exit with an error if any scanned module lacks DWARF |
| `--out-dir <DIR>` | Write companions to a directory; the deployable is always stripped in place |
| `--strip-names` | Also drop the name section from the deployable; the companion keeps it |
| `--build-id <UUID>` | Use an explicit `build_id` instead of a random one |
| `--include-sources` | Upload source bundles alongside companions |
| `--wait` / `--wait-for <SECS>` | Wait for server-side processing |
| `--ignore` / `--ignore-file` | Skip paths while scanning |
| `--json` | Machine-readable output |

`--require-dwarf` is the CI guard. Use it when every module in a build is expected to ship debug info, and you want the pipeline to fail rather than deploy something unsymbolicatable.

`--dry-run` writes nothing at all, including the `build_id`, so it is safe to run against a tree you do not want modified.

### Relationship to `wasm-split`

The split follows the same algorithm as `wasm-split` and uses the same `wasmbin` library, so the companions it writes are equivalent to what `wasm-split -d <companion> --strip` produces: a shared `build_id`, a companion retaining the Code section, and DWARF removed from the deployable. Stamping behavior matches too — both tools give every module they touch a `build_id`.

It is a reimplementation rather than a shared codepath. `wasm-split` is a binary-only crate in Symbolicator with no library target, so there is no API to depend on. Because the two are maintained separately, treat equivalence as a property this command aims for, not a guarantee inherited from upstream.

`prepare` is also narrower than `wasm-split` by design. It is a pipeline for Sentry uploads, not a general-purpose WASM tool.

### Limitations

**A module without DWARF gets a `build_id` but no debug file.** Stamping makes the module identifiable; it does not make it symbolicatable. Nothing is uploaded for it, so stack traces resolve only as far as the runtime itself manages. For a name/symtab-only module that is usually function names and no line numbers, since browsers read the `name` section straight out of the deployed module. Uploading that module as its own debug file would not improve on what the runtime already reports, so the command does not. To get line-level frames, rebuild with DWARF and run `prepare` again — the `build_id` from the earlier run is preserved.

**`external_debug_info` always points at the companion filename.** There is no equivalent of `wasm-split --external-dwarf-url`, so you cannot embed an absolute URL for browser DevTools to fetch. This does not affect Sentry, which resolves debug files by `build_id`.

**Splitting always strips.** There is no mode that writes a companion while leaving the deployable module unstripped.

**`--out-dir` redirects the companion only.** The deployable module is always stripped in place so the path you deploy is the one that carries the `build_id`.

**`--strip-names` applies only to modules that are split.** A skipped module keeps its `name` section, which is what lets the runtime still report function names for it.

### Compatibility

The command reads module sections only, so it works regardless of which compiler produced the module. Build with DWARF enabled (Emscripten `-g`, wasm-pack `dwarf-debug-info`).

### Node API

The npm wrapper exposes the same flow:

```js
await cli.debugFiles.prepare({
  path: './dist',
  includeSources: true,
  wait: true,
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 
+- (debug-files) Add `debug-files prepare` to split WASM DWARF companions and upload them
 - (build) Add dSYM support to IPA uploads ([#3393](https://github.com/getsentry/sentry-cli/pull/3393))
 
 ### Fixes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -754,6 +754,63 @@ dependencies = [
 ]
 
 [[package]]
+name = "custom_debug"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2da7d1ad9567b3e11e877f1d7a0fa0360f04162f94965fc4448fbed41a65298e"
+dependencies = [
+ "custom_debug_derive",
+]
+
+[[package]]
+name = "custom_debug_derive"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a707ceda8652f6c7624f2be725652e9524c815bf3b9d55a0b2320be2303f9c11"
+dependencies = [
+ "darling",
+ "proc-macro2",
+ "quote",
+ "syn",
+ "synstructure",
+]
+
+[[package]]
+name = "darling"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc7f46116c46ff9ab3eb1597a45688b6715c6e628b5c133e288e709a29bcb4ee"
+dependencies = [
+ "darling_core",
+ "darling_macro",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d00b9596d185e565c2207a0b01f8bd1a135483d02d9b7b0a54b11da8d53412e"
+dependencies = [
+ "fnv",
+ "ident_case",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "syn",
+]
+
+[[package]]
+name = "darling_macro"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc34b93ccb385b40dc71c6fceac4b2ad23662c7eeb248cf10d529b7e055b6ead"
+dependencies = [
+ "darling_core",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "data-encoding"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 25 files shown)


szokeasaurusrex:
Hey @d2anamaria, just wondering why you are opening this PR in this repo?

This is the repo for the legacy `sentry-cli`, which is no longer under active development. This feature should probably be getting implemented in the [new `sentry` CLI's repo](https://github.com/getsentry/cli), instead.

Is there a concrete reason that this cannot be implemented in the new CLI? If not, please close this PR and implement the feature in the new CLI 🙏

PR #3408, opened 2026-09-11
feat(snapshots): Follow the server-selected objectstore usecase

NicoHinderling:
Snapshot uploads now request `usecase=auto` and build their Objectstore session from the usecase returned by the server rather than hardcoding `preprod`. Both batched HEAD checks and image uploads use that session, keeping their paths aligned with the minted token. This lets the backend move new snapshot content into `preprod_snapshots` without another CLI change.

Servers that omit the response field default to `preprod`, so existing Sentry versions retain their current behavior. The integration coverage exercises actual upload-options requests, authenticated Objectstore batches, and snapshot creation for `preprod_snapshots`, explicit `preprod`, and an omitted usecase. Object keys stay unchanged; the first upload after a usecase switch re-uploads images once because deduplication is scoped by usecase.

Pairs with getsentry/sentry#124194 (cross-usecase reads and the upload-options contract) and getsentry/sentry#124210 (default-off write selection). getsentry/cli#1587 implements the same contract in the replacement CLI; both clients are being updated. Register `preprod_snapshots` in Objectstore before deploying the backend prerequisite and enable the option only after the backend rollout. Older CLI releases keep writing to `preprod` and remain readable through fallback reads.


NicoHinderling:
Closing: the server side will move size-analysis content to a separate objectstore usecase instead, so snapshots keep using preprod and no CLI change is needed.

szokeasaurusrex (approved):
I trust you have tested this end to end? If yes, lgtm!

